PR #45, opened 2019-10-23
Port v1/v2 integration tests

roblourens:
Massive PR. This introduces about 80 integration tests from the v1/v2 codebase. It starts a debug server for each test, and tests send commands to it using `vscode-debugprotocol-testSupport`. Puppeteer is used to download chrome and for some tests, to click buttons on the page to trigger events.

There are a couple generations of tests represented. There are "simpler" tests that are years old, the ones at the top level. Then there are tests using new harnesses that were written recently, mainly by @digeff and @EricCornelson.

You can run them with the "Run int test" launch config. I've only tried on Mac. Currently they almost all fail, either due to issues that I've written down here: https://github.com/microsoft/vscode-pwa/issues/44 or other flakiness like Chrome not shutting down. Also we don't have a logger hooked up which makes them hard to debug.


roblourens:
Also I've left a lot of TODOs for myself, mainly where I had to do weird casts to get these to build with slightly different types and stricter TS rules.

And there is a lot that can be simplified. I copied code out of vscode-chrome-debug-core and the vscode-chrome-debug-core-testSupport packages (like the 2 testSetup.ts files and testUtils.ts), and some of it is overly complex and not all of it is used.

PR #47, opened 2019-10-24
feat: initial implementation of node attach

connor4312:
This doesn't totally work yet, but I figured this set of changes is big enough.
This implements attaching to an existing/running Node process. It refactors
some of the logic from `NodeLauncher` into an abstract `NodeLauncherBase`
class. We bring in the existing process picker and process tree logic, and
then to attach to a process we run the existing watchdog script (with a few
tweaks) against the target. I'm fairly happy with how this structure ended
up looking. Existing tests for the node runtime still pass.

The code in this PR successfully attaches to a running process, handles
breakpoint state and shows the callstack, but some things are still missing.
I'll add those in a separate PR, or this one if they end up being
one-line changes (I have a feeling there's just a few more .enable()
calls I need to make).

For a separate PR is setting environment variables in the attached process
so that child processes are tracked just like the programs we launch ourselves.

### DIFF
--- a/package.json
+++ b/package.json
@@ -911,7 +911,7 @@
                   "stopOnEntry": {
                     "type": "boolean",
                     "description": "%node.stopOnEntry.description%",
-                    "default": true
+                    "default": false
                   },
                   "console": {
                     "type": "string",
--- a/package.json
+++ b/package.json
@@ -107,7 +107,7 @@
   "activationEvents": [
     "onDebugInitialConfigurations",
     "onDebugResolve:NAMESPACE(node)",
-    "onCommand:extension.pickNodeProcess",
+    "onCommand:extension.NAMESPACE(node-debug).pickNodeProcess",
     "onCommand:extension.NAMESPACE(node-debug).prettyPrint",
     "onCommand:extension.NAMESPACE(node-debug).pickLoadedScript",
     "onCommand:extension.NAMESPACE(node-debug).toggleSkippingFile",
@@ -175,6 +175,11 @@
           "light": "resources/light/remove-all.svg",
           "dark": "resources/dark/remove-all.svg"
         }
+      },
+      {
+        "command": "extension.NAMESPACE(node-debug).attachNodeProcess",
+        "title": "%attach.node.process%",
+        "category": "Debug"
       }
     ],
     "menus": {
@@ -241,6 +246,9 @@
       {
         "type": "NAMESPACE(node)",
         "label": "%node.label%",
+        "variables": {
+          "PickProcess": "onCommand:extension.NAMESPACE(node-debug).pickNodeProcess"
+        },
         "configurationSnippets": [
           {
             "label": "%node.snippet.launch.label%",
--- a/src/targets/node/nodeAttacher.ts
+++ b/src/targets/node/nodeAttacher.ts
@@ -8,8 +8,11 @@ import { getWSEndpoint } from '../browser/launcher';
 import { spawnWatchdog } from './watchdogSpawn';
 import { NodeLauncherBase, IRunData } from './nodeLauncherBase';
 import { findInPath } from '../../common/pathUtils';
-import { Socket } from 'net';
 import { SubprocessProgram } from './program';
+import Cdp from '../../cdp/api';
+import { isLoopback } from '../../common/urlUtils';
+import { INodeTargetLifecycleHooks } from './nodeTarget';
+import { LeaseFile } from './lease-file';
 
 /**
  * Attaches to ongoing Node processes. This works pretty similar to the
@@ -19,6 +22,13 @@ import { SubprocessProgram } from './program';
  * child processes operate just like those we boot with the NodeLauncher.
  */
 export class NodeAttacher extends NodeLauncherBase<INodeAttachConfiguration> {
+  /**
+   * Tracker for whether we're waiting to break and instrument into the main
+   * process. This is used to avoid instrumenting unecessarily into subsequent
+   * children.
+   */
+  private capturedBreaker = false;
+
   /**
    * @inheritdoc
    */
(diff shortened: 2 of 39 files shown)


src/ui/processPicker.ts
```diff
@@ -0,0 +1,252 @@
+/*---------------------------------------------------------
```
connor4312: This is lightly modified code from the existing debug adapter, primarily just removing support for the legacy `--debug` protocol and protocol detection.

src/ui/processPicker.ts
```diff
@@ -0,0 +1,252 @@
...
+    name: 'process',
+    processId: `\${command:${Contributions.PickProcessCommand}}`,
+  };
+
+  if (!(await resolveProcessId(config))) {
+    return;
+  }
+
+  const cwd = guessWorkingDirectory(config);
+  const assignWorkspaceFolder = (obj: any) =>
```
connor4312: I wasn't super happy with this... in our defaults we have some thing, like sourcemap overrides, that map based on the workspaceFolder, and if you attach to a process in a multiroot workspace VS Code will yell at us saying it can't find a valid folder. This goes through and replaces instances of the `${workspaceFolder}` with the one we guessed. (Guessing logic is the same as what the config resolution already uses and used in v1/2 of the debug adapter)

src/ui/processTree.ts
```diff
@@ -0,0 +1,206 @@
+/*---------------------------------------------------------------------------------------------
```
connor4312: This file is identical, with Prettier formatting, to the one in node-debug

src/targets/node/nodeAttacher.ts
```diff
@@ -0,0 +1,45 @@
...
+import { NodeLauncherBase, IRunData } from './nodeLauncherBase';
+import { findInPath } from '../../common/pathUtils';
+import { Socket } from 'net';
+import { SubprocessProgram } from './program';
+
+/**
+ * Attaches to ongoing Node processes. This works pretty similar to the
+ * existing Node launcher, except with how we attach to the entry point:
+ * we don't have the bootloader in there, so we manually attach and enable
+ * the debugger, then evaluate and set the environment variables so that
```
roblourens: Not sure we should set environment variables when we attach. The existing DA doesn't work that way and it could cause strange behavior because they will be set at some random point in the program's execution

src/targets/node/watchdogSpawn.ts
```diff
@@ -0,0 +1,61 @@
...
+   */
+  ppid?: string;
+}
+
+const watchdogPath = join(__dirname, 'watchdog.js');
+
+/**
+ * Spawns a watchdog attached to the given process.
+ */
+export function spawnWatchdog(execPath: string, { ipcAddress, ...inspectorInfo }: IWatchdogInfo) {
```
roblourens: Can you explain what the watchDog is? Why does it run in a separate process?

src/ui/processPicker.ts
roblourens: `${workspaceFolder}` in a launch config should actually be handled by vscode, not the DA.

package.json
```diff
@@ -175,6 +175,11 @@
           "light": "resources/light/remove-all.svg",
           "dark": "resources/dark/remove-all.svg"
         }
+      },
+      {
+        "command": "extension.NAMESPACE(node-debug).attachNodeProcess",
```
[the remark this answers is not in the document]
connor4312: Yea, that's an option. By default `NAMSPACE(node-debug)` will convert into `pwa-node-debug`, but if you run gulp with the `--drop-in` flag it doesn't get a prefix. This was to allow us to develop vscode-pwa without having to self-host the extension, but still allowing us to do so with a build flag when we're ready.

src/targets/node/nodeAttacher.ts
connor4312: The downside is that not doing so will cause any child processes to not be instrumented. I think having an option for this would be good, I'll make a note to do that when I implement it.

src/ui/processPicker.ts
connor4312: Yea, but in a multi-root workspace where you use the process picker to choose a process to attach to (i.e. not running out of a launch.json), VS Code was failing to resolve the workspace folder. Do you know if there's a better approach here?

src/targets/node/watchdogSpawn.ts
connor4312: Pavel could explain more, the idea of the watchdog is to give us a target API for Node.js. Out of the box, Node doesn't include the Target namespace--there's no representation of child processes within its inspector protocol. The watchdog sits as a proxy which connects the Node process' CDP to the server we stand up within the BaseNodeLauncher. It handles the target IDs, opener IDs, making the node process 'look' like a webworker or any other thread within CDP.

src/targets/node/nodeAttacher.ts
connor4312: Something I also talked about with Kai yesterday was having an option to aggressively attach to any existing child processes of the tree (in an option). We already have the logic needed to get the process tree, if the children are run with `--inspect` we can connect to them out of the box, otherwise we could try to attempt to `putPidInDebugMode`. That's not for this PR, but could be a useful extension for scenarios where people run a cluster of long-running processes.

src/ui/processPicker.ts
roblourens: Where is `${workspaceFolder}` coming from in this case?

package.json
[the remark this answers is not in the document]
roblourens: Right, and this isn't really related to the PR, but after talking about the space path issue I was thinking that we might want to make it possible to use this as a drop in replacement and also run it side by side with the other versions. More like we did with node/node2. I can explain more later.

src/targets/node/nodeAttacher.ts
roblourens: Which env var controls that, NODE_OPTIONS? I agree that we should set ones for the debugger. Not user specified ones.

That's a good idea, there is also the 'cluster' module which will spawn children in debug mode if the parent is in debug mode. We already support that, we should support it here too.

src/targets/node/watchdogSpawn.ts
roblourens: Oh, I remember that now.

src/ui/processPicker.ts
connor4312: It comes from the configuration defaults. e.g. sourceMapOverrides depends on it, to resolve `webpack:///foo/bar.js` to an absolute path.

src/targets/node/nodeAttacher.ts
connor4312: Yep, NODE_OPTIONS, and then we have some extra options for the watchdog. I wouldn't touch any of the extra/user-specified variables.

package.json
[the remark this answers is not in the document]
connor4312: Ok. I'll keep it in the PR for now since we're doing this for the other command already, let's chat Monday 🙂

src/ui/processPicker.ts
roblourens: Oh ok. I guess I don't know what to do about that. We can leave it for now.

PR #48, opened 2019-10-24
Shennie/scriptskipping

Shenniey:
These changes are to add script skipping feature


connor4312:
Thanks for the PR! I can take care of the merge conflicts when it's time to merge, most of them are my fault 😄

Shenniey:
Ok! Thanks so much. That might be in a bit though cuz I just found a couple bugs.. 😅

roblourens:
Make sure we have some unit tests for the ScriptSkipper logic!

src/adapter/scriptSkipper.ts
```diff
@@ -0,0 +1,86 @@
...
+  private blackboxSender?: (params: cdp.Debugger.SetBlackboxPatternsParams) => Promise<cdp.Debugger.SetBlackboxPatternsResult | undefined>;
+
+  constructor(skipPatterns: string[]) {
+    this._userSkipPatterns = skipPatterns;
+    this._preprocessNodeInternals();
+    this._setRegexForNonNodeInternals();
+  }
+
+  private _preprocessNodeInternals(): void {
+    let nodeInternalRegex = new RegExp("^<node_internals>\/(.*)$");
```
digeff: Can we use const instead of let here?

src/adapter/scriptSkipper.ts
```diff
@@ -0,0 +1,86 @@
...
+    let nodeInternalPatterns = this._userSkipPatterns!.filter(pattern => pattern.includes("<node_internals>")).map( nodeInternal => {
+      return nodeInternalRegex.exec(nodeInternal)![1];
+    });
+    if (nodeInternalPatterns.length > 0) { // set variables to use to filter node internals when in module
+      if (nodeInternalPatterns.includes("**/*.js"))
+        this.skipAllNodeInternals = true; // this value is checked later at a later processing point
+      else
+        this._nodeInternalsRegex = this._createRegexString(nodeInternalPatterns);
+    }
+  }
```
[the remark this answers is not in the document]
Shenniey: Yup! Done

src/adapter/scriptSkipper.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  constructor(skipPatterns: string[]) {
+    this._userSkipPatterns = skipPatterns;
+    this._preprocessNodeInternals();
+    this._setRegexForNonNodeInternals();
+  }
+
+  private _preprocessNodeInternals(): void {
+    let nodeInternalRegex = new RegExp("^<node_internals>\/(.*)$");
+    let nodeInternalPatterns = this._userSkipPatterns!.filter(pattern => pattern.includes("<node_internals>")).map( nodeInternal => {
```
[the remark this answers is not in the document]
Shenniey: Done!

src/adapter/scriptSkipper.ts
```diff
@@ -0,0 +1,86 @@
+import cdp from '../cdp/api';
```
roblourens: Need a copyright heaader

src/adapter/scriptSkipper.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export class ScriptSkipper {
+  private _userSkipPatterns: string[];
+  private _nonNodeInternalRegex: string = "";
```
roblourens: We are using single quotes (should be in tslint)

src/adapter/scriptSkipper.ts
roblourens: You don't have to use the RegExp constructor, you can use a regex literal (like `/^<node_internals>/.*`/)

src/adapter/scriptSkipper.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  private _setRegexForNonNodeInternals(): void {
+    var nonNodeInternalGlobs = this._userSkipPatterns.filter(pattern => !pattern.includes("<node_internals>"));
+    this._nonNodeInternalRegex += this._createRegexString(nonNodeInternalGlobs);
+  }
+
+  private _createRegexString(patterns: string[]): string {
+    if (patterns.length == 0)
+      return ".^";
+    return patterns.map(pattern => pattern.replace('.', '\.')).join('|');
```
roblourens: Escaping `.` isn't sufficient for escaping regex characters, see the functions in the old codebase for doing that.

src/adapter/scriptSkipper.ts
```diff
@@ -0,0 +1,86 @@
...
+      if (nodeInternalPatterns.includes("**/*.js"))
+        this.skipAllNodeInternals = true; // this value is checked later at a later processing point
+      else
+        this._nodeInternalsRegex = this._createRegexString(nodeInternalPatterns);
+    }
+  }
+
+  private _setRegexForNonNodeInternals(): void {
+    var nonNodeInternalGlobs = this._userSkipPatterns.filter(pattern => !pattern.includes("<node_internals>"));
+    this._nonNodeInternalRegex += this._createRegexString(nonNodeInternalGlobs);
```
roblourens: Keeping regexes and strings and concatenating the string with no `|` looks suspicious but I haven't looked at it in depth. Also, I am not sure why you are keeping the "non node internal" separate from the "node internal" patterns.

I know you want to get this into master so you don't have to resolve conflicts every day, I am ok merging this as a work in progress. I will wait to do an in depth review, I am guessing it will change when dynamic updating is implemented. Or I can review it more before that if you want.

Shenniey: I'm joining the patterns in the _createRegexString function, which concatenates them with a `|`. For the "non node internal" separation, it's mostly because for Node, I'm waiting until a Node target is registered so that we get the node target from the process itself. I guess we can concatenate the two lists into a single regex, if that'd be better. 

Yes, I'd like to get this merged as a work in process -- but I'd also like to clean this up as much as possible 😅. But here is a todo list of changes I'd like to have in a second PR:
- working unit tests
- working integration tests
- error handling (don't show stacktrace when error occurs)
- browser launch/attach support
- add skipfiles to intellisense
- dynamic toggling of skip files

PR #49, opened 2019-10-24
fix: pause CDP events until the adapter is fully connected to avoid races

connor4312:
Stacked on #47. The issue was that when we attach to a running Node process, it already has all the scripts and threads loaded in memory, and it was sending this effectively instantly--before we finished setting up our listeners on the CDP connection.

My initial approach was to move the call to `await this._cdp.Debugger.enable({});` to the `afterBind()` hook, but I realize it was probably just luck that made it work with Node launches initially, so instead I implementing a pause/resume mechanic in the CDP connection, allowing us to pause events sent down the CDP until after we're listening and ready to start receiving them.

### DIFF
--- a/src/targets/node/nodeTarget.ts
+++ b/src/targets/node/nodeTarget.ts
@@ -37,6 +37,7 @@ export class NodeTarget implements Target {
   ) {
     this.connection = connection;
     this._cdp = cdp;
+    cdp.pause();
     this._targetId = targetInfo.targetId;
     this._scriptName = targetInfo.title;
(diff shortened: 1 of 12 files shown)


src/targets/node/nodeTarget.ts
```diff
@@ -161,6 +162,10 @@ export class NodeTarget implements Target {
     return this._cdp;
   }
 
+  public async afterBind() {
+    this._cdp.resume();
```
roblourens: I don't see where afterBind is called from

connor4312: fixed :)

PR #50, opened 2019-10-25
Stricter tsconfig settings

roblourens:
https://github.com/microsoft/vscode-pwa/issues/36

Adds a couple tsconfig rules. Some 'any' casts where the intent is not clear, some actual type fixes.


src/adapter/threads.ts
```diff
@@ -398,7 +398,7 @@ export class Thread implements VariableStoreDelegate {
       }
 
       this._pausedDetails = this._createPausedDetails(event);
-      this._pausedDetails[kPausedEventSymbol] = event;
```
connor4312: Not sure how much refactoring you wanna do here, but this is a good use case for a `WeakMap` (and actually most WeakMap polyfills work almost the same way here--defining a non-enumerable symbol on the key objects 😛 )

src/common/objUtils.ts
```diff
@@ -35,7 +35,7 @@ export function caseInsensitiveMerge<V>(
     return {};
   }
 
-  const out = {};
+  const out: any = {};
```
connor4312:
```suggestion
  const out: { [key: string]: V } = {};
```

should work

src/int-chrome/testSetup.ts
```diff
@@ -57,7 +57,7 @@ function formLaunchArgs(
 
   const argsWithDefaults = { ...chromeLaunchConfigDefaults, ...launchArgs };
   for (let k in argsWithDefaults) {
-    launchArgs[k] = argsWithDefaults[k];
+    launchArgs[k] = (argsWithDefaults as any)[k]; // TODO@rob
```
connor4312: TIL [multiline suggestions aren't a thing, somehow](https://github.com/isaacs/github/issues/1452), but

```
Object.assign(launchArgs, argsWithDefaults)
```

should do the trick without casts

src/nodeDebugConfigurationProvider.ts
```diff
@@ -228,7 +228,7 @@ function createLaunchConfigFromContext(
           dir += '/';
         }
       }
-      config['preLaunchTask'] = 'tsc: build - tsconfig.json';
+      (config as any)['preLaunchTask'] = 'tsc: build - tsconfig.json';
```
connor4312: You can probably add this as `preLaunchTask?: string;` in IMandatedConfiguration

src/targets/node/nodeLauncher.ts
```diff
@@ -323,7 +323,7 @@ function readEnvFile(file: string): { [key: string]: string } {
   }
 
   const buffer = stripBOM(fs.readFileSync(file, 'utf8'));
-  const env = {};
+  const env: any = {};
```
connor4312:
```suggestion
  const env: { [key: string]: string } = {};
```

PR #51, opened 2019-10-25
fix: don't send the initialized event until after we respond to initi…

EricCornelson:
…alize.

The DAP specifically states that the adapter should not send any events or messages before it responds to the initialize request. VS Code will handle things just fine if it does, but VS treats this as a fatal error and closes the session.

### DIFF
--- a/src/adapter/debugAdapter.ts
+++ b/src/adapter/debugAdapter.ts
@@ -61,8 +61,9 @@ export class DebugAdapter {
   async _onInitialize(params: Dap.InitializeParams): Promise<Dap.InitializeResult | Dap.Error> {
     console.assert(params.linesStartAt1);
     console.assert(params.columnsStartAt1);
-    this.dap.initialized({});
-    return DebugAdapter.capabilities();
+    const capabilities = DebugAdapter.capabilities();
+    setTimeout(() => this.dap.initialized({}), 0);
+    return capabilities;
   }
 
   static capabilities(): Dap.Capabilities {
--- a/src/binder.ts
+++ b/src/binder.ts
@@ -49,8 +49,11 @@ export class Binder implements Disposable {
 
     this._dap.then(dap => {
       dap.on('initialize', async () => {
-        dap.initialized({});
-        return DebugAdapter.capabilities();
+        const capabilities = DebugAdapter.capabilities();
+        setTimeout(() => {
+          dap.initialized({});
+        }, 0);
+        return capabilities;
       });
       dap.on('setExceptionBreakpoints', async () => ({}));
       dap.on('setBreakpoints', async params => {


connor4312:
Hah good timing, I just ran into a failing test that was caused by this issue 😄

roblourens:
Yes, in fact we can wait to send it until we are ready to receive breakpoint requests. I think it is currently buffering these to handle later.

src/dap/connection.ts
```diff
@@ -186,7 +186,12 @@ export default class Connection {
           } else {
             response.body = result;
           }
+          const postHandlerAction = result.__postHandlerAction;
+          delete result['__postHandlerAction'];
```
connor4312: I'm not sure I like adding a side-channel like this in the generic connection code. Would using `setTimeout(() => this.dap.initialized({}))` work instead? That's also not the prettiest thing, but perhaps it might be better?

EricCornelson: I agree it's not the best. I still prefer it over `setTimeout` though because at least it's deterministic, `setTimeout` has the possibility of introducing race conditions, which we don't like 😃

connor4312:
whoops, sorry, deleted my comment. I'm looking at this right now, I think the solution here might fix VS but there's another race that's happening...

EricCornelson:
Another option is rather than having the handlers return a response and have the connection automatically send it over dap, we just have the handlers return void and each handler can take care of sending its response as it likes.

This results in a little more code and a little less "magic" with responses being handled automatically, but allows for much greater flexibility.

roblourens:
Sorry I thought this was an issue, not a PR, because the beta notifications UI is confusing.

re: races, in the node launch case and break on load cases you are still starting in paused mode so that isn't an issue.

I still like automatically sending responses.

connor4312:
I'm not sure I like the idea of an 'after response' invokation in the connection level. I'm not sure to what degree we'd want to use it, or even whether it'd be entirely appropriate for the `initialized()` response long term. For instance, if we later want to wait for another event before sending that we're initialized in the future, this won't be as useful. Most of our calls are driven by events on our side, I think that giving a call synchronously after an instigating request is pretty rare--but I could be wrong, I don't claim to be a CDP/DAP guru just yet 😛 

Imo a `setTimeout(() => initialized(), 0)` is _probably_ fine for now.

EricCornelson:
> I'm not sure I like the idea of an 'after response' invokation in the connection level. I'm not sure to what degree we'd want to use it, or even whether it'd be entirely appropriate for the `initialized()` response long term. For instance, if we later want to wait for another event before sending that we're initialized in the future, this won't be as useful. It also opens the question of how we handle responses and errors, if we wanted to invoke a method that returned any non-void value.
> 
> Most of our calls are driven by events on our side, I think that giving a call synchronously after an instigating request is pretty rare--but I could be wrong, I don't claim to be a CDP/DAP guru just yet 😛
> 
> Imo a `setTimeout(() => initialized(), 0)` is _probably_ fine for now.

Fair enough. Changed it to setTimeout.

connor4312:
Not here, it was elsewhere in the adapter, fixed in one of my attach PRs

PR #53, opened 2019-10-25
fix: use default sourceMapOverrides if not in launch config.

EricCornelson:
Also pulling the default sourceMapPathOverrides from v1/v2

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -319,7 +319,7 @@ export type AnyLaunchConfiguration = AnyChromeConfiguration | AnyNodeConfigurati
 export type ResolvingNodeConfiguration = IMandatedConfiguration & Partial<AnyNodeConfiguration>;
 export type ResolvingChromeConfiguration = IMandatedConfiguration & Partial<AnyChromeConfiguration>;
 
-const baseDefaults: IBaseConfiguration = {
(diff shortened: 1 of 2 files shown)


connor4312:
This seems to break the mapping in the webpack demo project in vscode. It doesn't recognize that the index.js from the sourcemap is the same as the index.js in the source tree. Could you take a look?

EricCornelson:
Reverted the change of defaults to the v1/v2 version. For now I'm more worried about making sure we use the defaults if they're not specified.

src/configuration.ts
```diff
@@ -335,6 +335,7 @@ const baseDefaults: IBaseConfiguration = {
   skipFiles: [],
   smartStep: true,
   sourceMaps: true,
+  // keep in sync with sourceMapPathOverrides in package.json
```
connor4312: @roblourens do you have thoughts on perhaps generating the `contributes` section of the package.json? We're passing it through a Gulp build anyway, injection would be feasible. We have a bunch of common options duplicated and defaults we need to make sure to keep in sync, automating it would make it cleaner (and also let us create it in a type-safe way, which would be neat)

roblourens: I'm not really a fan. Once it's set up, the contributions almost never change, so it seems like we should prefer simplicity.

PR #54, opened 2019-10-25
Added basic telemetry

digeff:
Added some basic telemetry.

For DAP because they tend to match with user actions, and they seem to be not a lot of quantity, we send the telemetry events individually.

For CDP we get a lot of operations, so we batch them all together.

/* Sample telemetry:
 *
{
  "cdp/Target.attachedToTarget": {
    "succesful": {
      "totalTime": 7,
      "maxTime": 4,
      "avgTime": 2.3333333333333335,
      "count": 3,
      "breakdown": {
        "time": "[1,2,4]",
        <more properties might appear here in the future>
      }
    },
    <if there are failures, we'll have a failed: section here>
  },
"cdp/Debugger.scriptParsed": {
    "succesful": {
      "totalTime": 8,
      "maxTime": 5,
      "avgTime": 4,
      "count": 2,
      "breakdown": {
        "time": "[5,3]"
      }
    }
  },
  <etc...>
 }
}
 */

This is just the first iteration. At the moment we send all the CDP operations batched telemetry when we disconnect.


src/telemetry/telemetryReporter.ts
```diff
@@ -0,0 +1,114 @@
...
+    this.reportWithElapsedTime(receivedTime, cdpEventName, extractErrorDetails(error), RequestOutcome.Failed);
+  }
+
+  reportSuccesfullyHandledEvent(cdpEventName: string, receivedTime: HighResolutionTime) {
+    this.reportWithElapsedTime(receivedTime, cdpEventName, {}, RequestOutcome.Succesful);
+  }
+
+  private reportWithElapsedTime(receivedTime: HighResolutionTime, cdpEventName: string, properties: TelemetryEntityProperties, outcome: RequestOutcome) {
+    // Floating point numbers use too much space and we don't need that much precision, we use integers to use less space when batching timings
+    const elapsedTime = Math.ceil(calculateElapsedTime(receivedTime));
```
EricCornelson: Slightly different from the elapsedTime from the DAP reporter? Should we just do the Math.ceil inside `calculateElapsedTime`?

src/telemetry/telemetryReporter.ts
```diff
@@ -0,0 +1,114 @@
...
+    this.reportWithElapsedTime(receivedTime, dapCommand, {}, RequestOutcome.Succesful);
+  }
+
+  private reportWithElapsedTime(receivedTime: [number, number], dapCommand: string, properties: TelemetryEntityProperties, outcome: RequestOutcome) {
+    const elapsedTime = calculateElapsedTime(receivedTime);
+    this._rawTelemetryReporter.report('dap/' + dapCommand, { ...properties, time: elapsedTime, succesful: outcome === RequestOutcome.Succesful });
+  }
+}
+
+export class CdpTelemetryReporter {
```
EricCornelson: This class seems almost identical to `DapRequestTelemetryReporter` except `reportWithElapsedTime` batches the events.

Can we consolidate the duplicated code between the two?

src/telemetry/telemetryReporter.ts
```diff
@@ -0,0 +1,114 @@
...
+export function isRawTelemetryReporter(unknownObject: unknown): unknownObject is RawTelemetryReporter {
+  return !!(unknownObject as RawTelemetryReporter).report;
+}
+
+export interface RawTelemetryReporter {
+  flush: EventEmitter<void>;
+  report(entityName: string, entityProperties: TelemetryEntityProperties);
+}
+
+export class NullRawTelemetryReporter implements RawTelemetryReporter {
```
EricCornelson: Looks like this isn't used anywhere?

src/telemetry/telemetryReporter.ts
```diff
@@ -0,0 +1,114 @@
...
+  private reportWithElapsedTime(receivedTime: HighResolutionTime, cdpEventName: string, properties: TelemetryEntityProperties, outcome: RequestOutcome) {
+    // Floating point numbers use too much space and we don't need that much precision, we use integers to use less space when batching timings
+    const elapsedTime = Math.ceil(calculateElapsedTime(receivedTime));
+
+    // CDP generates a lot of operations, so instead of reporting them one by one, we batch them all together
+    this._batcher.add(new UnbatchedOpReport('cdp/' + cdpEventName, { ...properties, time: elapsedTime, succesful: outcome === RequestOutcome.Succesful }));
+  }
+}
+
+export function isRawTelemetryReporter(unknownObject: unknown): unknownObject is RawTelemetryReporter {
```
EricCornelson: Used anywhere?

src/telemetry/telemetryReporter.ts
```diff
@@ -0,0 +1,114 @@
...
+  flush: EventEmitter<void>;
+  report(entityName: string, entityProperties: TelemetryEntityProperties);
+}
+
+export class NullRawTelemetryReporter implements RawTelemetryReporter {
+  public readonly flush = new EventEmitter<void>();
+  report(entityName: string, entityProperties: object) {}
+}
+
+export class RawTelemetryReporterToDap implements RawTelemetryReporter {
```
EricCornelson: Is the "ToDap" necessary? Do we send telemetry anywhere else?

src/telemetry/telemetryReporter.ts
digeff: Space is not an issue on the DAP reporter, so I'm inclined to only do it for CDP.

src/telemetry/telemetryReporter.ts
digeff: Deleted

src/telemetry/telemetryReporter.ts
digeff: Deleted

src/telemetry/telemetryReporter.ts
digeff: I think it makes it clearer what this class is doing. At the moment we don't send telemetry anywhere else.

src/telemetry/telemetryReporter.ts
digeff: Done

src/telemetry/telemetryReporter.ts
```diff
@@ -14,28 +14,53 @@ enum RequestOutcome {
...
 
-  reportErrorWhileHandlingDapMessage(dapCommand: string, receivedTime: HighResolutionTime, error: unknown) {
-    this.reportWithElapsedTime(receivedTime, dapCommand, extractErrorDetails(error), RequestOutcome.Failed);
+  reportError(dapCommand: string, receivedTime: HighResolutionTime, error: unknown) {
+    this.reportOutcome(receivedTime, dapCommand, extractErrorDetails(error), RequestOutcome.Failed);
   }
 
-  reportSuccesfullyHandledDapMessage(dapCommand: string, receivedTime: HighResolutionTime) {
-    this.reportWithElapsedTime(receivedTime, dapCommand, {}, RequestOutcome.Succesful);
+  reportSucces(dapCommand: string, receivedTime: HighResolutionTime) {
```
EricCornelson: Spelling, "success" :)

digeff: Fixed

src/telemetry/telemetryReporter.ts
```diff
@@ -0,0 +1,124 @@
...
+      data: JSON.stringify(entityProperties)
+    });
+  };
+}
+
+// Pattern: The pattern recognizes file paths and captures the file name and the colon at the end.
+// Next line is a sample path aligned with the regexp parts that recognize it/match it. () is for the capture group
+//                                C  :     \  foo      \  (in.js:)
+//                                C  :     \  foo\ble  \  (fi.ts:)
+const extractFileNamePattern = /[A-z]:(?:[\\/][^:]*)+[\\/]([^:]*:)/g;
```
roblourens: I think we need to do this for non-windows paths too.

src/utils/performance.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+export type HighResolutionTime = [number, number];
+
+export function calculateElapsedTime(startProcessingTime: HighResolutionTime): number {
```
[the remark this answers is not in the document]
digeff: Handling nanoseconds is overkill. We did that before.

For CDP because it's a lot of information, we are going to drop the nanoseconds.
For DAP because it's not that much information, and it's already there, we are leaving it as-is.

src/telemetry/telemetryReporter.ts
digeff: Good catch. Fixed

src/targets/browser/browserLauncher.ts
```diff
@@ -44,7 +45,7 @@ export class BrowserLauncher implements Launcher {
     this._disposables = [];
   }
 
-  async _launchBrowser({ runtimeExecutable: executable, runtimeArgs, timeout, userDataDir, env, cwd }: IChromeLaunchConfiguration): Promise<CdpConnection> {
+  async _launchBrowser({ runtimeExecutable: executable, runtimeArgs, timeout, userDataDir, env, cwd }: IChromeLaunchConfiguration, rawTelemetryReporter: RawTelemetryReporter): Promise<CdpConnection> {
```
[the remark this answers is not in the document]
digeff: Sadly we don't have the Dap.Api object where we call the constructor

PR #57, opened 2019-10-25
feat: attach to child processes of attached children, support stopOnEntry

connor4312:
A similar mechanism was needed for both scenarios, so they're here. The
NodeTarget now has a lifecycle object, and for node launches we hook into
the `initialized()` point to set a breakpoint on the main script file that
the user asked us to run.

We need to be paused to evaluate code, so the NodeAttacher uses the
initialized() point to request a pause, and when the first pause happens it
retrieves telemetry from the child process, and also installs the environment
variables necessary to instrument children. After doing this, any forked
children are detected automagially--you can try this out by running
`node demos/node/parent.js --inspect` and attaching to it; you can view and
debug in the child process.

The thorny point is around uninstallation when we detach; we'll have a bunch
of orphaned options, and subsequently spawned children will then run into some
issues when trying to connect back to the debug server. Ideally we could have
some piece of code that the inspectee runs automatically when the debugger
detaches, but as far as I can see there's no such mechanism in place. We could
try to do this manually when we disconnect, but not all disconnections
will be graceful.

So next I looked for a way to detect and disable inspection if we can find out
the server went away. And any detection needs to be synchronous so that we can
run it inline in our `--required` script, so we can't try to connect to the
debug server. I ended up just using a lease file. It's not perfect, but it
seems to do the trick.

### DIFF
--- a/src/targets/node/nodeAttacher.ts
+++ b/src/targets/node/nodeAttacher.ts
@@ -8,8 +8,11 @@ import { getWSEndpoint } from '../browser/launcher';
 import { spawnWatchdog } from './watchdogSpawn';
 import { NodeLauncherBase, IRunData } from './nodeLauncherBase';
 import { findInPath } from '../../common/pathUtils';
-import { Socket } from 'net';
 import { SubprocessProgram } from './program';
+import Cdp from '../../cdp/api';
+import { isLoopback } from '../../common/urlUtils';
+import { INodeTargetLifecycleHooks } from './nodeTarget';
+import { LeaseFile } from './lease-file';
 
 /**
  * Attaches to ongoing Node processes. This works pretty similar to the
@@ -19,6 +22,13 @@ import { SubprocessProgram } from './program';
  * child processes operate just like those we boot with the NodeLauncher.
  */
 export class NodeAttacher extends NodeLauncherBase<INodeAttachConfiguration> {
+  /**
+   * Tracker for whether we're waiting to break and instrument into the main
+   * process. This is used to avoid instrumenting unecessarily into subsequent
+   * children.
+   */
+  private capturedBreaker = false;
+
   /**
    * @inheritdoc
    */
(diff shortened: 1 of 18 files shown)


src/targets/node/nodeAttacher.ts
```diff
@@ -46,5 +56,100 @@ export class NodeAttacher extends NodeLauncherBase<INodeAttachConfiguration> {
...
+  private async setEnvironmentVariables(
+    cdp: Cdp.Api,
+    run: IRunData<INodeAttachConfiguration>,
+    leasePath: string,
+  ) {
+    if (!run.params.attachSpawnedProcesses) {
+      return;
+    }
+
+    if (!isLoopback(run.params.address)) {
```
connor4312: (Because we need to --require the bootstrap file, which almost certainly won't exist on remote systems. Maybe one toggleable approach would be to send its contents over CDP and write it to a tmpdir on the remote machine.)

src/targets/node/nodeLauncher.ts
```diff
@@ -92,4 +96,37 @@ export class NodeLauncher extends NodeLauncherBase<INodeLaunchConfiguration> {
...
+    return {
+      initialized: async () => {
+        if (run.params.stopOnEntry) {
+          const params = {
+            url: absolutePathToFileUrl(resolve(run.params.cwd, run.params.program)),
+            lineNumber: 0,
+            columnNumber: 0,
+          };
+
+          await cdp.Debugger.setBreakpointByUrl(params);
```
roblourens: If stopOnEntry is set, we just want to stop where we are paused from `--inspect-brk`, ie don't autoresume. We don't need to set a breakpoint.

connor4312: That was the first thing I tried, after dissecting the existing node-debug2 extension. The issue is that that ends up breaking inside our bootloader.js, rather than the user's code, since that's the first thing that gets required.

PR #61, opened 2019-10-29
fix: breakpoint on first line not always applying

connor4312:
The issue was that we tell the debugger to break on a script that has a sourcemap in it so that we can set breakpoints. In the browser (where there's no offset / module preamble) this breaks at `(0, 0)`, so even if we then set a breakpoint at that location, once we automatically resume we'll never hit it because we'll already have passed it!

This PR adds a simple check that aborts automatically resuming if we have a breakpoint at the top of the current call stack. Fixes the last of the breakOnLoad tests (https://github.com/microsoft/vscode-pwa/issues/44)

image


src/adapter/threads.ts
```diff
@@ -382,6 +391,11 @@ export class Thread implements VariableStoreDelegate {
             scheduledPauseOnAsyncCall.id === event.asyncStackTraceId.id) {
           // Paused on the script which is run as a task for scheduled async call.
           // We are waiting for this pause, no need to resume.
+        } else if (
+          event.callFrames.length
+          && this.breakpoints.getBreakpointsAtLocation(event.callFrames[0].location).length
```
dgozman: I'd suggest this._handleSourceMapPause() to return boolean (whether to resume or not). It will end up in BreakpointManager anyway, which can make the decision. Otherwise, we start introducing more circular dependencies (BreakpointManager -> Thread -> BreakpointManager), which usually leads to less maintainable code. WDYT?

This way, it should pretty straightforward to check that any of uiLocations in Breakpoint.updateForSourceMap is at (0,0).

src/int-chrome/breakOnLoad.test.ts
```diff
@@ -117,7 +117,7 @@ suite('BreakOnLoad', () => {
       const bpCol = 1;
 
       await launchWithUrlAndSetBreakpoints(url, testProjectRoot, scriptPath, bpLine, bpCol);
-      await dc.assertStoppedLocation('breakpoint', {
+      await dc.assertStoppedLocation('step', {
```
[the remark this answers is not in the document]
connor4312: Oh, yea. I was thinking that this was coming from CDP but of course we rewrite it into DAP, so we can modify that readily enough 👍

src/adapter/threads.ts
```diff
@@ -17,6 +17,7 @@ import * as sourceUtils from '../common/sourceUtils';
 import { InlineScriptOffset } from '../common/sourcePathResolver';
 import { ScriptSkipper } from './scriptSkipper';
 import { AnyLaunchConfiguration, OutputSource } from '../configuration';
+import { BreakpointManager } from './breakpoints';
```
dgozman: Not used anymore?

src/adapter/breakpoints.ts
```diff
@@ -336,6 +383,16 @@ export class BreakpointManager {
     this._updateSourceMapHandler();
     return { breakpoints: breakpoints.map(b => b.toProvisionalDap()) };
   }
+
+  public hasBreakpointsAtLocation(location: AttachedLocation) {
```
dgozman: Unused?

src/adapter/breakpoints.ts
```diff
@@ -336,6 +383,16 @@ export class BreakpointManager {
...
+    return this.getBreakpointsAtLocation(location).length > 0;
+  }
+
+  public getBreakpointsAtLocation(location: AttachedLocation) {
```
dgozman: Unused?

src/adapter/breakpoints.ts
```diff
@@ -255,6 +288,20 @@ export class BreakpointManager {
       }
 
       await Promise.all(todo);
+
+      return {
```
dgozman: As I pointed before, this could be probably simplified, without the need to track attachedLoactions:

```
<here>
const remainPaused = (await Promise.all(todo)).some(x => !!x);
</here>

async updateForSourceMap(thread: Thread, script: Script): Promise<boolean> {
   ...
  return uiLocations.any(loc => !loc.lineNumber && !loc.columnNumber);
}
```

connor4312: Ah, I see what you mean. That seems to work.

src/adapter/breakpoints.ts
```diff
@@ -218,6 +230,26 @@ export class Breakpoint {
   }
 };
 
+/**
```
dgozman: This block about AttachedLocation seems unused.

PR #63, opened 2019-10-30
Report unhandled exceptions

digeff:
Now we report unhandled exceptions and unhandled rejections in telemetry

### DIFF
--- a/src/adapter/debugAdapter.ts
+++ b/src/adapter/debugAdapter.ts
@@ -106,9 +106,6 @@ export class DebugAdapter {
 
   async _onSetBreakpoints(params: Dap.SetBreakpointsParams): Promise<Dap.SetBreakpointsResult | Dap.Error> {
     const ids = generateBreakpointIds(params);
-    setTimeout(() => {
-       throw new Error('aweffeaw');
-    }, 10000);
     return this.breakpointManager.setBreakpoints(params, ids);
   }
 
--- a/src/adapter/debugAdapter.ts
+++ b/src/adapter/debugAdapter.ts
@@ -106,6 +106,9 @@ export class DebugAdapter {
 
   async _onSetBreakpoints(params: Dap.SetBreakpointsParams): Promise<Dap.SetBreakpointsResult | Dap.Error> {
     const ids = generateBreakpointIds(params);
+    setTimeout(() => {
+       throw new Error('aweffeaw');
+    }, 10000);
     return this.breakpointManager.setBreakpoints(params, ids);
   }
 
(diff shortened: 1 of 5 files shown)


src/adapter/debugAdapter.ts
```diff
@@ -106,6 +106,9 @@ export class DebugAdapter {
 
   async _onSetBreakpoints(params: Dap.SetBreakpointsParams): Promise<Dap.SetBreakpointsResult | Dap.Error> {
     const ids = generateBreakpointIds(params);
+    setTimeout(() => {
+       throw new Error('aweffeaw');
+    }, 10000);
```
connor4312: 🔥

src/telemetry/unhandledErrorReporter.ts
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { extractErrorDetails, RawTelemetryReporter } from './telemetryReporter';
+
+type ExceptionType = 'uncaughtException' | 'unhandledRejection';
+
+export function installUnhandledErrorReporter(telemetryReporter: RawTelemetryReporter): void {
+  process.addListener('uncaughtException', (exception: unknown) => {
```
connor4312: I believe this is run in the extension host within vscode, so we may end up getting exceptions thrown by other exceptions here. We probably want to filter these out, perhaps by checking its stacktrace to ensure it's coming from one of our scripts?

src/adapter/debugAdapter.ts
digeff: Thanks for catching this... This is how I was testing the code xD

src/telemetry/unhandledErrorReporter.ts
roblourens: I am also still not sure whether we should be moving it out of the EH to run in a separate process like other debug adapters.

digeff: Done

src/adapter/debugAdapter.ts
digeff: Done

src/telemetry/unhandledErrorReporter.ts
```diff
@@ -0,0 +1,61 @@
...
+  return errMsg;
+}
+
+const debugAdapterFolder = path.dirname(path.dirname(path.dirname(__dirname)));
+
+function shouldReportThisError(error: any): boolean {
+  // In VS Code, this debug adapter runs inside the extension host process, so we could capture
+  // errors from other pieces of software here. We check to make sure this is our error before reporting it
+  return !shouldFilterErrorsReportedToTelemetry
+    || (error && typeof error.stack === 'string' && error.stack.indexOf(debugAdapterFolder) !== -1);
```
connor4312:
```suggestion
    || (error && typeof error.stack === 'string' && error.stack.includes(debugAdapterFolder));
```

You can use the ES6 `String.prototype.includes`

PR #65, opened 2019-10-31
Now we report breakpoints statistics

digeff:
Now we report how many breakpoints were created, and were verified, and were hit at the end of each session.

Note: Because we assign breakpoints a new id each time they are returned from setBreakpoints, these metrics will not have the intuitive values that a user would guess at this point in time.


demos/web-worker/.vscode/launch.json
```diff
@@ -11,7 +11,8 @@
       "url": "http://localhost:5002",
       "server": {
         "program": "${workspaceFolder}/node_modules/serve/bin/serve.js",
-        "args": ["-p", "5002"]
+        "args": ["-p", "5002"],
+        "logging": {"cdp": "${workspaceFolder}/vscode-pwa-cdp.log", "dap": "${workspaceFolder}/vscode-pwa-dap.log"}
```
connor4312: I like the idea of having these in the demos -- can you add those filenames to the root .gitignore so people don't accidentally commit them?

src/adapter/breakpoints.ts
```diff
@@ -46,16 +48,7 @@ export class Breakpoint {
...
-      reason: 'changed',
-      breakpoint: {
-        id: this._dapId,
-        verified: true,
-        source: await this._resolvedUiLocation.source.toDap(),
-        line: this._resolvedUiLocation.lineNumber,
-        column: this._resolvedUiLocation.columnNumber
-      }
-    });
+    await this._manager.notifyBreakpointResolved(this._dapId, await this._resolvedUiLocation);
```
connor4312:
```suggestion
    await this._manager.notifyBreakpointResolved(this._dapId, this._resolvedUiLocation);
```

await is extraneous here

digeff: Fixed

demos/web-worker/.vscode/launch.json
digeff: Done

roblourens:
Why do we need this? v1 didn't have this right?

digeff:
v1 didn't have this. We were asking ourselves: "How can we know that our customers are happy with how the product is working?"

We realized that if a user sets a valid breakpoint and it doesn't hit when it should, then customers are going to be unhappy (e.g.: when the source map is invalid). This is the best idea we got so far as to how to track that.

For example, if we normally see 30% of all breakpoints hitting, and we release a fix, and only 10% breakpoints hits afterwards, it'll probably mean that we broke something with that release.

connor4312:
I set breakpoints that don't get hit all the time, if our quorum is large enough we might get a baseline but I'm not sure it'll be useful to indicate anything other than catastrophic failures (which our unit, integration, and manual testing should guard against.) I suppose we'll have to see how noisy that data is.

digeff:
I agree. It's not great. So far is the best idea we have...

roblourens:
We are still reporting info about everything the DA does, so is this info redundant? I mean, if there is a bug where many breakpoints are no longer reported, wouldn't we expect it to show up in there?

digeff:
At least with the current telemetry, we don't know how many breakpoints were set from the setBreakpoints call. I'm also not entirely sure if breakpointResolved  
calls are 1 to 1 to actual breakpoints. So this information is not redundant.

roblourens:
I'm not really convinced that this is useful info, but I'll merge it and we can remove it if it's not helpful.

PR #67, opened 2019-11-04
feat: implement logging

connor4312:
This is inspired by the previous debug adapter in functionality. It
buffers written logs until we attach sinks, as we can log information
before the DAP or extension context is available. It uses the existing
`trace` option, but is now (optionally) an object with some additonal information
which allows for filtering; since we now log all DAP and CDP messages,
I figured this could be useful during development to cut down on some
noise.

```ts
export interface ILoggingConfiguration {
  /**
   * Configures whether logs are also returned to the debug console.
   * Defaults to false.
   */
  console: boolean;

  /**
   * Configures the level of logs recorded. Defaults to "Verbose".
   */
  level: string;

  /**
   * Configures where on disk logs are written. If this is null, no logs
   * will be written, otherwise the extension log directory (in VS Code) or
   * OS tmpdir (in VS) will be used.
   */
  logFile: string | null;

  /**
   * Configures what types of logs recorded. For instance, `cdp` will log all
   * CDP protocol messages. If this is empty or not provided, tags will not
   * be filtered.
   */
  tags: ReadonlyArray<string>;
}
```

Logs are written to the file as JSON. I'll modify the existing log
viewer tool to be able to read JSON over its previous line-parsing logic.

### DIFF
--- a/src/baseConfigurationProvider.ts
+++ b/src/baseConfigurationProvider.ts
@@ -1,0 +1,53 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { AnyResolvingConfiguration, ResolvedConfiguration } from './configuration';
+import { fulfillLoggerOptions } from './common/logging';
+
+/**
+ * Base configuration provider that handles some resolution around common
+ * options and handles errors.
+ */
+export abstract class BaseConfigurationProvider<T extends AnyResolvingConfiguration>
+  implements vscode.DebugConfigurationProvider {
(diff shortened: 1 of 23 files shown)


src/baseConfigurationProvider.ts
```diff
@@ -0,0 +1,53 @@
...
+
+import * as vscode from 'vscode';
+import { AnyResolvingConfiguration, ResolvedConfiguration } from './configuration';
+import { fulfillLoggerOptions } from './common/logging';
+
+/**
+ * Base configuration provider that handles some resolution around common
+ * options and handles errors.
+ */
+export abstract class BaseConfigurationProvider<T extends AnyResolvingConfiguration>
```
connor4312: Since logging configuration is shared between all configuration providers, I refactored slightly and created the BaseConfigurationProvider which can handle these common options.

PR #72, opened 2019-11-05
feat: implement basic extension host debugging

connor4312:
This implements the foundation of extension host debugging. There are some things that don't work yet. For instance setting breakpoints is, as mentioned, buggy pending the outFiles work, but I wanted to get this in to avoid having one massive branch and PR later 🙂 

For this we create a new subtype of `NodeLauncherBase` to launch, and then attach to, the extension host. Extension host debugging is handled slightly differently in that VS Code gives us a 'launch' request, followed by an 'attach' request, rather than just a single 'launch' like Node.js and Chrome does. Aside from that, this is fairly straightforward, building upon most of our existing infrastructure.


gulpfile.js
```diff
@@ -74,14 +76,38 @@ gulp.task(
...
+  vsce.createVSIX({
+    ...minimist(process.argv.slice(2)),
+    cwd: 'out',
+    packagePath: path.join(__dirname, 'out', `${extensionId}.vsix`),
+  }),
+);
+
+gulp.task(
+  'package',
+  gulp.series('clean', gulp.parallel('compile', 'package:copy-modules'), 'package:vsix'),
```
connor4312: @EricCornelson I know you'd mentioned working on getting packaging working; I had these changes sitting on this branch from when I had dogfood'ed PWA, they might be useful to you.

connor4312: I'm actually going to cherry-pick these changes into another PR so that we can get CI and packaging working.

package.json
```diff
@@ -245,6 +246,161 @@
       }
     ],
     "debuggers": [
+      {
```
EricCornelson: I'm thinking that as we start to add even more debuggers, this list is going to start to get fairly unmanageable, and since there's a lot of duplication anyway, I wonder if it makes sense to dynamically compile the final package.json instead of cramming everything in here. (don't have to change anything for this PR, just a thought).

src/targets/node/program.ts
```diff
@@ -46,6 +46,28 @@ export class SubprocessProgram implements IProgram {
   }
 }
 
+/**
+ * A no-op program that never stops until stop() is called. Currently, we use
```
EricCornelson: Is this what we did in v1 as well?

package.json
connor4312: Yea that's what I'm thinking. Rob was hesitant on it before but editing this thing at this point is becoming a PITA. After you get your build PR in (so I don't step on your toes) I'll look at this if I have a few minutes.

src/targets/node/program.ts
connor4312: v1 actually does terminate it somehow. I had a mental bullet point, which is now a physical bullet point on https://github.com/microsoft/vscode-pwa/issues/58, to look at this some more.

PR #73, opened 2019-11-05
fix: get the package build target working

EricCornelson:
These changes get both the parcel bundling and vsix build targets working together.

A few important changes to note:
- Separated the packaging phase to use the dist folder instead of out. This helps keep packaging completely separate from the normal debug/dev builds
- As a result of the above change, the vsce package command now runs inside of the dist folder, so we need to copy any other needed files there as well (atm that's package.json, package.nls.json, and LICENSE)
- Changed the gulpfile to use the command-line version of vsce instead of the js lib. I found the js version very hard to debug because gulp tends to swallow console.log statements, so the build would fail with zero diagnostic output. Using the command line api solves that problem and makes builds easier to debug.
- Moved the tsconfig out of the src dir and into the root. I did this mainly because we are referencing the package.json in some parts of the code now, and if we use require(''), parcel can't resolve it, because we've gone one level deeper (from src to out/src). The solution to this is to use typescript import to import package.json. On compile, ts knows to change the relative path based on the output dir. But in order to let ts actually do this, we have to change the root of the ts project so that package.json is inside of it.


connor4312:
I ran into an issue running the package command--does it work for you?

image

EricCornelson:
It is working for me. Hmmm. Did the package script work for you before? I switched it from using the .ts files directly to using the compiled .js in out/ because it would ignore some typscript errors.

connor4312:
The package script didn't work before, but I [pushed some changes](https://github.com/microsoft/vscode-pwa/commit/0d49af0a06c762c8b91bb47e4cbe56339f630a23#diff-b9e12334e9eafd8341a6107dd98510c9) that I had lying around from a while ago that fixed it for me. I'll look around at what it's doing some more tomorrow, it's odd that it works for only one of us, package-lock'd installations _should_ be identical.

EricCornelson:
Indeed. Are you on Windows or Mac?

connor4312:
Mac

EricCornelson:
Ahh, I'll see if I get the same error in WSL

connor4312:
I'm still not sure why it works for you--we used to have `vscode` as a direct dependency, but no longer do (since it's advised to depend on the separate types file). The docs say that vscode should be [marked as external](https://code.visualstudio.com/api/working-with-extensions/bundling-extension#configure-webpack). They [don't have](https://github.com/parcel-bundler/parcel/issues/144) built-in support for externals like Webpack does, but someone [wrote a plugin](https://www.npmjs.com/package/parcel-plugin-externals) that might work. We could do that, or swap over to Webpack for this build. I have a preference to Webpack just because I've lots of experience with it as a user and plugin author, but don't feel strongly.

EricCornelson:
I'm fine with switching to webpack if it's easier to get working everywhere. I'll try out that config.

EricCornelson:
Alright, let's give it another shot. I merged in your webpack config and switched from parcel.

Everything is working for me on both Windows and Ubuntu 18.04, hopefully that means it works on Mac too (I don't have a Mac to test on ☹️)

EricCornelson:
Also a note if you install the extension, it will only work when launching with the proposed api flag due to: https://github.com/microsoft/vscode-pwa/issues/80

connor4312:
Cool, thanks! I made a couple tweaks to ensure the bootloader and terminateProcess.sh script can be found, seems to work well now.

.vscode/launch.json
```diff
@@ -43,6 +43,20 @@
       ],
       // "preLaunchTask": "npm: watch"
     },
+    {
+      "name": "Extension (packaged)",
+      "type": "extensionHost",
+      "request": "launch",
+      "runtimeExecutable": "${execPath}",
+      "args": [
+        "--enable-proposed-api=msjsdiag.pwa-debugger",
```
roblourens: I think this shouldn't be necessary?

PR #75, opened 2019-11-06
feat: option support in browser attach

connor4312:
This Adds filtering by `url`/`urlFilter` to the browser attacher --
previously it'd always connect to the first tab/target, which is not
always desirable if the user has several tabs open.

We also now support timeouts in the browser attacher and launcher,
set the network cache setting on the target page, and send and log
telemetry about the attached page, like the previous debug adapter did.

### DIFF
--- a/src/targets/browser/browserTargets.ts
+++ b/src/targets/browser/browserTargets.ts
@@ -144,11 +144,11 @@ export class BrowserTargetManager implements Disposable {
     });
     cdp.Target.setAutoAttach({ autoAttach: true, waitForDebuggerOnStart: true, flatten: true });
 
-    // For the 'top-level' page, gather telemetry and set the cache state.
-    if (!parentTarget) {
-      cdp.Network.setCacheDisabled({ cacheDisabled: this.launchParams.disableNetworkCache })
-        .catch(err => logger.info(LogTag.RuntimeTarget, 'Error setting network cache state', err));
+    cdp.Network.setCacheDisabled({ cacheDisabled: this.launchParams.disableNetworkCache })
+      .catch(err => logger.info(LogTag.RuntimeTarget, 'Error setting network cache state', err));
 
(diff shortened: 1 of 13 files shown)


src/targets/browser/browserTargets.ts
```diff
@@ -111,6 +144,14 @@ export class BrowserTargetManager implements Disposable {
     });
     cdp.Target.setAutoAttach({ autoAttach: true, waitForDebuggerOnStart: true, flatten: true });
 
+    // For the 'top-level' page, gather telemetry and set the cache state.
+    if (!parentTarget) {
+      cdp.Network.setCacheDisabled({ cacheDisabled: this.launchParams.disableNetworkCache })
```
dgozman: Disabling cache makes sense for all targets, not only the top page.

src/ui/sessionManager.ts
```diff
@@ -83,7 +83,11 @@ export class SessionManager implements vscode.DebugAdapterDescriptorFactory, Dis
     const session = new Session(debugSession);
     if (debugSession.configuration['__pendingTargetId']) {
       const pendingTargetId = debugSession.configuration['__pendingTargetId'] as string;
-      const target = this._pendingTarget.get(pendingTargetId)!;
+      const target = this._pendingTarget.get(pendingTargetId);
+      if (!target) {
```
dgozman: This worries me. Do you have a particular scenario for this? As I see, we only add to `this._pendingTarget` once, and remove from there once, and I am not sure about the situation which causes this issue.

connor4312: Another (possible future) direction I was looking for in this was being able to attach to _all_ browser tabs which matched the url or urlFilter. In that scenario I was able to get an error to be thrown here probably 20% of the time by quickly navigating between pages or opening/closing tables. I'm not able to repro in the current single-tab attachment though, so it's probably safe to remove this from this PR.

PR #76, opened 2019-11-06
feat: resurrect debug terminal and npm script

connor4312:
Pavel mentioned these today, the wiring here broke when we changed
config and commands around to support drop-in; the commit restores them
and brings back the 'special-cased' terminal launcher which we need
for the debugging terminal and npm script debugging.

I've also taken the opportunity to harden npm script debugging some
more. Now, it:
 - Supports multi-root workspaces (discoveres and merges npm scripts)
 - Has a friendly prompt if no package.json is found,
   or if there's a parse error.
 - Has a friendly prompt (instead of thrown error) if there are no
   open workspace folders.

Demo: https://memes.peet.io/img/9b84c8ad-cd47-4b96-9abd-f1731884803f.mp4


package.json
```diff
@@ -246,6 +259,15 @@
       }
     ],
     "debuggers": [
+      {
+        "type": "NAMESPACE(node-terminal)",
+        "label": "Javascript Auto-Attaching Terminal",
+        "configurationAttributes": {
+          "launch": {
+            "properties": {}
```
connor4312: I'm going to fill this out when I put in a PR (tomorrow/friday) to generate parts of the package.json, I don't want to copy all the common node launch stuff for a third time...

src/targets/node/terminalNodeLauncher.ts
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { AnyLaunchConfiguration, INodeTerminalConfiguration } from '../../configuration';
+import * as vscode from 'vscode';
```
roblourens: This is tying us to a model where the DA has to run in the extension host. I am still not sure about that, I think we should bring it up with Andre. Can leave it for now.

PR #77, opened 2019-11-07
feat: generate package.json debuggers

connor4312:
Came up in a few PRs. This commit adds a generation script for the
"debuggers" section of the contribution, and the language JSON. It's
fairly type-safe, enforcing that we translate all strings and have
all relevant options defined and described in the configuration JSON
schema (which caught a few existing problems!)


src/build/strings.ts
```diff
@@ -0,0 +1,156 @@
+import { sortKeys } from '../common/objUtils';
+
+const strings = {
```
EricCornelson: I think some of the logic here could be simplified if you just export `strings`, then you can just reference it from any other js/ts file and resolve things automatically rather than serializing everything to stdout and rebuilding it

connor4312: We do export `strings` and use it to generate our debuggers, but we also need to create the `package.nls.json`--we do so by running strings.ts from the gulpfile. We don't want to include it as a normal module in the gulpfile, since then Node will cache it which is undesirable for the `watch` task.

gulpfile.js
```diff
@@ -32,6 +33,46 @@ const allPackages = [];
  */
 const namespace = process.argv.includes('--drop-in') ? '' : 'pwa-';
 
+async function runBuildScript(name) {
```
EricCornelson: Could either `runBuildScript` or `readJson` be simplified with `util.promisify`? They're both marked as `async` but use no awaits.

src/build/strings.ts
EricCornelson: Makes sense

gulpfile.js
connor4312: readJson can, but util.promisify doesn't support the multiple callback items that execFile gives back to us

PR #81, opened 2019-11-12
Add resolveSourceMapLocations option

connor4312:
## Background

Today PWA works very well when debugging applications where the user's code is the main or only code running in the target. This is the case for most websites and applications. It's not the case for extension host debugging, however; when working on a VS Code extension, the user's code is running alongside that of other first and third party extensions.

The issue arises when these other extensions have absolutely-pathed sourcemaps--the kind Webpack produces. [We encourage](https://code.visualstudio.com/api/working-with-extensions/bundling-extension) authors to pack their extensions using a tool like Webpack, and in many configurations (like the same one our docs have) Webpack will produce sourcemaps for the compiled code. The paths in these sourcemaps are in the form `webpack:///src/extension.ts`. When a user is building their own extension, the debugger has no way to know that the `webpack:///src/extension.ts` being referred to from another extension sourcemap is not the same `src/extension.ts` that the user probably has in their "hello world" extension.

This causes breakpoints set in `src/extension.ts` to be mapped to code in other extensions. The user will 'hit' the breakpoint in their workspace files, but will actually be in some other author's code.

## Approaches

The v1 adapters solve this issue by searching for sourceMaps in `outFiles`, marking breakpoints in mapped files as 'pending', and then setting the breakpoints in the compiled code the first time we load a source referenced from those maps. This generally works well, but has a downside that if a source is present in multiple places in the compiled code (multiple Webpack chunks, for instance) we'll only set the breakpoint in one location. Rob mentioned the TypeScript team ran into this issue.

PWA works differently. We optimistically set breakpoints by searching disk sourcemaps, and then in all sources that match references or _paths_ from sourcemaps we encounter at runtime. Where v1 is conservative about ensuring we don't break in undesired locations, PWA is aggressive about ensuring that breakpoints are set and can be hit in all potential locations.

The PWA approach probably the more correct approach, and as mentioned it works very well with zero-configuration for most scenarios.

## This PR

There's a new option `resolveSourceMapLocations`. Once again, I expect this to be set to a non-default value only in rare circumstances. It's a set of minimatch patterns, empty by default for Chrome, and for Node.js set to:

```json
"resolveSourceMapLocations": [
  "${workspaceFolder}/**",
  "!**/node_modules/**"
],
```

Conveniently this also works for URLs; say for example you're loading sources from a CDN that has source maps which conflict with those in your workspace, you could exclude that:

```json
"resolveSourceMapLocations": [
  "**",
  "!https://cdn.contoso.com/**"
],
```

<details>
<summary>Previous implementation (before PR updates)</summary>
In this PR, we add a launch.json option, turned on by default only for the extension host debugger, which omits setting absolute paths of sourcemap-derived sources if the original sourcemap was not present in the workspace/rootPath. This prevents the breakpoints from being set by-path for external sources, while also allowing sources to be viewed even if their filenames conflict work workspace sources (previously, if you broke in any `src/extension.ts`, we'd assume it was your local version and not show the correct contents).

As far as implementation goes, this works by extracting some of the scanning we were already doing for the breakpoint predictor into a `LocalSourceMapRepository`, and sharing that with the SourceContainer. The repository keeps track of md5 hashes of content we've seen, against which we can check remote sourcemaps.

The primary weakness of this approach is that if the local compilation results are a different than the ones being debugged, breakpoints won't be hit. However, we're unsure if this will end up being a problem; few people will actually _need_ this option, and it remains to be seen how many of those will also attempt to debug mismatched versions. (Also, if the sourcemaps are different, the user's debugging experience will already be degraded to a greater or lesser extent.)

There's an open question around performance of the the hashing; I think we should be fine but we'll keep an eye on it. At least for the disk scanning phase, we can cache file hashes by their mtime if necessary.
<details>

Almost finishes https://github.com/microsoft/vscode-pwa/issues/58


src/adapter/breakpointPredictor.ts
```diff
@@ -52,126 +59,122 @@ export class BreakpointsPredictor {
...
-      try {
-        const map = await sourceUtils.loadSourceMap(sourceMapUrl, 0);
-        if (!map)
+      }
+
+      for (const b of params.breakpoints || []) {
+        const entry = map.generatedPositionFor({
+          source: metadata.sourceUrl,
+          line: b.line,
+          column: b.column || 1,
```
EricCornelson: Aren't columns in source-maps zero-based? Would this create a problem if we wanted a breakpoint at column 0 (e.g. column would evaluate to false because it's 0, and we'd always swap in 1)?

connor4312: This was just moving around whitespace here, but. DAP columns and lines are 1-based, CDP columns and lines and 0-based. In PWA we keep them 1-based until right before we send the CDP request (in _setByScriptId and _setByUrl in breakpoints.ts)

EricCornelson:
How might this work for projects whose source maps never exist on disk in the first place (e.g. React, Angular, anything hosted in the webpack dev server)?

src/adapter/breakpointPredictor.ts
EricCornelson: Right, but in the source-map library, lines are 1-based, and columns are 0 based. We have to convert from DAP here.

connor4312:
This wouldn't do much for those scenarios, we'd advise to leave the option off for those scenarios. It's more rare to hit this scenario running off the webpack-dev-server, generally users will be bundling their single build with Webpack and not including other sources outside that build. I don't think v1 has any special binding today either; like this approach, outFiles depends on  paths on disk, and webpack-dev-server by default keeps a memory-based filesystem.

src/adapter/breakpointPredictor.ts
```diff
@@ -52,126 +59,122 @@ export class BreakpointsPredictor {
...
-      const entries = await fsUtils.readdir(dirOrFile);
-      const filtered = entries.filter(entry => entry !== 'node_modules' && entry[0] !== '.');
-      await Promise.all(filtered.map(entry => this._scan(path.join(dirOrFile, entry))));
-    }
-  }
+  private async _createInitialMapping(absolutePath: string) {
+    const sourcePathToCompiled: MetadataMap = new Map();
+    for (const [, metadata] of Object.entries(await this.repo.findAllChildren(absolutePath))) {
+      const baseUrl = metadata.sourceMapUrl.startsWith('data:')
+        ? sourceUtils.parseSourceMappingUrl(await fsUtils.readfile(metadata.compiledPath!))
```
dgozman: I think this should be just `metadata.compiledPath`.

connor4312:
> perhaps we can disallow webpack source maps outside of workspace to point into workspace sources while resolving the paths/urls? Would this solve the problem without a flag?

Unfortunately that's kind of hard to do. The URL that sourcemaps are served from may not indicate the workspace they're from. For extensions, most sourcemaps during development are files on disk, but if you're debugging a website we don't know what local workspace `http://localhost/foo.map.js` might be served from.

dgozman:
> Unfortunately that's kind of hard to do. The URL that sourcemaps are served from may not indicate the workspace they're from. For extensions, most sourcemaps during extension development are files on disk, but if you're debugging a website we don't know what local workspace `http://localhost/foo.map.js` might be served from without some kind of matching like we implement here.

But I think in this particular case (where the problem with multiple extension.ts files comes from) source maps have file urls, which makes it work. Am I missing something?

Basically, this PR does exactly that (disallowing sourcemaps not present in the workspace to refer to workspace files), and I just propose to detect that automatically (crossing workspace boundary from source map to sources). We can always introduce the config option which forces some behavior, but the less options the better the user experience.

connor4312:
I see. We could do that, gain some simplicity and reliability in that scenario in exchange for not being able to `forceLocalSourceMap` on code running in browser. @roblourens curious to hear your thoughts ^

roblourens:
That does sound simpler. For other cases maybe we could consider sourcemaps coming from `*/node_modules/*` to be "out of the project" also. 

I looked for the discussions about this issue with Typescript or other projects, and couldn't find them. But it's easy to see the issue in Typescript, they build several targets into the same folder, `tsc.js`, `tsserver.js`, etc, and use `outFiles` that just points at one file, which is how you would work around it for v1.

connor4312:
I've updated the PR. I am much happier with this solution. It resolves the 'outdated file' scenario and provides some additional usability.

There's a new option `resolveSourceMapLocations`. Once again, I expect this to be set to a non-default value only in rare circumstances. It's a set of minimatch patterns, empty by default for Chrome, and for Node.js set to:

```json
"resolveSourceMapLocations": [
  "${workspaceFolder}/**",
  "!**/node_modules/**"
],
```

Conveniently this also works for URLs; say for example you're loading sources from a CDN that has source maps, you could exclude that:

```json
"resolveSourceMapLocations": [
  "**",
  "!https://cdn.contoso.com/**"
],
```

I opted to keep the bit of directory scanning refactors I implemented earlier, since there's now a decent set of unit tests for them and it provides an extension point if there's any local discovery work we decide to do in the future.

Thanks for your suggestions!

src/targets/sourcePathResolver.ts
```diff
@@ -47,7 +85,7 @@ export abstract class SourcePathResolverBase<T extends ISourcePathResolverOption
 
     localPath = fixDriveLetter(localPath);
     logger.verbose(LogTag.RuntimeSourceMap, `Mapped remoteToLocal: ${remotePath} -> ${localPath}`);
-    return path.resolve(localPath);
+    return resolve(localPath);
```
connor4312: This was changed a few days ago and was causing the `posix -> win` test case to fail on osx (and would have failed win -> posix on windows). Fixed it here.

src/common/urlUtils.ts
```diff
@@ -9,6 +9,20 @@ import Cdp from '../cdp/api';
 import { escapeRegexSpecialChars } from './sourceUtils';
 import { AnyChromeConfiguration } from '../configuration';
 
+let isCaseSensitive = process.platform !== 'win32';
```
roblourens: This again 😁 @digeff

Long history, we need to be careful to be case insensitive, but case "preserving", meaning we send back the correct case to vscode and set breakpoints with the correct case. And in v1 we set the bps with a case-insensitive regex.

connor4312: Duly noted! For now we just use this for the micromatch test, but it's good to know that we'll need to be careful around this going forward.

Issue #40, opened 2019-10-22
Column breakpoints not implemented


roblourens:
- Have this code `x++;   y++; `
- Set a breakpoint after `x`
- The breakpoint moves to `y++`, it should prefer the left breakpoint position.

roblourens:
There is a broken test for this, but actually this is not a big deal now that breakpointLocations exist.

connor4312:
PWA doesn't implement column breakpoints; we should support them as a parity item.

PR #87, opened 2019-11-13
fix: breakpoints not applying on first line of block statements

connor4312:
In sourcemaps, if you have a function like we do in the webworker
example containing:

```ts
first.onchange = function() {
  console.log('hello');
```

...then `function() {\n  ` is a single segment in the sourcemap. When
we try to set a breakpoint at the start of the line containing
`console.log`, the previous behavior resolved the position at the start
of the containing segment, so the breakpoint would get set right
before the `function()` expression rather than the console.log
statement. This PR updates the positioning to prefer the next 
segment, rather than previous segment, when resolving positions.

Sourcemaps also use 0-indexing for columns, which we didn't account
for. I'm not sure if this was causing any issues, but I fixed it here
nonetheless. (Perhaps related to #40.)

### DIFF
--- a/src/adapter/sources.ts
+++ b/src/adapter/sources.ts
@@ -12,6 +12,7 @@ import * as utils from '../common/urlUtils';
 import * as errors from '../dap/errors';
 import { ScriptSkipper } from './scriptSkipper';
 import { delay } from '../common/promiseUtil';
+import { SourceMapConsumer } from 'source-map';
 
 const localize = nls.loadMessageBundle();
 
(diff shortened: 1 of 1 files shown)


src/adapter/sources.ts
```diff
@@ -373,10 +374,19 @@ export class SourceContainer {
       const map = this._sourceMaps.get(compiled._sourceMapUrl!)!.map;
       if (!map)
         continue;
-      const entry = map.generatedPositionFor({source: sourceUrl, line: uiLocation.lineNumber, column: uiLocation.columnNumber});
+      const entry = map.generatedPositionFor({
+        source: sourceUrl,
+        line: uiLocation.lineNumber,
+        column: uiLocation.columnNumber - 1, // source map columns are 0-indexed
```
EricCornelson: I think technically speaking, we should be checking the `linesStartAt1` and `columnsStartAt1` properties of the initialize args and adjusting appropriately, but I'm not sure if or when that might ever be false.

connor4312: Not sure either. We assume 1-indexing elsewhere quite heavily. We should probably adjust this for correctness, https://github.com/microsoft/vscode-pwa/issues/88

PR #89, opened 2019-11-13
feat: track vscode lifecycles correctly

connor4312:
Just something we weren't doing. The existing adapter reads the process
ID when it attaches to a target, and so did we (for Node targets). I
shifted around some logic into a `NodeAttacherBase` so that we can
reuse that query in the extension host, and then from there we reuse
the `TerminalProgram` to track the liveness of the vscode instance
by process ID.

Things I tested that work (from a vscode 'parent' debugging a 'child'):
 - Halting debugging from inside the parent instance kills the child
 - Closing the child clears the debugging state in the parent (I think
   it's actually a minor bug in the existing debug adapters where
   this doesn't happen...)
 - Restarting works as expected

Finishes/Fixes https://github.com/microsoft/vscode-pwa/issues/58

### DIFF
--- a/src/targets/node/nodeAttacherBase.ts
+++ b/src/targets/node/nodeAttacherBase.ts
@@ -10,6 +10,7 @@ import { INodeTargetLifecycleHooks } from './nodeTarget';
 import { IDisposable } from '../../common/disposable';
 import { logger } from '../../common/logging/logger';
 import { LogTag } from '../../common/logging';
+import { delay } from '../../common/promiseUtil';
 
 /**
  * Base class that implements common matters for attachment.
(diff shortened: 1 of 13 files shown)


src/targets/node/nodeAttacherBase.ts
```diff
@@ -0,0 +1,89 @@
...
+      return {};
+    }
+
+    let toDispose: Promise<ReadonlyArray<IDisposable>> = Promise.resolve([]);
+    this.capturedEntryProgram = this.program;
+
+    return {
+      initialized: async () => {
+        toDispose = this.onFirstInitialize(cdp, run);
+        await toDispose;
```
connor4312: We set it like this so that in the event a close happens before initialization finishes, we don't leak resources

roblourens:
> Closing the child clears the debugging state in the parent (I think
it's actually a minor bug in the existing debug adapters where
this doesn't happen...)

I don't understand, what do you mean "clear debugging state"?

connor4312:
This works well once https://github.com/microsoft/vscode/pull/84933 or equivalent is in. It doesn't hurt to merge before then (no more broken than it is already and it fixes some things...), so I'm good getting this in.

> I don't understand, what do you mean "clear debugging state"?

Ending the debug session when the vscode instance is closed. This actually works in the existing debuggers _most_ of the time, but I've seen it sometimes not happen occasionally. I'll keep an eye on it, but this doesn't regress.

Issue #26, opened 2019-10-16
Launch Config Parity


connor4312:
## New Options

The following new options are tracked for support:

 - [x] logging -- new logging object that improves granularity
 - [ ] nodeFilter
 - [x] baseURL -- (reconcile with path mappings?)

Additionally, vscode-pwa supports booting and debugging a server and client in a single launch config. It would make sense to preserve that behavior. Therefore, the chrome `launch` action config can have an additional property `server` that contains the Node launch config.

```json
{
  "type": "pwa-chrome",
  "request": "launch",
  "name": "Web page 2",
  "url": "http://localhost:5002",
  "server": {
    "program": "${workspaceFolder}/main.js"
  }
}
```

## Existing Options

We want to make vscode-pwa match and reach parity with the existing launch config options. This tracks that works.

### node: launch

- [x] address
- [x] args
- [x] autoAttachChildProcesses
- [x] console
- [x] cwd
- [x] disableOptimisticBPs
- [x] env
- [x] envFile
- [x] externalConsole
- [x] localRoot
- [x] ~outDir~ -- previously deprecated, now removed 
- [x] outFiles -- merge with resolveSourceMapLocations (?)
- [x] outputCapture
- [x] port
- [x] program
- [x] remoteRoot
- [x] restart
- [x] runtimeArgs
- [x] runtimeExecutable
- [x] runtimeVersion
- [x] showAsyncStacks
- [ ] skipFiles -- depends on https://dev.azure.com/devdiv/DevDiv/_workitems/edit/431549 @Shenniey
- [x] smartStep @roblourens 
- [x] sourceMapPathOverrides
- [x] sourceMaps
- [x] stopOnEntry
- [x] timeout
- [x] trace -- #34 
- [x] useWSL

### node: attach
 - [x] address
 - [x] cwd
 - [x] disableOptimisticBPs
 - [x] localRoot
 - [x] ~outDir~ -- previously deprecated, now removed 
- [x] outFiles
 - [x] port
 - [x] processId
 - [x] protocol
 - [x] remoteRoot
 - [x] restart
 - [x] showAsyncStacks
 - [ ] skipFiles -- depends on https://dev.azure.com/devdiv/DevDiv/_workitems/edit/431549 @Shenniey
 - [x] smartStep @roblourens 
 - [x] sourceMapPathOverrides
 - [x] sourceMaps
 - [x] stopOnEntry
 - [x] timeout
 - [x] trace -- #34 

### chrome: launch
[...]

PR #90, opened 2019-11-14
Implement smartStep

roblourens:
For #26

With this we will auto-step through lines in files without mappings in files that have sourcemap. For example, code inserted by TS into a transpiled file that doesn't have source lines associated with it. Skipped frames are also grayed out and marked with "skipped by smartStep".


roblourens:
Will add some tests too

PR #92, opened 2019-11-14
Apply defaults to config inside adapter, without the config provider involved

roblourens:
It won't be invoked when old node-debug dispatches to pwa-node, or for the VS case
- And always activate when node-debug activates
- And add "request" back for tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,6 +99,7 @@
     "npm-run-all": "^4.1.5",
     "nyc": "^14.1.1",
     "puppeteer": "^1.20.0",
+    "request": "^2.88.0",
     "request-promise-native": "^1.0.7",
     "sinon": "^7.5.0",
     "tslint": "5.11.0",
@@ -113,6 +114,7 @@
   "enableProposedApi": true,
   "activationEvents": [
     "onDebugInitialConfigurations",
+    "onDebugResolve:node",
     "onDebugResolve:NAMESPACE(node)",
     "onDebugResolve:NAMESPACE(node-terminal)",
     "onCommand:extension.NAMESPACE(node-debug).pickNodeProcess",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -99,6 +99,7 @@
     "npm-run-all": "^4.1.5",
     "nyc": "^14.1.1",
     "puppeteer": "^1.20.0",
+    "request": "^2.88.0",
```
connor4312: Was this intentional?

roblourens: Yes, it's needed for tests (peerDependency of "request-promise-native")

roblourens:
Actually, should I remove the code from the debug config providers that does the same thing?

connor4312:
The default configs have some portions that reference the workspace folder that vs code needs to resolve. I think it's fine as a stepping stone here but it might not always be safe to fill in the defaults outside of the config provider.

roblourens:
I realized that after some debugging, we won't resolve `${workspaceFolder}` here. I think I should still merge this for the VS scenario but they should know if they need to pass `rootPath`. Also I think we should merge rootPath/outFiles.

connor4312:
Sounds good.

In addition to the breakpoint predictor, `rootPath` is used as a fallback for some other scenarios but I don't _think_ it's ever needed/hit there with our fulfilled defaults. I agree we should probably coalesce it into something more specific such as outFiles.

roblourens:
I'm not sure what you're referring to

connor4312:
`assignWorkspaceFolder` in `attachProcess` is a special-case equivalent of `resolveVariablesInConfig`. I guess not entirely remove the fact that we need it, but deduplicate it at least.

PR #94, opened 2019-11-14
Shenniey/scriptskippingUpdates

Shenniey:
Added toggling, browser launch, and browser attach.

Still need to add unit tests and a couple of bug fixes


src/adapter/debugAdapter.ts
```diff
@@ -260,6 +262,38 @@ export class DebugAdapter {
...
+  private async _isSourceInCurrentStack(params: Dap.ToggleSkipFileStatusParams): Promise<Dap.Source | undefined> {
+    const currentStack = await this._onStackTrace({ threadId: this._thread?.id! });
+    const stackFrames = (currentStack as Dap.StackTraceResult).stackFrames;
+
+    let foundStackFrame;
+    if (params.resource && stackFrames) {
+      foundStackFrame = stackFrames.find(frame => frame.source && frame.source.path?.toLocaleLowerCase() === params.resource?.toLocaleLowerCase());
+    } else {
+      foundStackFrame = stackFrames.find(frame => frame.source && frame.source.sourceReference === params.sourceReference);
+    }
```
connor4312: Do you need any special handling for dealing with sourcemapped scripts here?

src/adapter/scriptSkipper.ts
```diff
@@ -87,10 +91,15 @@ export class ScriptSkipper {
...
-    let fullLibNames = nodeInternalsNames.map(name => name + '.js');
-    fullLibNames.push('^internal/.+\.js|');
-    this._nodeInternalsRegex = this._createRegexString(fullLibNames);
+  public setAllNodeInternals(nodeInternalsNames: string[]): void {
+    this._allNodeInternals = nodeInternalsNames.map(name => name + '.js');
+  }
+
+  public toggleSkipFileStatus(url: string | undefined): void {
+    if (url) {
+      let currentSkipValue = this._isUrlSkippedMap.get(url);
```
connor4312: Prefer `const` here

src/adapter/scriptSkipper.ts
```diff
@@ -87,10 +91,15 @@ export class ScriptSkipper {
...
 
-  public setAllNodeInternalsToSkip(nodeInternalsNames: string[]): void {
-    let fullLibNames = nodeInternalsNames.map(name => name + '.js');
-    fullLibNames.push('^internal/.+\.js|');
-    this._nodeInternalsRegex = this._createRegexString(fullLibNames);
+  public setAllNodeInternals(nodeInternalsNames: string[]): void {
+    this._allNodeInternals = nodeInternalsNames.map(name => name + '.js');
+  }
+
+  public toggleSkipFileStatus(url: string | undefined): void {
```
connor4312: Since it's not meaningful to pass in undefined to this method, I'd prefer to drop the undefined. Mentally it's like having an extra function overload that takes `undefined` and does nothing with it.

src/common/sourceUtils.ts
```diff
@@ -330,3 +330,58 @@ export function escapeRegexSpecialChars(str: string, except?: string): string {
   const r = new RegExp(`[${useRegexChars}]`, 'g');
   return str.replace(r, '\\$&');
 }
+
+export function fileUrlToPath(urlOrPath: string): string {
```
connor4312: We have this as `fileUrlToAbsolutePath` in urlUtils

src/common/sourceUtils.ts
```diff
@@ -330,3 +330,58 @@ export function escapeRegexSpecialChars(str: string, except?: string): string {
...
+          urlOrPath = '/' + urlOrPath;
+      }
+
+      urlOrPath = fixDriveLetterAndSlashes(urlOrPath);
+  }
+
+  return urlOrPath;
+}
+
+export function fixDriveLetterAndSlashes(aPath: string, uppercaseDriveLetter = false): string {
```
connor4312: This already exists in pathUtils :)

src/common/sourceUtils.ts
```diff
@@ -330,3 +330,58 @@ export function escapeRegexSpecialChars(str: string, except?: string): string {
...
+          aPath.substr(0, prefixLen + 1) +
+          aPath.substr(prefixLen + 1).replace(/\//g, '\\');
+  } else if (aPath.match(/^[A-Za-z]:/)) {
+      aPath = aPath.replace(/\//g, '\\');
+  }
+
+  return aPath;
+}
+
+export function fixDriveLetter(aPath: string, uppercaseDriveLetter = false): string {
```
connor4312: Same here and below

src/adapter/scriptSkipper.ts
roblourens: Should this be calling `this._updateBlackboxedUrls` too, or how does that actually get called after toggling?

src/targets/node/nodeSourcePathResolver.ts
```diff
@@ -20,11 +20,18 @@ export class NodeSourcePathResolver extends SourcePathResolverBase<IOptions> {
       return '';
     }
 
-    const withBase = path.resolve(this.options.basePath, this.applyPathOverrides(url));
-    return this.rebaseRemoteToLocal(withBase);
+    const modified = this.applyPathOverrides(url);
+
+    if (modified !== url) {
```
roblourens: Why this change?

src/targets/node/nodeTarget.ts
```diff
@@ -57,7 +58,8 @@ export class NodeTarget implements Target {
     if (targetInfo.title)
       this._targetName = `${basename(targetInfo.title)} [${targetInfo.targetId}]`;
     else this._targetName = `[${targetInfo.targetId}]`;
-
+    if (skipFiles.length)
```
roblourens: Seems skipping is only enabled if I have configured it at the beginning, but how does that work with toggling?

src/adapter/scriptSkipper.ts
Shenniey: Fixed!

src/common/sourceUtils.ts
Shenniey: Oh sorry, I missed that! Change to use urlUtils

src/common/sourceUtils.ts
Shenniey: Fixed

src/common/sourceUtils.ts
Shenniey: Sorry for missing all of these!

src/targets/node/nodeSourcePathResolver.ts
Shenniey: Without this change, this function would return the incorrect absolute path for the node internals. It would apply the path overrides, which would come back unchanged, and then tack on the base path, resulting in something that looked like "C:\projects\appbase\internal\modules\cjs\loader.js", which would throw the logic off since I'm basing filtering of node's internals on whether the source has an absolute path.

src/targets/node/nodeTarget.ts
Shenniey: Good catch! Thank you! I've since fixed it.

src/adapter/scriptSkipper.ts
Shenniey: Yes, it should! Sorry about that. Fixed

Shenniey: Yep that makes sense. Fixed

src/adapter/debugAdapter.ts
Shenniey: I... don't think so? When I tested this the stackFrame's source path is the correct source value for sourcemapped scripts. As in, when it's running index.js, the stackFrame has the index.ts as its source. But there could definitely be some cases that I'm missing . 

@roblourens Are there special cases for handling sourcemapped scripts?

src/adapter/debugAdapter.ts
```diff
@@ -262,6 +264,48 @@ export class DebugAdapter {
...
+      const source = this.sourceContainer.source(dapSource);
+      if (source) {
+        source._blackboxed = !source._blackboxed;
+      }
+      this.sourceContainer.scriptSkipper!.toggleSkipFileStatus(params.resource);
+      this._withThread(thread => thread.pause());
+    }
+    return {};
+  }
+
```
roblourens: I'm guessing this would throw if not actually paused. Make it safe/more predictable by checking that.

src/adapter/debugAdapter.ts
```diff
@@ -262,6 +264,48 @@ export class DebugAdapter {
...
+    if (dapSource && params.resource) {
+      const source = this.sourceContainer.source(dapSource);
+      if (source) {
+        source._blackboxed = !source._blackboxed;
+      }
+      this.sourceContainer.scriptSkipper!.toggleSkipFileStatus(params.resource);
+      this._withThread(thread => thread.pause());
+    }
+    return {};
+  }
```
roblourens: Maybe call it `lookupSourceInCurrentStack` or something, "is" sounds like it would return a boolean

src/adapter/debugAdapter.ts
```diff
@@ -262,6 +264,48 @@ export class DebugAdapter {
...
+    }
+
+    if (!this.sourceContainer.scriptSkipper) {
+      if (!this._thread || !this._thread.cdp()) {
+        return {};
+      }
+      this.sourceContainer.initializeScriptSkipper(new ScriptSkipper([]));
+      this.sourceContainer.scriptSkipper!.setBlackboxSender(this._thread.cdp().Debugger);
+    }
+
```
roblourens: This requires `params.resource` but `_isSourceInCurrentStack` doesn't, should it work or not?

src/adapter/debugAdapter.ts
```diff
@@ -262,6 +264,48 @@ export class DebugAdapter {
...
+    if (params.resource) {
+      const absPath = urlUtils.fileUrlToAbsolutePath(params.resource);
+      if (absPath) {
+        params.resource = absPath;
+      }
+    }
+
+    if (!this.sourceContainer.scriptSkipper) {
+      if (!this._thread || !this._thread.cdp()) {
+        return {};
```
roblourens: Not clear to me why this is lazily initialized. Seems like some code would be simpler if was created immediately on startup

src/adapter/debugAdapter.ts
```diff
@@ -262,6 +264,48 @@ export class DebugAdapter {
...
+    }
+
+    const dapSource = await this._isSourceInCurrentStack(params);
+    if (dapSource && params.resource) {
+      const source = this.sourceContainer.source(dapSource);
+      if (source) {
+        source._blackboxed = !source._blackboxed;
+      }
+      this.sourceContainer.scriptSkipper!.toggleSkipFileStatus(params.resource);
+      this._withThread(thread => thread.pause());
```
roblourens: Is the goal of this just to refresh the callstack? This is sending a pause request to the runtime

src/adapter/debugAdapter.ts
```diff
@@ -262,6 +264,48 @@ export class DebugAdapter {
...
+      }
+      this.sourceContainer.initializeScriptSkipper(new ScriptSkipper([]));
+      this.sourceContainer.scriptSkipper!.setBlackboxSender(this._thread.cdp().Debugger);
+    }
+
+    const dapSource = await this._isSourceInCurrentStack(params);
+    if (dapSource && params.resource) {
+      const source = this.sourceContainer.source(dapSource);
+      if (source) {
+        source._blackboxed = !source._blackboxed;
```
roblourens: Not clear to me why you toggle a property on the Source, then call scriptSkipper.toggle which toggles another property in a map. Ideally the state would just be kept in one place

roblourens: I think you could get rid of the `_blackboxed` property and have the ScriptSkipper be the source of truth, and have the Source able to query the ScriptSkipper as needed?

src/adapter/scriptSkipper.ts
```diff
@@ -59,26 +55,38 @@ export class ScriptSkipper {
...
-    if (this._isUrlSkippedMap.get(url) && this.blackboxSender) {
-      this._blackboxedUrls.push(url);
-      let blackboxPattern = '^' + this._blackboxedUrls.join('|') + '$';
+    if (this.blackboxSender) {
+      if (this._isUrlSkippedMap.get(url)) {
+        this._blackboxedUrls.push(url);
+      } else {
+        this._blackboxedUrls = this._blackboxedUrls.filter(blackBoxedUrl => blackBoxedUrl !== url);
+      }
+      let blackboxPattern = '^' + this._createRegexString(this._blackboxedUrls) + '$';
```
roblourens: I think you might need parens around the regex string, or the ^ and $ will just apply to the first and last patterns?

src/targets/node/nodeTarget.ts
```diff
@@ -57,7 +58,8 @@ export class NodeTarget implements Target {
...
+      this._scriptSkipper = new ScriptSkipper(skipFiles);
```
roblourens: If each Target has a ScriptSkipper, does that mean that if I toggle a script's skip state, that only applies to one target? It seems like that should apply to the whole debug session. But, I'm not totally sure about that...

src/targets/node/nodeTarget.ts
```diff
@@ -176,9 +178,19 @@ export class NodeTarget implements Target {
     this._cdp.Runtime.on('executionContextDestroyed', event => {
       if (event.executionContextId === defaultCountextId) this.connection.close();
     });
+    await this._setAllNodeInternals();
```
roblourens: `_initializeAllNodeInternals`

src/targets/node/nodeSourcePathResolver.ts
roblourens: I think the sourceMapPathOverrides shouldn't apply at all to paths that we know are node internals. We don't want to break this if one pattern accidentally applies.

roblourens:
Check the CI failures too. They might be fixed if you merge in the latest master.

Also, the strategy that you use here of checking each loaded script against patterns is different than v1, but I like it. Please add a short description to the ScriptSkipper about the basic strategy.

src/adapter/debugAdapter.ts
roblourens: Where does it get mapped back to `index.js` though? It looks like if this is `index.ts`, then that goes directly to the ScriptSkipper and would get sent to the runtime, when it needs to be resolved to `index.js` because node doesn't know what `index.ts` is.

PR #95, opened 2019-11-15
feat: implement column breakpoints

connor4312:
Self-explanitory. This was the clearest approach I saw; it looks like we
do a lot of work to map back and forth, but in the majority of cases
we'll only ever enter any of these loops once (e.g. a source without
a sourcemap, or a simple source present in one bundle of compiled code.)
Seems to work well, I'll work on getting some unit tests going tomorrow
morning.

image

Fixes https://github.com/microsoft/vscode-pwa/issues/40


src/adapter/breakpoints.ts
```diff
@@ -11,8 +11,10 @@ import * as urlUtils from '../common/urlUtils';
 import { rewriteLogPoint } from '../common/sourceUtils';
 import { BreakpointsStatisticsCalculator } from '../statistics/breakpointsStatistics';
 import { TelemetryEntityProperties } from '../telemetry/telemetryReporter';
+import { logger } from '../common/logging/logger';
+import { LogTag } from '../common/logging';
 
-type LineColumn = { lineNumber: number, columnNumber: number }; // 1-based
+type LineColumn = { lineNumber: number; columnNumber: number }; // 1-based
```
connor4312: Here to `getBreakpointLocations` is mostly just Prettier auto-formatting

src/adapter/sources.ts
```diff
@@ -326,28 +326,37 @@ export class SourceContainer {
...
-    const result: UiLocation[] = [];
-    this._addSourceMapUiLocations(uiLocation, result);
-    result.push(uiLocation);
-    this._addCompiledUiLocations(uiLocation, result);
-    return result;
+  /**
+   * Returns all the possible locations the given original location can map
+   * to, taking into account source maps.
+   */
+  public allPossibleLocations(uiLocation: UiLocation): UiLocation[] {
```
connor4312: Aside from making this public, no functional changes in this file; I was anticipating using some of these methods directly but it turns out that this existing method was what I needed. Left it as the slightly more idiomatic implementation (at least for someone a FP background 😛 )

src/adapter/breakpoints.ts
```diff
@@ -273,29 +279,119 @@ export class BreakpointManager {
...
+    // we'll only every find one viable location with a script.
+    let todo: Promise<Dap.BreakpointLocation[]>[] = [];
+    for (let i = 0; i < startLocations.length; i++) {
+      const start = startLocations[i];
+      const end = endLocations[i];
+
+      // Only take the first script that matches this source. The breakpoints
+      // are all coming from the same source code, so possible breakpoints
+      // at one location where this source is present should match every other.
+      const scripts = thread.scriptsFromSource(start.source);
```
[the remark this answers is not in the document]
connor4312: I'll skip and log when this happens. Curious about the frequency here.

Issue #60, opened 2019-10-28
webpack library configuration does not resolve source map sources


dgozman:
Here is a sample webpack configuration file

```
module.exports = {
  target: 'node',
  devtool: 'source-map',
  entry: './src/lib.ts',
  module: {
    rules: [
      {
        loader: 'ts-loader',
        exclude: /node_modules/
      }
    ]
  },
  node: {
    __dirname: false,
  },
  resolve: {
    extensions: [ '.ts', '.js' ]
  },
  externals: [nodeExternals()],
  output: {
    filename: 'lib.js',
    path: path.resolve(__dirname, 'build'),
    library: 'lib',
    libraryTarget:'umd'
  }
};
```

This produces source map source urls which look like `webpack://lib/./src/lib.ts` and should be mapped to `src/lib.ts`.

Here is a [sample fix](https://github.com/dgozman/vscode-pwa/commit/d81febb82695b4d48b209bb91192f33ebf5d57ab) against an outdated branch.

connor4312:
@dgozman it looks like your fork is private, can you make it public or push the fix to a branch on this repo so we can see it?

connor4312:
Thanks, I can see it now!

You might have seen we moved over this handling to the user-configurable `sourceMapOverrides` config matching the existing adapters, which doesn't support full regexes right now. I'm not sure how/whether this scenario works with the current adapters, but I was able to reproduce the issue with PWA using the sample code you provided. I think my next step would be to first see if this reproduces with the existing adapters; if it doesn't, then there's something we missed in porting the logic to PWA, if it does then we should probably figure out how to take your linked fix into the `sourceMapOverrides`-driven world.

This is endgame week for vscode so I don't anticipate getting too much headway on this issue this week. If you have time to look at applying the fix to the current repo, that would be great, otherwise I'll look into this issue late this week or, more likely, next week 🙂

dgozman:
Thanks, Connor! I don't have much context about `sourceMapOverrides`, that's why I didn't port the fix to the master myself, but I will be happy to review if needed.

PR #97, opened 2019-11-15
feat: support webpack namespaces in source map overrides

connor4312:
This involved adding a 'non-capturing group' syntax for the host map
overrides section. I used the RegEx-style `?:*` sequence, which is not
that pretty, I'm open to ideas. Anyway, this works. The default
source map URIs in webpack are actually formed[1] like:

```
webpack://[namespace]/[resourcePath]
```

It just happens the namespace is usually empty for main targets, but
this is configurable is defaults to `lib` for library targets,
hence #60. This PR adds a non-capturing group for the namespace.

Also makes source map overrides its own class, which saves processing
time (we can pre-parse and generate the regexes) and makes it more
readily testable.

 1. https://github.com/webpack/webpack/blob/91a9f97657b34375cd99e8ba8ef13dd795f1e7c3/lib/SourceMapDevToolPlugin.js#L115

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -436,7 +436,7 @@ export const baseDefaults: IBaseConfiguration = {
   rootPath: '${workspaceFolder}',
   // keep in sync with sourceMapPathOverrides in package.json
   sourceMapPathOverrides: {
-    'webpack:///*': '*',
+    'webpack://?:*/*': '*',
     'webpack:///./~/*': '${workspaceFolder}/node_modules/*',
     'meteor://💻app/*': '${workspaceFolder}/*',
   },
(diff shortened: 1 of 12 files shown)


src/configuration.ts
```diff
@@ -436,7 +436,7 @@ export const baseDefaults: IBaseConfiguration = {
   rootPath: '${workspaceFolder}',
   // keep in sync with sourceMapPathOverrides in package.json
   sourceMapPathOverrides: {
-    'webpack:///*': '*',
+    'webpack://?:*/*': '*',
```
roblourens: So this basically ignores the first segment and matches with the second segment? This was added for projects that end up with `webpack:///absolute/path` and it sounds like we can't handle both that and this other syntax without more smarts

roblourens: Oh, it goes from 3 / to 2, and I read the description more closely, it sounds like you are on top of it 👍

connor4312: Yep 😄

PR #100, opened 2019-11-18
feat: chrome behavior parity, allow debugging vscode

connor4312:
We use the Chrome debugger to debug VS Code itself. Launching this
depended on some specific behaviors  we didn't have before:

 - the URL is optional by default, only navigate
   the page if a URL is provided
 - don't set the user data dir unless explicitly requested when a
   custom runtime executable is provided
 - when a port is provided, skip scanning the browser stdout and instead
   just poll for the debugger on that port (vscode's launch script
   didn't display the magic string the launcher was after)
 - some cleanup as I was debugging in the chrome launcher; allow
   providing the `--remote-debugging-port` to be equivalent to an
   explicit `port` option.

### DIFF
--- a/src/cdp/transport.ts
+++ b/src/cdp/transport.ts
@@ -6,7 +6,7 @@ import WebSocket from 'ws';
 import * as events from 'events';
 import { HighResolutionTime } from '../utils/performance';
 import { CancellationToken } from 'vscode';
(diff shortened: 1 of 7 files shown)


connor4312:
Going to go ahead and merge this to enable self-hosting. Feel free to still drop comments here if there are any concerns though.

PR #101, opened 2019-11-18
Ignore telemetry events in tests

digeff:
This solves part of the telemetry events not being matched in the expected text for some tests.

I have a lot of unit test failures when I run the tests, so it's difficult to see whether this addresses all the telemetry issues in the tests or not.

I added a .only option to itIntegrates


connor4312:
Thanks!

PR #104, opened 2019-11-19
Fixes for case sensitivity issues on windows

connor4312:
Last night I managed to get one of my local projects on Windows, I managed to get into a case where the path casing it was using stopped matching up what was going around in the adapter, which caused a whole assortment of issues. This PR attempts to fix them...

### DIFF
--- a/src/adapter/breakpointPredictor.ts
+++ b/src/adapter/breakpointPredictor.ts
@@ -11,6 +11,7 @@ import { uiToRawOffset } from './sources';
 import { LocalSourceMapRepository } from '../common/sourceMaps/sourceMapRepository';
 import { ISourceMapMetadata } from '../common/sourceMaps/sourceMap';
 import { SourceMapConsumer } from 'source-map';
+import { MapUsingProjection } from '../common/datastructure/mapUsingProjection';
 
 // TODO: kNodeScriptOffset and every "+/-1" here are incorrect. We should use "defaultScriptOffset".
 const kNodeScriptOffset: InlineScriptOffset = { lineOffset: 0, columnOffset: 62 };
@@ -100,7 +101,9 @@ class DirectoryScanner {
   }
 
   private async _createInitialMapping(absolutePath: string) {
-    const sourcePathToCompiled: MetadataMap = new Map();
+    const sourcePathToCompiled: MetadataMap = new MapUsingProjection(
+      urlUtils.lowerCaseInsensitivePath,
+    );
     for (const [, metadata] of Object.entries(await this.repo.findAllChildren(absolutePath))) {
       const baseUrl = metadata.sourceMapUrl.startsWith('data:')
         ? metadata.compiledPath
@@ -121,11 +124,10 @@ class DirectoryScanner {
           continue;
         }
 
-        const normalizedPath = urlUtils.lowerCaseInsensitivePath(resolvedPath);
-        let set = sourcePathToCompiled.get(normalizedPath);
+        let set = sourcePathToCompiled.get(resolvedPath);
         if (!set) {
           set = new Set();
-          sourcePathToCompiled.set(normalizedPath, set);
+          sourcePathToCompiled.set(resolvedPath, set);
         }
         set.add({ ...metadata, sourceUrl: url });
       }
@@ -137,7 +139,7 @@ class DirectoryScanner {
   public async predictResolvedLocations(params: Dap.SetBreakpointsParams) {
     const sourcePathToCompiled = await this._sourcePathToCompiled;
     const absolutePath = params.source.path!;
-    const set = sourcePathToCompiled.get(urlUtils.lowerCaseInsensitivePath(absolutePath));
+    const set = sourcePathToCompiled.get(absolutePath);
 
     if (!set) return;
     for (const metadata of set) {
--- a/src/adapter/breakpointPredictor.ts
+++ b/src/adapter/breakpointPredictor.ts
@@ -10,6 +10,7 @@ import { InlineScriptOffset, ISourcePathResolver } from '../common/sourcePathRes
 import { uiToRawOffset } from './sources';
 import { LocalSourceMapRepository } from '../common/sourceMaps/sourceMapRepository';
 import { ISourceMapMetadata } from '../common/sourceMaps/sourceMap';
+import { SourceMapConsumer } from 'source-map';
 
 // TODO: kNodeScriptOffset and every "+/-1" here are incorrect. We should use "defaultScriptOffset".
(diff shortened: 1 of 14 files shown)


src/adapter/breakpointPredictor.ts
```diff
@@ -120,10 +121,11 @@ class DirectoryScanner {
           continue;
         }
 
-        let set = sourcePathToCompiled.get(resolvedPath);
+        const normalizedPath = urlUtils.lowerCaseInsensitivePath(resolvedPath);
```
connor4312: This is a new function that returns a lowercase version of the path if we're on a case-insensitive filesystem. I'm using this as hashmap keys here and elsewhere to avoid duplication.

src/adapter/breakpointPredictor.ts
```diff
@@ -153,13 +155,14 @@ class DirectoryScanner {
...
           column: b.column || 1,
+          bias: SourceMapConsumer.LEAST_UPPER_BOUND,
         });
         if (entry.line === null) {
           continue;
         }
 
         const { lineNumber, columnNumber } = uiToRawOffset(
-          { lineNumber: entry.line || 1, columnNumber: entry.column || 1 },
+          { lineNumber: entry.line || 1, columnNumber: entry.column ? entry.column + 1 : 1 },
```
connor4312: This and the bias adjustment above are changes we applied in the 'runtime' breakpoints earlier that I forgot to add over here

src/adapter/breakpoints.ts
```diff
@@ -178,7 +179,7 @@ export class Breakpoint {
     this._setUrlLocations.add(urlLocation);
 
     const location = {
-      url,
+      url: fixDriveLetter(url, false),
```
connor4312: CDP seems to (at least on my machine...) always want lowercased drive letters. @EricCornelson I'm curious if you ran into this before/whether it's a platform specific thing.

src/adapter/sources.ts
```diff
@@ -295,7 +295,7 @@ export class SourceContainer {
     if (ref.sourceReference)
       return this._sourceByReference.get(ref.sourceReference);
     if (ref.path)
-      return this._sourceByAbsolutePath.get(ref.path);
+      return this._sourceByAbsolutePath.get(utils.lowerCaseInsensitivePath(ref.path));
```
connor4312: Similar thing here. Without this we weren't marking breakpoints as resolved since source lookups failed.

src/common/sourceMaps/sourceMapRepository.ts
```diff
@@ -77,10 +80,6 @@ class Directory {
         .filter(bn => bn !== 'node_modules' && bn !== '.')
         .sort()
         .map(async bn => {
-          if (!getCaseSensitivePaths()) {
-            bn = bn.toLowerCase();
-          }
```
connor4312: Rather than just lowercasing all paths, we now get the 'true' casing of the path, if necessary, before passing it into our internal logic. This ensures that all paths we return are exactly correct to what's on the filesystem for the case-sensitive CDP.

src/adapter/breakpoints.ts
EricCornelson: I hadn't noticed that as an issue with CDP, but I have noticed vscode is case-sensitive (even on windows) when matching paths, e.g. when matching a loaded script to a file in the workspace, (some example issues: https://github.com/microsoft/vscode/issues/44074 https://github.com/microsoft/vscode/issues/43959 https://github.com/microsoft/vscode/issues/12448)

At one point I think Rob proposed using https://github.com/microsoft/vscode-uri to help keep things more standardized, but that lib also has its inconsistencies (but at least they would be consistent with vscode 😄 )

src/adapter/sources.ts
```diff
@@ -438,7 +438,7 @@ export class SourceContainer {
     this._sourceByReference.set(source.sourceReference(), source);
     if (source._compiledToSourceUrl)
       this._sourceMapSourcesByUrl.set(source._url, source);
-    this._sourceByAbsolutePath.set(source._absolutePath, source);
+    this._sourceByAbsolutePath.set(utils.lowerCaseInsensitivePath(source._absolutePath), source);
```
EricCornelson: Does it make sense to create a specific "file map" class for this, so we don't have to remember to call `utils.lowerCaseInsensitivePath` every time we access this map?

src/common/objUtils.ts
```diff
@@ -142,3 +145,19 @@ export function getCaseInsensitiveProperty<R>(
 
   return undefined;
 }
+
+/**
+ * Memoizes the single-parameter function.
+ */
+export function memoize<T, R>(fn: (arg: T) => R): (arg: T) => R {
```
digeff: Is this different than the lodash one?

src/adapter/breakpoints.ts
connor4312: Diego suggested the `MapUsingProjection` from v2. I'll grab that here.

roblourens: v1 sets the driveletter it in a case-insensitive way: `Debugger.setBreakpointByUrl\",\"params\":{\"urlRegex\":\"file:\\\\/\\\\/\\\\/[Cc]:\\\\/Users\\\\/roblou\\\\/code\\\\...`

roblourens: I think we need that, because the drive letter case in node depends on the windows environment. https://github.com/microsoft/vscode-node-debug2/issues/37

connor4312: Ah, good thought. I'll adjust that, that makes me much more comfortable :)

src/common/objUtils.ts
connor4312: Same functionality, yea

src/adapter/sources.ts
connor4312: Done :)

src/common/urlUtils.ts
```diff
@@ -23,6 +25,53 @@ export function getCaseSensitivePaths() {
...
+ * disambiguation.
+ */
+export function lowerCaseInsensitivePath(path: string) {
+  return isCaseSensitive ? path : path.toLowerCase();
+}
+
+/**
+ * Returns the path in its true case, correcting for case-insensitive file systems.
+ */
+export const truePathCasing = memoize(
```
roblourens: Not used anywhere yet right?

connor4312: It was used in the repo but we don't need to any more.

src/adapter/breakpoints.ts
connor4312: Added in the latest commit

PR #107, opened 2019-11-20
Fix/vscode slowness

connor4312:
This PR fixes most of the the slow startup time while debugging VS Code.
Like I talked about at various points, the issue is that we have an
instrumentation breakpoint that asks the debugger to pause before we
load any sourcemapped script, so that we can ensure breakpoints get set.
For the large VS Code project, this slows things down a great deal.

If the breakpoint predictor is on and working for this workspace, this
actually isn't necessary--and fortunately, it does work for VS Code.
This PR adds an option that disables the runtime breakpoint once the
breakpoint predictor finishes up. This drops the boot time of the VS
Code debug launch from ~30s to ~6s on my machine, and performance
optimizations to the BreakpointPredictor will drop that further.

Right now this is somewhat of a special case. The breakpoint predictor
_should_ work well on _most_ Node.js apps and some Electron apps. It
might be possible to make the predictor smarter in the future.

Fixes https://github.com/microsoft/vscode-pwa/issues/99

### DIFF
--- a/src/adapter/breakpoints.ts
+++ b/src/adapter/breakpoints.ts
@@ -466,7 +466,6 @@ export class BreakpointManager {
     this._totalBreakpointsCount += breakpoints.length;
     if (this._thread) {
       breakpoints.forEach(b => b.set(this._thread!));
-      this._updateSourceMapHandler(this._thread);
     }
     const dapBreakpoints = breakpoints.map(b => b.toProvisionalDap());
     this._breakpointsStatisticsCalculator.registerBreakpoints(dapBreakpoints);
--- a/src/adapter/breakpoints.ts
+++ b/src/adapter/breakpoints.ts
@@ -249,7 +249,11 @@ export class BreakpointManager {
   private _predictorDisabledForTest = false;
   private _breakpointsStatisticsCalculator = new BreakpointsStatisticsCalculator();
 
-  constructor(dap: Dap.Api, sourceContainer: SourceContainer) {
+  constructor(
+    dap: Dap.Api,
+    sourceContainer: SourceContainer,
+    private readonly pauseForSourceMaps: boolean,
+  ) {
     this._dap = dap;
     this._sourceContainer = sourceContainer;
 
@@ -275,6 +279,7 @@ export class BreakpointManager {
         remainPaused: result.some(r => r.some(l => l.columnNumber <= 1 && l.lineNumber <= 1)),
       };
     };
+
     if (sourceContainer.rootPath)
       this._breakpointsPredictor = new BreakpointsPredictor(
         sourceContainer.rootPath,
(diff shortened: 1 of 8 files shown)


src/adapter/threads.ts
```diff
@@ -868,37 +869,52 @@ export class Thread implements VariableStoreDelegate {
       (source as any)[kScriptsSymbol] = new Set();
     (source as any)[kScriptsSymbol].add(script);
 
-    if (!this._pauseOnSourceMapBreakpointId && event.sourceMapURL) {
```
connor4312: This is a somewhat unrelated thing I ran into when testing with an Electron apps. Electron apps are generally an 'attach' scenario, and we don't seem to hit `beforeScriptWithSourceMapExecution` when the scripts are already being executed, even if we're able to set the breakpoint. I adjusted this so that we kick off the source map loading (and eventually `_scriptWithSourceMapHandler`) when we get the loaded event, and just await that promise if/when we hit an instrumentation breakpoint for the script.

src/adapter/threads.ts
```diff
@@ -868,37 +869,52 @@ export class Thread implements VariableStoreDelegate {
...
-      const timeout = this._sourceContainer.sourceMapTimeouts().scriptPaused;
-      const sources = await Promise.race([
-        this._sourceContainer.waitForSourceMapSources(script.source),
-        delay(timeout),
-      ]);
-      if (sources && this._scriptWithSourceMapHandler) {
-        remainPaused = (await this._scriptWithSourceMapHandler(script, sources)).remainPaused;
-      }
+    if (!script) {
+      return false;
```
dgozman: You want to reset `this._pausedForSourceMapScriptId` here.

src/adapter/breakpoints.ts
```diff
@@ -449,8 +464,10 @@ export class BreakpointManager {
       await Promise.all(previous.map(b => b.remove()));
     }
     this._totalBreakpointsCount += breakpoints.length;
-    if (this._thread) breakpoints.forEach(b => b.set(this._thread!));
-    this._updateSourceMapHandler();
+    if (this._thread) {
+      breakpoints.forEach(b => b.set(this._thread!));
+      this._updateSourceMapHandler(this._thread);
```
dgozman: It seems like you don't need to call this method here anymore, since it does not depend on `this._totalBreakpointsCount`.

Note that it was added to avoid slowdown for the browser scenario where predictor is not always available, and user launches without any breakpoints often enough so we can skip extra pauses. I think this might still be a valid concern.

src/adapter/threads.ts
connor4312: Good catch!

src/adapter/breakpoints.ts
connor4312: Ah, yea, I see. I didn't mentally parse this code before. It seems to work fine without that line. It doesn't look like we ever _don't_ try to install the source map handler, but it's probably safe to do so when we know there aren't any breakpoints set. I'll look at that in a followup PR :)

connor4312:
Tests fixed in #118

PR #108, opened 2019-11-20
Ignore telemetry events when we await a once output message

digeff:
The telemetry changes broke the await p.dap.once('output') logic because it returned for the telemetry event instead of the output event. Filtering out the telemetry events in once fixes that issue


connor4312:
It looks like there's a couple more locations where this happens, e.g. manually adding `onceOnOutput` to the "logpoints basic" breakpoint test makes that pass. looking for `'output'` in the test file should show you any possible issues.

Alternately maybe we add a telemetry switch or filter at a lower level to avoid this altogether?

roblourens:
Yeah it seems better to just disable sending telemetry entirely

PR #112, opened 2019-11-21
Write logs when tests fail

roblourens:


src/test/logTestReporter.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    runner.on('pass', (test: TestWithLogfile) => {
+      if (LoggingReporter.alwaysDumpLogs) {
+        return this.dumpLogs(test);
+      }
+    });
+
+    runner.on('fail', (test: TestWithLogfile) => {
+      return this.dumpLogs(test);
+    });
```
connor4312: huh, that's neat. I never really used the `runner` API before, this is a useful pattern

src/test/logReporterUtils.ts
```diff
@@ -0,0 +1,14 @@
...
+import os from 'os';
+import mocha from 'mocha';
+import path from 'path';
+
+export interface TestWithLogfile extends mocha.Test {
+  logPath?: string;
+}
+
+export function getLogFileForTest(testTitlePath: string) {
+  return path.join(os.tmpdir(), `${testTitlePath.replace(/[\s*^<>\\\/|?:]+/g, '-')}.json`);
```
connor4312: Maybe just anything not alphanum

```suggestion
  return path.join(os.tmpdir(), `${testTitlePath.replace(/[^a-z0-9]/gi, '-')}.json`);
```

PR #114, opened 2019-11-21
fix: proper treatment of path tree across platforms, fix tests

connor4312:


connor4312:
Tests passed on a rebuild: https://dev.azure.com/vscode/VS%20Code%20debug%20adapters/_build/results?buildId=32396

PR #116, opened 2019-11-21
fix: close browser tree forcefully on exit

connor4312:
Fixes launching electron apps where closing the browser window is a no-op.


connor4312:
Test failure is an unrelated flake

PR #117, opened 2019-11-21
Fix pause reason for debugger statement

roblourens:
The only way to tell the difference between "debugger" and a step is to keep track of whether the user asked to step.


src/int-chrome/adapter.test.ts
```diff
@@ -85,14 +85,14 @@ suite('Chrome Debug Adapter etc', () => {
       },
     );
 
-    test('should stop on debugger statement in http://localhost', () => {
+    test.only('should stop on debugger statement in http://localhost', () => {
```
connor4312: rm

connor4312:
Tests issues are flakes or Windows problems with the new webpack tests. Unrelated to this PR and fixed in #118

PR #118, opened 2019-11-21
Fix/test fixes 1

connor4312:
Assorted fixes for a bunch of tests. I'll comment what/why inline in the PR,


src/adapter/threads.ts
```diff
@@ -572,18 +572,41 @@ export class Thread implements VariableStoreDelegate {
...
     });
   }
 
+  /**
+   * Gets a script ID if it exists, or waits to up maxTime. In rare cases we
+   * can get a request (like a stacktrace request) from DAP before Chrome
+   * finishes passing its sources over. We *should* normally know about all
+   * possible script IDs; this waits if we see one that we don't.
+   */
+  private async getScriptByIdOrWait(scriptId: string, maxTime: number = 500) {
```
connor4312: We were getting this sometimes in the stacktrace-related tests. I think this is a simple and safe solution.

src/adapter/variables.ts
```diff
@@ -215,7 +215,6 @@ export class VariableStore {
     this._referenceToVariables.clear();
     this._objectToReference.clear();
     this._referenceToObject.clear();
-    this._cdp.Runtime.releaseObjectGroup({ objectGroup: 'console' });
```
connor4312: This `clear()` is called on `executionContextsCleared`, which happens when pages are first loaded. The problem arises if we start setting variables before `executionContextsCleared` is received, to which we send back this instruction that releases anything we had set.

We didn't use `releaseObjectGroup` in the previous adapters, and I don't believe it's strictly necessary--any page navigation/refresh will automatically discard any objects set by the console.

src/common/logging/logger.ts
```diff
@@ -139,7 +139,11 @@ export class Logger implements ILogger, Disposable {
       // options. For instance, `cdp` adds the tags `cdp`, `cdp.send`, etc.
       this.tags = new Set(
         options.tags
-          .map(src => allLogTags.filter(t => t === src || t.startsWith(`${src}.`)))
+          .map(src =>
```
connor4312: Tag filtering just broke at some point, which failed tests.

src/targets/node/nodeSourcePathResolver.ts
```diff
@@ -25,7 +26,7 @@ export class NodeSourcePathResolver extends SourcePathResolverBase<IOptions> {
       return '';
     }
 
-    const withBase = path.resolve(this.options.basePath, this.applyPathOverrides(url));
```
connor4312: This was causing some cross-platform failures. `properResolve` joins and normalizes to the OS of the base path.

src/test/breakpoints/breakpointsTest.ts
```diff
@@ -255,7 +255,7 @@ describe('breakpoints', () => {
   });
 
   describe('logpoints', () => {
-    itIntegrates.skip('basic', async ({ r }) => {
+    itIntegrates('basic', async ({ r }) => {
```
connor4312: This was fixed by the change to variables.ts

src/test/extension/nodeConfigurationProvidersTests.ts
```diff
@@ -29,9 +29,9 @@ describe('NodeDebugConfigurationProvider', () => {
     EnvironmentVars.platform = process.platform;
   });
 
-  describe.skip('logging resolution', () => {
+  describe('logging resolution', () => {
```
connor4312: these tests just weren't updated when we moved from `logging` to `trace` as the log option

src/adapter/threads.ts
```diff
@@ -572,18 +572,41 @@ export class Thread implements VariableStoreDelegate {
...
       lineNumber,
       columnNumber,
       source: script.source
     });
   }
 
+  /**
+   * Gets a script ID if it exists, or waits to up maxTime. In rare cases we
+   * can get a request (like a stacktrace request) from DAP before Chrome
+   * finishes passing its sources over. We *should* normally know about all
```
roblourens: So you get a Paused event from Chrome that includes a script that you haven't gotten a scriptParsed event for yet? We should file a bug for that.

connor4312: It's actually

 1. We attach, initial events from CDP start coming, including script parsed events
 2. We hit a breakpoint and tell DAP
 3. DAP asks for the stacktrace, before we got `scriptParsed` events for all scripts in the stacktraces
 4. We respond and omit the `source` of scripts we haven't seen yet in stackframes.

This change delays step 4

roblourens: Right, but it seems like CDP shouldn't tell us that we hit a breakpoint until it tells us about the loaded scripts in the stack.

src/adapter/variables.ts
roblourens: I never used this before because I wasn't sure whether it's necessary and seemed fine without. Chrome devtools uses it so we might want to look into it more at some point?

connor4312: I think there might be a different issue at hand here. I added some more logging to see.

PR #119, opened 2019-11-22
Fix smartStep in ignored sourcemaps, add test

### LINKED ISSUE #109: smartStep incorrectly steps through "ignored" files
- Set a bp in a compiled file
- The sourcemap is temporarily ignored when hitting the BP
- But smartStep steps through this file

roblourens:
Fixes #109

### DIFF
--- a/src/adapter/sources.ts
+++ b/src/adapter/sources.ts
@@ -25,7 +25,7 @@ export interface UiLocation {
   source: Source;
 }
 
-function isUiLocation(loc: any): loc is UiLocation {
+function isUiLocation(loc: unknown): loc is UiLocation {
   return typeof (loc as UiLocation).lineNumber === 'number' &&
     typeof (loc as UiLocation).columnNumber === 'number' &&
     !!(loc as UiLocation).source;
@@ -245,7 +245,13 @@ export interface PreferredUiLocation extends UiLocation {
 }
 
 export enum UnmappedReason {
+  /** The map has been disabled temporarily, due to setting a breakpoint in a compiled script */
   MapDisabled,
+
+  /**
+   * The location cannot be sourcemapped, due to not having a sourcemap,
+   * failing to load the sourcemap, not having a mapping in the sourcemap, etc
+   */
   CannotMap
 }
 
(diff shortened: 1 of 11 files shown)


src/adapter/sources.ts
```diff
@@ -355,29 +369,29 @@ export class SourceContainer {
...
+    if (!isUiLocation(sourceMapUiLocation))
       return [];
 
     const r = this.getSourceMapUiLocations(sourceMapUiLocation);
     r.push(sourceMapUiLocation);
     return r;
   }
 
-  _sourceMappedUiLocation(uiLocation: UiLocation, map: SourceMap): UiLocation | undefined {
+  _sourceMappedUiLocation(uiLocation: UiLocation, map: SourceMap): UiLocation | UnmappedReason {
```
roblourens: This is awkward, let me know if you have a better idea

src/adapter/sources.ts
```diff
@@ -23,7 +23,13 @@ export interface UiLocation {
   lineNumber: number; // 1-based
   columnNumber: number;  // 1-based
   source: Source;
-};
+}
+
+function isUiLocation(loc: any): loc is UiLocation {
```
connor4312: `unknown` would be the more correct type to use here, TS will force you to narrow it safely :)

src/adapter/sources.ts
connor4312: I think this is fine. Throwing some typed error would be traditional OO thing to do but I don't mind this; it forces consumers to handle failures explicitly.

src/adapter/sources.ts
```diff
@@ -233,8 +239,14 @@ export class Source {
...
 
-export interface PreferredUILocation extends UiLocation {
+export interface PreferredUiLocation extends UiLocation {
   isMapped: boolean;
+  unmappedReason?: UnmappedReason;
+}
+
+export enum UnmappedReason {
+  MapDisabled,
+  CannotMap
```
connor4312: Docblocks may be useful here for future readers

src/adapter/sources.ts
roblourens: Yeah and since I need to keep the failure type around later on PreferredUiLocation, I wanted to just introduce it here.

PR #130, opened 2019-12-02
feat: retry nodejs attachment

### LINKED ISSUE #129: "attach" does not retry
"timeout" looks hooked up for the browser attacher, but not node. It should retry on a 10s timeout by default to match the current DA.

Example scenario, the "Search and Renderer" config launches vscode and tries to attach to the search process at the same time, but it needs to wait for things to start up.

connor4312:
Fixes #129

### DIFF
--- a/src/targets/browser/launcher.ts
+++ b/src/targets/browser/launcher.ts
@@ -134,7 +134,10 @@ export async function launch(
       const endpoint = await waitForWSEndpoint(browserProcess, cancellationToken);
       transport = await WebSocketTransport.create(endpoint, cancellationToken);
     } else {
-      const endpoint = await waitForDebuggerServerOnPort(suggestedPort, cancellationToken);
+      const endpoint = await retryGetWSEndpoint(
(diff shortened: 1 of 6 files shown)


src/targets/browser/launcher.ts
```diff
@@ -134,7 +134,10 @@ export async function launch(
       const endpoint = await waitForWSEndpoint(browserProcess, cancellationToken);
       transport = await WebSocketTransport.create(endpoint, cancellationToken);
     } else {
-      const endpoint = await waitForDebuggerServerOnPort(suggestedPort, cancellationToken);
+      const endpoint = await retryGetWSEndpoint(
+        `http://localhost:${suggestedPort}`,
+        cancellationToken,
```
roblourens: So this is canceled after the timeout? Where does that happen?

connor4312: It is `_boot` in the Binder creates a CancellationTokenSource for launch and attach requests, on which cancellation is requested after a certain amount of time.

PR #143, opened 2019-12-03
feat: add quick pick to open a loaded script

connor4312:
For https://github.com/microsoft/vscode/issues/85950. Works just like the v1 adapter!


connor4312:
It's less useful now that tree views can be filtered. This is a parity item with the v1 adapter, though it doesn't look like the v1 command is used very often.

roblourens:
I forgot we have this in v1, I don't think I've ever used it

weinand:
The plan was to remove the quickpick as soon as we get tree filtering.

connor4312:
Ok, I'll close this then, thanks for confirming!

PR #159, opened 2019-12-05
fix: source map override handling issues

connor4312:
- The previous adapter manually replaces[1] webRoot in the source map
   overrides. Do the same here for parity.
 - The `source-map` module normalizes URLs before they're given back to
   us, but the previous adapter didn't do this. Overrides that target
   prefixes like "webpack:///./*" don't work, because the extraneous
   "./" gets removed.

 1. https://github.com/microsoft/vscode-chrome-debug/blob/075b94d825294b5aa7992fee55125982ee44a669/src/chromeDebugAdapter.ts#L566

Fixes https://github.com/microsoft/vscode-pwa/issues/147
And most likely https://github.com/microsoft/vscode-pwa/issues/140

### DIFF
--- a/src/targets/sourceMapOverrides.ts
+++ b/src/targets/sourceMapOverrides.ts
@@ -6,6 +6,7 @@ import { forceForwardSlashes } from '../common/pathUtils';
 import { logger } from '../common/logging/logger';
 import { LogTag } from '../common/logging';
 import { escapeRegexSpecialChars } from '../common/stringUtils';
+import { URL } from 'url';
 
 // Patterns to match against various patterns:
 const capturingGroup = '*';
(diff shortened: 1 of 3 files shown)


src/targets/sourceMapOverrides.ts
```diff
@@ -44,6 +45,17 @@ export class SourceMapOverrides {
       const rightPattern = sourceMapOverrides[leftPattern];
       const entryStr = `"${leftPattern}": "${rightPattern}"`;
 
+      // Normalize patterns as URLs when we can, needed for back compat: the
+      // old adapter read the literal source URLS like `webpack:///./foo.ts`,
```
roblourens: Isn't this different from `webpack:///foo.ts` though? In the case I am vaguely remembering, this refers to a relative path in the project where the other format usually has an absolute path.

connor4312: I can't recall ever having seen webpack produce an absolute path that way and glancing through their docs I don't see an obvious way to tickle it into doing so... do you have any pointers to look into that? If not I can dive a little into their code in the morning.

roblourens: it's possible that it's like a webpack circa 2016 thing. I think it's fine to leave it alone if you think projects aren't doing that now.

connor4312: It looks like this [is configurable](https://webpack.js.org/configuration/output/#outputdevtoolmodulefilenametemplate) where people could put in a `[absolute-resource-path]`. Just searching through Github though it seems like that's generally used instead of the `webpack:///` uri, rather than in addition to. I'll probably merge this now and we can see if we get reports of any issues around it.

PR #162, opened 2019-12-06
feat: implement differential breakpoints

connor4312:
Fairly straightforward: when we get a breakpoint request, we compare
against the existing breakpoints in the file, and don't touch any that
remain.

I also made a modification to have us wait for the initial breakpoint
`set()` before returning the breakpoint response, so that points are
resolved from the outset. VS Code doesn't move around breakpoints after
the first response comes back, so this allows the column or line to be
adjusted as in the v1 adapters. (This lack of movement was causing https://github.com/microsoft/vscode-pwa/issues/156)

Fixes https://github.com/microsoft/vscode-pwa/issues/155 and https://github.com/microsoft/vscode-pwa/issues/156

### DIFF
--- a/src/adapter/breakpoints.ts
+++ b/src/adapter/breakpoints.ts
@@ -19,9 +19,26 @@ const localize = nls.loadMessageBundle();
 
 type LineColumn = { lineNumber: number; columnNumber: number }; // 1-based
 
+/**
+ * Differential result used internally in setBreakpoints.
+ */
+interface ISetBreakpointResult {
+  /**
+   * Breakpoints that previous existed which can be destroyed.
+   */
+  unbound: Breakpoint[];
+  /**
+   * Newly created breakpoints;
+   */
+  new: Breakpoint[];
+  /**
+   * All old and new breakpoints.
+   */
+  list: Breakpoint[];
+}
+
 export class Breakpoint {
   private _manager: BreakpointManager;
-  private _dapId: number;
   _source: Dap.Source;
   private _condition?: string;
   private _lineColumn: LineColumn;
@@ -34,12 +51,12 @@ export class Breakpoint {
 
   constructor(
     manager: BreakpointManager,
-    dapId: number,
+    public readonly dapId: number,
     source: Dap.Source,
     params: Dap.SourceBreakpoint,
   ) {
     this._manager = manager;
-    this._dapId = dapId;
+    this.dapId = dapId;
     this._source = source;
     this._lineColumn = { lineNumber: params.line, columnNumber: params.column || 1 };
     if (params.logMessage)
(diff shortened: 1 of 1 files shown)


src/adapter/breakpoints.ts
```diff
@@ -430,41 +477,72 @@ export class BreakpointManager {
...
-    if (this._thread) {
-      breakpoints.forEach(b => this._setBreakpoint(b));
+    this._totalBreakpointsCount -= result.unbound.length;
+    await Promise.all(result.unbound.map(b => b.remove()));
+
+    this._totalBreakpointsCount += result.new.length;
+
+    const thread = this._thread;
+    if (thread) {
+      await (this._launchBlocker = Promise.all([
```
dgozman: Do not await breakpoint.set() here. We report "provisionalDap" below to always notify once breakpoint is set. This way we avoid any races - always immediately report as "provisional", and then notify with real location once ready, unless the breakpoint was removed in the meantime. Also, 'notifyResolved' will be unnecessary.

src/adapter/breakpoints.ts
```diff
@@ -50,20 +67,31 @@ export class Breakpoint {
         : params.condition;
   }
 
-  toProvisionalDap(): Dap.Breakpoint {
+  async toProvisionalDap(): Promise<Dap.Breakpoint> {
```
dgozman: nit: toProvisionalDap means it's not resolved. Perhaps rename it toDap?

src/adapter/breakpoints.ts
connor4312: Could you describe a little more around the racing you're concerned about here? The only thing we do after this `await` is register info about the breakpoints for our own statistics, and then return the breakpoints to the UI. I think that if the UI has a race around breakpoints being set or updated before they come back from the debug adapter, then that will be a concern regardless of whether we await here or not since the call is already being made asynchronously and e.g. waits on the predictor.

src/adapter/breakpoints.ts
```diff
@@ -231,6 +262,20 @@ export class Breakpoint {
...
     await Promise.all(promises);
   }
+
+  /**
+   * Compares this breakpoint with the other. String comparison-style return:
+   *  - a negative number if this breakpoint is before the other one
+   *  - zero if they're the same location
+   *  - a positive number if this breakpoint is after the other one
+   */
+  public compare(other: Breakpoint) {
```
roblourens: Not obvious that it's only valid for breakpoints on the same Source, maybe it should assert that

connor4312:
Oops, I had this merged locally to test something and didn't mean to merge it so soon here. I'll address comments in master/followup prs

src/adapter/breakpoints.ts
dgozman: - While setting breakpoint in multiple scripts, it might already trigger in one of them. This will lead to "paused" event from adapter referencing breakpoint id which is not reported to UI just yet, because we didn't return from this method.
- Setting breakpoints might be slow, and it's nice to show "unbound" status to the user immediately until we actually verified breakpoint location.

What do you think?

PR #168, opened 2019-12-11
fix: ignore query parameters on parsed scripts

### LINKED ISSUE #166: Breakpoints not hit in scripts with a query param in their url
Simple repro: have index.html include any .js file, and add a bogus query param on the end, e.g.:
```html
<script src="test.js?abc=123" type="text/javascript"></script>
```

v1 can hit this just fine, pwa will never bind the bp. Particularly, it just looks like we are using the entire url to match our file on disk, where we should (generally?) ignore the query params, as I believe we already look at the content hashes anyway.

connor4312:
- 'eat' query parameters in the url pattern we send with breakpoints
 - remove params when creating absolute paths so that all mapping back
   and forth works.
 - add a test case :)

Fixes #166

### DIFF
--- a/src/common/urlUtils.ts
+++ b/src/common/urlUtils.ts
@@ -178,6 +178,13 @@ export function escapeForRegExp(s: string): string {
 }
 
 /**
+ * Remove a slash of any flavor from the end of the path
+ */
+export function stripTrailingSlash(aPath: string): string {
+  return aPath.replace(/\/$/, '').replace(/\\$/, '');
+}
+
+/**
  * If urlOrPath is a file URL, removes the 'file:///', adjusting for platform differences
  */
 export function fileUrlToAbsolutePath(urlOrPath: string): string | undefined {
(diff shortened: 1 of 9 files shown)


src/common/urlUtils.ts
```diff
@@ -217,10 +224,10 @@ export function isAbsolute(_path: string): boolean {
  */
 export function urlToRegex(aPath: string) {
   const absolutePath = fileUrlToAbsolutePath(aPath);
-
-  aPath = escapeRegexSpecialChars(aPath);
+  const eatQueryString = '(\\?.*)?';
+  aPath = escapeRegexSpecialChars(aPath) + eatQueryString;
   if (absolutePath) {
-    aPath += `|${escapeRegexSpecialChars(absolutePath)}`;
+    aPath += `|${escapeRegexSpecialChars(absolutePath)}${eatQueryString}`;
```
roblourens: This seems unnecessary because the regex will still match up to the `?`

connor4312: It seems that Chrome implicitly wraps the provided regex in `^$` when I tested it earlier

connor4312: actually, nevermind! I was probably running into something else at the time

Issue #169, opened 2019-12-12
A breakpoint update event can sometimes be sent before the response to the breakpoint request that created it


EricCornelson:
It looks like there is a race condition where a breakpoint update event can be sent out before the response to the "setBreakpointsRequest" that creates it. This causes an error in VS, as there is no breakpoint existing with the referenced id.

![image](https://user-images.githubusercontent.com/38218905/70671984-bc38c100-1c32-11ea-8201-ff772fc19752.png)

Logs (see line 212):
[vscode-debugadapter.log](https://github.com/microsoft/vscode-js-debug/files/3953228/vscode-debugadapter.log)

connor4312:
Thanks, Dmitry actually mentioned the possibility of this on the associated PR yesterday. I'll have a fix in tomorrow morning.

PR #171, opened 2019-12-12
fix: ensure bp set completes before updates or pauses

connor4312:
This is a followup to #169 and the discussion in #162. We now store
a deferred 'set complete' on the breakpoint, which we resolve in a
after returning the set response over DAP. We then guard updates around
this promise and skip sending them if we haven't sent the initial
`setBreakpoints` response yet (the response will contain their full
state with any updates that came in).

Also, if we see that we pause on a breakpoint that we haven't sent
back yet, we'll wait a moment for it to try to be set. I don't _think_
this will ever be hit in VS Code, since it sends all breakpoint
updates individually after the initial configuration, but it's there
regardless and seems to work well.

### DIFF
--- a/src/adapter/breakpoints.ts
+++ b/src/adapter/breakpoints.ts
@@ -15,6 +15,7 @@ import { BreakpointsStatisticsCalculator } from '../statistics/breakpointsStatis
 import { TelemetryEntityProperties } from '../telemetry/telemetryReporter';
 import { logger, assert } from '../common/logging/logger';
 import { LogTag } from '../common/logging';
+import { getDeferred, delay } from '../common/promiseUtil';
 
 const localize = nls.loadMessageBundle();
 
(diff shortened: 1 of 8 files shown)


src/adapter/threads.ts
```diff
@@ -1157,8 +1159,24 @@ export class Thread implements IVariableStoreDelegate {
     slot(output);
   }
 
-  _onThreadPaused(details: IPausedDetails) {
+  async _onThreadPaused(details: IPausedDetails) {
```
EricCornelson: Doesn't look like `_onThreadPaused` is being awaited anywhere now that it's async. Is this intended?

src/adapter/breakpoints.ts
```diff
@@ -553,30 +583,38 @@ export class BreakpointManager {
...
       ]));
     }
 
-    const dapBreakpoints = await Promise.all(result.list.map(b => b.toProvisionalDap()));
+    const dapBreakpoints = await Promise.all(result.list.map(b => b.toDap()));
     this._breakpointsStatisticsCalculator.registerBreakpoints(dapBreakpoints);
+
+    // In the next task after we send the response to the adapter, mark the
+    // breakpoints as having been set.
+    delay(0).then(() => result.new.forEach(bp => bp.markSetCompleted()));
```
EricCornelson: Looks like another place where a "post-event handler" type thing would be useful.

dgozman:
I feel like this is piling up complexity to fix an issue introduced by other PRs. I suggest to revisit the whole idea and always send response to `setBreakpoints` before actually trying to set them. See my comments in #162 with reasoning. This issue looks like another reason to do that.

connor4312:
Yea, I agree. I'd like to keep this fix in-place for the moment as it avoids a fairly common and confusing issue. Once https://github.com/microsoft/vscode/issues/86749 is resolved I will look at reverting this behavior to simplify things.

Issue #170, opened 2019-12-12
Not able to hit the breakpoints set on TS files which are present inside the node_modules folder


gauravmahto-at:
VSCode Version: Version: 1.40.2 (user setup)
OS Version: Win10 Enterprise x64 1903
Steps to Reproduce:

Have node_modules JS sources with mapping to TS sources present inside the node_modules folder.
Apply breakpoints in the TS file present inside node_modules folder
Breakpoints are NOT properly applied as the VSCode informs that the breakpoints are not bound.
Try to execute the code path which should hit the breakpoints
None of the breakpoints is hit
Does this issue occur when all extensions are disabled?: Yes

Note: After enabling the 'trace' option, I can see that the files are properly mapped and so does the breakpoints. Also, if I use the Chrome Debugger the breakpoints get properly hit.

Attached is the trace log file.
[vscode-debugadapter.zip](https://github.com/microsoft/vscode-js-debug/files/3953732/vscode-debugadapter.zip)

This is related to https://github.com/microsoft/vscode/issues/86363

gauravmahto-at:
@roblourens Please find the attached log files.

connor4312:
There were a few things we needed to tweak to allow this. I put in a PR which, once merged, should appear on the next daily build (8 AM PST). https://github.com/microsoft/vscode-js-debug/pull/172

To use this, you'll just need to tell the debug adapter that it's OK to match against source files in your node modules by adding this to your launch.json:

```json
"resolveSourceMapLocations": ["**/*"]
```

gauravmahto-at:
@connor4312 I'll give it a try once the new build is available.
Also, `"resolveSourceMapLocations": ["**/*"]` must be added into the launch configuration?

connor4312:
Yep, that's right.

gauravmahto-at:
@connor4312 I tried the latest build but now I'm NOT even to hit the applied breakpoints in files outside the node_modules as well as inside the node_modues folder? Sometimes, I see 'unbound breakpoints' message when I try to apply the breakpoints and sometimes not.
I did add the `"resolveSourceMapLocations": ["**/*"]` configuration.

connor4312:
The next build won't run until 8 AM today (another two hours). Please give that a spin, and if things still don't work for you then another drop of the trace logs would be much appreciated! 🙂

gauravmahto:
Thank you! Will grab the next build once available and let you know.

gauravmahto-at:
[...]

Issue #164, opened 2019-12-09
Provide a prompt/notification if the BP predictor trips on a large directory


connor4312:
Right now it can just hang for a while or eventually run out of file descriptors since it scans everything in parallel without backpressure. We could trigger a prompt to users asking whether they want to continue scanning large directories.

PR #175, opened 2019-12-16
feat: allow configuring bp predict with outFiles, warn on long predict

connor4312:
> outFiles 2: electric bugaloo!

With several issues including #170 and #164, it's clear that the
breakpoint predictor will cause problems for some users and there needs
to be a way to configure the grepping behavior. This brings back the
"outFiles" option which, like v1, can be pointed to compiled sources to
direct it what to discover. We still default to the entire workspace,
but have a prompt that appears if we see prediction takes more than 10s:


src/ui/longPredictionUI.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  const doc = await vscode.workspace.openTextDocument(
+    join(workspaceFolder.uri.fsPath, '.vscode', 'launch.json'),
+  );
+  await vscode.window.showTextDocument(doc);
+}
+
+export function registerLongBreakpointUI(context: vscode.ExtensionContext) {
+  context.subscriptions.push(
+    vscode.debug.onDidReceiveDebugSessionCustomEvent(event =>
```
roblourens: Shouldn't this be filtering on the event type somewhere?

src/build/strings.ts
```diff
@@ -156,6 +156,14 @@ const strings = {
   'node.stopOnEntry.description': 'Automatically stop program after launch.',
   'node.timeout.description':
     'Retry for this number of milliseconds to connect to Node.js. Default is 10000 ms.',
+
+  'configuration.warnOnLongPrediction':
+    'Whether a loading prompt should be shown if breakpoint prediction takes a while.',
+  'longPredictionWarning.message':
+    "It's taking a while to configure your breakpoints. You can speed this up by updating the 'outFiles' in your launch.json.",
```
roblourens: Is `outFiles` used right now?

src/ui/longPredictionUI.ts
connor4312: Oof, yea, good catch!

src/build/strings.ts
connor4312: It is with this change; we now use that as the pattern grep for, rather than the strategy where we looked for all .js files except for node_modules wherever we had a breakpoint set.

roblourens: Ok I missed that

roblourens: We might also want to say that if outFiles is set, we don't set the instrumentation breakpoint, could help with slowness in that scenario

connor4312: Yea. That could help speed up the sliver of intersection between browsers serving filesystem paths (where vscode lies)  pretty dramatically.

Issue #190, opened 2019-12-16
Debug Console: Completion preceded by hidden properties with __


octref:
Version: 1.42.0-insider
Commit: 2d2ac8f75b539d505443ac0733b196851c0b218d
Date: 2019-12-16T12:08:26.094Z
Electron: 6.1.6
Chrome: 76.0.3809.146
Node.js: 12.4.0
V8: 7.6.303.31-electron.0
OS: Darwin x64 18.6.0

Repro:

- Have a file `test.js` with following content, breakpoint on the second line:

```js
const a = 5
const b = 10
```

- Type `a.` to complete in editor

![image](https://user-images.githubusercontent.com/4033249/70952830-59716c00-201c-11ea-8784-677bc753244e.png)

- Type `a.` to complete in debug console

![image](https://user-images.githubusercontent.com/4033249/70952883-7f970c00-201c-11ea-9fd0-8847b34fa288.png)

- Type `a.` in Chrome

![image](https://user-images.githubusercontent.com/4033249/70953747-3d22fe80-201f-11ea-9ed4-8a2b2c5d5cfd.png)

PR #193, opened 2019-12-18
feat: revamp completion

connor4312:
Completion previously was pretty basic. This does more work to get it
better data and correct type...

1. I made a "template function" helper. It's very scary, but in the
   debug adapter we heavily rely on sending non-trivial chunks of JS
   over to the adapter, and being able to write these in TypeScript with
   all the associated checking and linter is, in my opinion, preferable.
2. Did some work on the property walker (template) function to have it
   try to infer class types, properties, etc. correctly. Previously it
   only ever emitted `property` and `function`.
3. Better detection and handling of element kinds, particularly handling
   arrays correctly, which we didn't really do before.
4. Allow referencing locals declared elsewhere in the REPL expression,
   useful for writing more complex statements.

Fixes https://github.com/microsoft/vscode/issues/87123
Fixes https://github.com/microsoft/vscode-js-debug/issues/190
Fixes https://github.com/microsoft/vscode-js-debug/issues/186

### DIFF
--- a/src/adapter/templates/index.ts
+++ b/src/adapter/templates/index.ts
@@ -3,6 +3,7 @@
  *--------------------------------------------------------*/
 
 import * as ts from 'typescript';
+import Cdp from '../../cdp/api';
 
 /**
  * Creates a template for the given function that replaces its arguments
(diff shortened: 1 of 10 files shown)


src/adapter/templates/index.ts
```diff
@@ -0,0 +1,102 @@
...
+ * raw JavaScript functions in strings.
+ *
+ * Example:
+ *
+ * ```js
+ * const multiply = (a, b) => {
+ *   return a * b;
+ * };
+ * const template = templateFunction(multiply);
+ * console.log(multiple('42', 'foo()));
```
roblourens: Should be calling `template(...`?

src/adapter/templates/index.ts
```diff
@@ -0,0 +1,102 @@
...
+ * ```
+ *
+ * Outputs:
+ *
+ * ```
+ * (() => {
+ *   let __arg0 = 42;
+ *   let __arg1 = foo();
```
roblourens: I don't think you have to inline the arguments, I think it's possible to pass a function and then an array of arguments to call it with in that cdp API.

connor4312: One big drawback for this particular scenario there is that callFunction doesn't have access to the [command line API](https://github.com/GoogleChrome/devtools-docs/blob/master/docs/commandline-api.md), which in the REPL would be useful to have.

However, callFunction should be useful for most other scenarios :) I'll push up the (much simpler wrapper) I just wrote for making callFunction's so that we can use it later on.

PR #195, opened 2019-12-19
Evaluate request should return a real evaluate response,

### LINKED ISSUE #184: Avoid the newline and arrow in debug console
Now that vscode shows a chevron for the in/out calls.
image

roblourens:
instead of sending all results as OutputEvents.
Fix #184

### DIFF
--- a/src/dap/connection.ts
+++ b/src/dap/connection.ts
@@ -9,7 +9,7 @@ import { TelemetryReporter, RawTelemetryReporterToDap } from '../telemetry/telem
 import { installUnhandledErrorReporter } from '../telemetry/unhandledErrorReporter';
 import { logger, assert } from '../common/logging/logger';
 import { LogTag } from '../common/logging';
-import { isDapError } from './errors';
+import { isDapError, isExternalError } from './errors';
 
 export type Message = (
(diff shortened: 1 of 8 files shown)


roblourens:
Also filed https://github.com/microsoft/vscode/issues/87443 for further discussion with how errors should work here.

src/dap/connection.ts
```diff
@@ -220,13 +220,16 @@ export default class Connection {
         this._telemetryReporter?.reportSuccess(msg.command, receivedTime);
       } catch (e) {
         console.error(e);
+        const format = isExternalError(e)
+          ? e.message
+          : `Error processing ${msg.command}: ${e.stack || e.message}`;
```
connor4312: Won't this end up showing the stacktrace of the ExternalError  error instance within `js-debug`, rather than the stacktrace of the originating code?

roblourens: I think you are looking at it backwards

roblourens: Right now we don't get the stack at all -

image

which is suboptimal but I figured I'd wait for microsoft/vscode#87443 to do more, I think most likely we'd end up getting rid of the ExternalError code.

PR #200, opened 2019-12-20
feat: initial implementation of hit condition breakpoints

connor4312:
> **Note:** this is stacked on #174. That PR should be fast-forwards-merged first, and then this should be changed to point at master.

First of all, apologies for the churn. While this change is relatively
simple, I realized that hit condition breakpoints further diverges the
user-defined breakpoints and the managed 'entrypoint' breakpoints (as
of #174), where we don't have hit conditions or DAP representations.

Therefore, there's now a base "Breakpoint" class which implements the
purely CDP-side of things, which a `UserDefinedBreakpoint` extends with
metadata around DAP, and the `EntryBreakpoint` extends in a minimal way.
After doing the split there was almost no changes the existing code
around the breakpoints, which is good evidence that we are indeed
dealing with two different types of things!

That done, there's a new `HitBreakpoint` class whose parsing logic is
copied from the existing debug adapters. The UserDefinedBreakpoint
optionally includes a HitBreakpoint, and we check on these when we hit
a location to see whether we should automatically continue, or not.

This works in my quick manual tests. Haven't had a chance to write unit
tests for this, but I wanted to get it in PR before I head out for the
holidays.

### DIFF
--- a/src/adapter/breakpoints/hitCondition.ts
+++ b/src/adapter/breakpoints/hitCondition.ts
@@ -1,0 +1,49 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+import { ProtocolError, invalidHitCondition } from '../../dap/errors';
+
+/**
+ * Regex used to match hit conditions. It matches the operator in group 1 and
+ * the constant in group 2.
+ */
+const hitConditionRe = /^(>|>=|={1,3}|<|<=|%)?\s*([0-9]+)$/;
+
+/**
+ * A hit condition breakpoint encapsulates the handling of breakpoints hit on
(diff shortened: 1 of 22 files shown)


src/adapter/breakpoints/hitCondition.ts
```diff
@@ -0,0 +1,49 @@
...
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+import { ProtocolError, invalidHitCondition } from '../../dap/errors';
+
+/**
+ * Regex used to match hit conditions. It matches the operator in group 1 and
+ * the constant in group 2.
+ */
+const hitConditionRe = /^(>|>=|={1,3}|<|<=|%)?\s*([0-9]+)$/;
```
connor4312: The one change I made here was allowing 1 to 3 `=`, since I type == or === by force of habit.

PR #202, opened 2020-01-02
Fix angular in VS

digeff:
Due to some limitations we can't always figure out the right webRoot in VS for Angular projects. This fix worksaround that fact by trying both <webRoot> and <webRoot>/../ClientApp at the same time.

(This fix was ported from v1: https://github.com/microsoft/vscode-chrome-debug-core/blob/87d2c194179033e38d68308027c1d1dbbc43c454/src/sourceMaps/sourceMapUtils.ts#L118 )

### DIFF
--- a/src/targets/browser/browserPathResolver.ts
+++ b/src/targets/browser/browserPathResolver.ts
@@ -12,6 +12,7 @@ import { properResolve } from '../../common/pathUtils';
 interface IOptions extends ISourcePathResolverOptions {
   baseUrl?: string;
   webRoot?: string;
+  clientID: string | undefined;
 }
 
 export class BrowserSourcePathResolver extends SourcePathResolverBase<IOptions> {
@@ -59,7 +60,11 @@ export class BrowserSourcePathResolver extends SourcePathResolverBase<IOptions> 
 
       // Prefixing ../ClientApp is a workaround for a bug in ASP.NET debugging in VisualStudio because the wwwroot is not properly configured
       const clientAppPath = properResolve(webRoot, '..', 'ClientApp', unmappedPath);
-      if (!(await fsUtils.exists(webRootPath)) && (await fsUtils.exists(clientAppPath))) {
+      if (
+        this.options.clientID === 'visualstudio' &&
+        !(await fsUtils.exists(webRootPath)) &&
+        (await fsUtils.exists(clientAppPath))
+      ) {
         return clientAppPath;
       } else {
         return webRootPath;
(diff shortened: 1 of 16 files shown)


src/targets/browser/browserPathResolver.ts
```diff
@@ -54,7 +55,15 @@ export class BrowserSourcePathResolver extends SourcePathResolverBase<IOptions>
 
     const unmappedPath = this.sourceMapOverrides.apply(url);
     if (unmappedPath !== url) {
-      return properResolve(webRoot, unmappedPath);
+      const webRootPath = properResolve(webRoot, unmappedPath);
+
+      // Prefixing ../ClientApp is a workaround for a bug in ASP.NET debugging in VisualStudio because the wwwroot is not properly configured
+      const clientAppPath = properResolve(webRoot, '..', 'ClientApp', unmappedPath);
+      if (!fs.existsSync(webRootPath) && fs.existsSync(clientAppPath)) {
```
roblourens: Can it be scoped to VS only, and possibly made async?

The original fix also had `leftPattern === 'webpack:///./*'` I dunno if you want to keep that and just do this inside the sourceMapPathOverrides code.

digeff: Scoped to VS only. Made async. Added the webkpack prefix check.
sourceMapPathOverrides seems to be for all targets, I think it makes sense to leave this check on browserPathResolver which is browser specific (It doens't make sense to talk about webroot for the other targets).

connor4312:
Can you add a unit test for this? This isn't a case we're likely to consistently test manually when touching this code 🙂

src/test/browser/browserPathResolverTest.ts
```diff
@@ -0,0 +1,61 @@
...
+          throw Error('Unknown path');
+      }
+    });
+  });
+
+  [
+    ['visualstudio', 'ClientApp'],
+    ['vscode', 'wwwroot'],
+  ].forEach(([client, folder]) => {
+    it.only(`returns ${folder} for ${client} if the webroot path doesn't exist and the modified path does`, async () => {
```
connor4312: rm `only`

src/test/browser/browserPathResolverTest.ts
```diff
@@ -0,0 +1,61 @@
...
+      const webRoot = 'C:\\Users\\user\\Source\\Repos\\Angular Project\\wwwroot';
+
+      const defaultSourceMapOverrides = {
+        'meteor://💻app/*': '/*',
+        'webpack://?:*/*': '*',
+        'webpack:///./~/*': '/node_modules/*',
+      };
```
connor4312: You can import the defaults from configuration.ts if you want

Issue #201, opened 2020-01-02
stopOnEntry does not work when 'program' is omitted


jcrben:
VSCode version:

```
Version: 1.42.0-insider
Commit: 7c0095ee2d064033fc13184127a9adc603454729
Date: 2019-12-20T07:55:38.353Z
Electron: 6.1.6
Chrome: 76.0.3809.146
Node.js: 12.4.0
V8: 7.6.303.31-electron.0
OS: Darwin x64 18.6.0
```

Launch.json:

```
    {
      "stopOnEntry": true,
      "name": "node relativeFile",
      "type": "node",
      "request": "launch",
      "args": ["${workspaceFolder}/${relativeFile}"],
      "cwd": "${workspaceFolder}/backend"
    },
```

Script file:

```
const fs = require('fs');
console.log('hmm');
```

This is adapted from a ts-node launch json where I have an additional `"runtimeArgs": ["-r", "ts-node/register"],`. Of course does this not run, but it's the next step.

I did notice that this does work:
```
    {
      "name": "node launch Program",
      "type": "node",
      "request": "launch",
      "program": "${file}"
    },
```

It just hangs:
![image](https://user-images.githubusercontent.com/5614134/71654182-cce1d180-2cdc-11ea-8c79-36971633aa48.png)

No output in the log:
![image](https://user-images.githubusercontent.com/5614134/71654196-e1be6500-2cdc-11ea-811a-c65cb8f36fcb.png)

So it seems like passing the file in as an argument rather than program is not working? But the same command does work on the CLI, e.g. `/Users/bencreasy/.local/nodenv/shims/node  /Users/bencreasy/code/stock-analysis/backend/src/db/migrations/migrate-csv-all-trading.js`

Works in CLI:
![image](https://user-images.githubusercontent.com/5614134/71654198-edaa2700-2cdc-11ea-995c-1246b4bcee2e.png)

I was originally motivated to try this due to "Step Over" in the current debugger continuously stepping in.

jcrben:
I ended up pulling this down and running it with the Extension Development Host. I set a breakpoint at
https://github.com/microsoft/vscode-js-debug/blob/842b32f90c54313e922719795f0ea1c286189b08/src/ui/sessionManager.ts#L126 and stepped around.

I haven't figured out the issue yet - eventually the call stack is empty but it's still running.
![image](https://user-images.githubusercontent.com/5614134/71659943-dc214900-2cf5-11ea-8737-9963d889de16.png)

[...]

PR #203, opened 2020-01-03
fix: allow omitting "program" from launch.json

jcrben:
Closes https://github.com/microsoft/vscode-js-debug/issues/201

I also noticed `gulp build --nightly` does not seem to be a task; `gulp package` worked for me.


connor4312:
Thanks for your contribution!

Issue #215, opened 2020-01-08
Not able to set a breakpoint in React app


EricCornelson:
I'm currently not able to hit any breakpoints in a basic React app created with `create-react-app` on Windows (haven't tested Linux yet)

Repro steps:
1. Create a new react app using `create-react-app`
2. Start the dev server with `yarn start`
3. Create a basic launch config
4. Set a breakpoint in App.js, any line
5. Launch Chrome, and see that the breakpoint set in App.js is never bound or hit

connor4312:
I don't seem to repro this running the latest nightly or OSS. What version are you on?

![image](https://user-images.githubusercontent.com/2230985/71999922-1b203480-31f7-11ea-9ebe-d4176c29ccaa.png)

EricCornelson:
Happening for me on latest master (b66fe7de628bc5b07795964d9b602edfb7156832). Have you tried on Windows?

connor4312:
I'll give it a spin.

connor4312:
Found the issue, will create a fix and trigger a new nightly build momentarily

connor4312:
To verify:

 1. `npx create-react-app my-app`
 2. Lunch config:

```json
    {
      "trace": true,
      "type": "pwa-chrome",
      "request": "launch",
      "name": "Launch Chrome against localhost",
      "url": "http://localhost:3000",
      "webRoot": "${workspaceFolder}"
    }
```

 3. Start the server and verify you can hit a breakpoint in App.tsx

PR #216, opened 2020-01-08
fix: absolute path handling in source map

connor4312:
We previously allowed source map file URIs, but not full absolute paths. It appears that create-react-app on Windows specifically creates sourcemaps containing absolute paths (with forward slashes instead of back slashes). This PR adds handling and a test case for those.

Fixes https://github.com/microsoft/vscode-js-debug/issues/215


EricCornelson:
LGTM

Issue #199, opened 2019-12-20
Don't start a debug session in VS code on initial terminal open


connor4312:
Right now you're in a weird state when the terminal is open, where you're apparently in a debug session with the red action bar and debug overlay, but no running program. We should not trigger a debug session in VS code until a child attaches.

connor4312:
To verify, open the JavaScript Debug Terminal, see that VS Code doesn't enter debug mode. Run a program like `node test.js`, observe that it enters debug mode and exits debug mode when the process exits, keeping the terminal open.

PR #217, opened 2020-01-08
feat: add code lens for npm scripts

### LINKED ISSUE #196: Code lens for scripts in package.json
Now that we have the superpower to debug any npm script without additional configuration, I want to look at adding a code lens to the package.json that allow `Run | Debug`.

This was previously explored by @egamma in https://github.com/microsoft/vscode/issues/51613, which received feedback that having a code lens over every script became too overloaded. The implementation there ended with having a run | debug in the script hover, but this is less discoverable (I didn't know it existed until Erich showed me!)

I'd like to try placing a single code lens over the entire scripts section, which would then open a quick pick to allow the user to select which script they want.

```
  Run | Debug
  "scripts": {
    "start": "node server.js",
    // ...
```

This would, of course, be togglable with a user option. We could also prompt the user to save a launch config if we see they're repeatedly running the same script, for ease of access.

connor4312:
Fixes #196. Adds a code lens to debug an npm script. This adapter
currently doesn't support `nodebug` very well, and there are changes
coming to the terminal in #199, so I didn't implement a "Run" action
yet, but that should be straightforward once those are resolved.

By default the code lens appears atop the "script" section of the
package.json from where it opens a quick-pick, but it can also be
shown above individual scripts or hidden.

This also includes a slight modification to the contrib generator for
configuration to add some type safety.

Kapture 2020-01-08 at 11 30 49

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4201,7 +4201,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "aproba": {
           "version": "1.2.0",
@@ -4222,12 +4223,14 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
           "dev": true,
+          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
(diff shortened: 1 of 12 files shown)


karlhorky:
@connor4312 not sure if this only got rolled out in VS Code May 2020 (1.46), but I just noticed it now.

Is there any way to disable this?

karlhorky:
Ah I just got an answer over here: https://github.com/microsoft/vscode/issues/101107

Looks like it's:

```json
"debug.javascript.codelens.npmScripts": "never"
```

PR #218, opened 2020-01-08
Added support to launch Chrome unelevated in VS

digeff:
* Added support to launch Chrome unelevated in VS
* Fixed enqueueRequest because it was using the wrong seq number

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -404,7 +404,7 @@ export interface IChromeLaunchConfiguration extends IChromeBaseConfiguration {
   /**
    * The debug adapter is running elevated. Launch Chrome unelevated to avoid the security restrictions of running Chrome elevated
    */
-  launchUnelevatedFlag?: boolean;
+  launchUnelevated?: boolean;
 }
 
 /**
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -400,6 +400,11 @@ export interface IChromeLaunchConfiguration extends IChromeBaseConfiguration {
    * with your default user profile.
    */
   userDataDir: string | boolean;
+
+  /**
+   * The debug adapter is running elevated. Launch Chrome unelevated to avoid the security restrictions of running Chrome elevated
+   */
+  launchUnelevatedFlag?: boolean;
 }
 
 /**
--- a/src/targets/browser/browserProcess.ts
+++ b/src/targets/browser/browserProcess.ts
@@ -13,7 +13,7 @@ type stdioType = [
 
 export interface IBrowserProcess {
   readonly stdio: stdioType;
-  readonly pid: number;
+  readonly pid: number | undefined;
   readonly stderr: NodeJS.ReadableStream;
   readonly stdout: NodeJS.ReadableStream;
   removeListener(eventName: 'exit', onExit: (code: number) => void): void;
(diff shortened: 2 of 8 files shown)


src/configuration.ts
```diff
@@ -405,6 +405,11 @@ export interface IChromeLaunchConfiguration extends IChromeBaseConfiguration {
    * with your default user profile.
    */
   userDataDir: string | boolean;
+
+  /**
+   * The debug adapter is running elevated. Launch Chrome unelevated to avoid the security restrictions of running Chrome elevated
+   */
+  launchUnelevatedFlag?: boolean;
```
connor4312: nit: `flag` in the naming is redundant since it's typed as a boolean, `launchUnelevated?: boolean` is good

src/targets/browser/browserProcess.ts
```diff
@@ -0,0 +1,90 @@
...
+  removeListener(eventName: 'exit', onExit: (code: number) => void): void;
+  removeListener(eventName: 'error', onError: (error: Error) => void): void;
+  on(eventName: string, listener: Function): void;
+}
+
+export class BrowserProcessByPid implements IBrowserProcess {
+  public constructor(public readonly pid: number) {}
+
+  // TODO: Figure out how to use a WeakSet here so we don't have a memory leak (At the moment we don't use it because it can't be iterated)
+  private static instancesToPoll = new Set<BrowserProcessByPid>();
```
connor4312: You could use or abstract logic from the `TerminalProcess` in `program.ts` that we use for launching Node programs, it implements similar polling logic already.

src/targets/browser/browserProcess.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  private static pollInterval: NodeJS.Timeout | null = null;
+  private exitListeners = new Set<Function>();
+
+  public get stdio(): stdioType {
```
connor4312: Assuming you added this in the interface to avoid type errors creating the `PipeTransport` down below, I wonder if `launch()` could be restructured to avoid needing to to this?

src/configuration.ts
digeff: Fixed

src/targets/browser/browserProcess.ts
digeff: Removed this code

src/targets/browser/browserProcess.ts
digeff: I can't think of any easy way, if you come up with something please let me know

src/targets/browser/unelevatedChome.ts
```diff
@@ -0,0 +1,18 @@
...
+  dap: Dap.Api,
+  chromePath: string,
+  chromeArgs: string[],
+): Promise<number> {
+  const response: any = (dap as any).launchUnelevatedRequest({
+    process: chromePath,
+    args: chromeArgs,
+  });
+
+  return (await promiseTimeout(response, 10000 /* 10 seconds */)).processId;
```
connor4312: You could pass the `CancellationToken` through here instead and use `timeoutPromise`

digeff: Done

PR #219, opened 2020-01-09
Fix/jest scenarios

### LINKED ISSUE #214: Should it debug typescript files ?
I try too debug my jest test's but its writen in typescript and with this js debugger it didnt run.
Just stays with 2,9mb ram in the taskmanager and didnt say debugger atteched.

Without this extension the default debugger works. (no changes at the launch config)

Current launch config:
```

        {
            "name": "Debug Jest Tests",
            "type": "node",
            "request": "launch",
            "runtimeArgs": [
                "--inspect-brk",
                "${workspaceRoot}/node_modules/jest/bin/jest.js",
                "-i",
                "--runInBand"
            ],
            "console": "integratedTerminal",
            "internalConsoleOptions": "neverOpen",
            "port": 9229
        }
```

connor4312:
1. fix: preserve source map handler even without pauses

    Previously removing this would prevent source maps getting resolved at runtime at all.

2. fix: disable instrumentation bp in node by default

    It looks like in Jest the instrumentation BP is set, but because everything is on disk and we have the manually added entrypoint breakpoints, this is not needed.

3. feat: move content hashing to a child process 

    I noticed in profiles that this was actually a bottleneck for Jest tests
    (in #214) when running as a cluster. I wanted to use a worker thread
    for it, but it looks like there's an issue in vscode preventing that[1]
    for the moment.
    
    This cuts the time-to-first-breakpoint in half for the jest tests, which
    is fairly nice. The child process is killed after 30 seconds of
    inactivity. I may do an algorithmic optimization pass on the hash in
    the future. In particular, Node/V8 now has native bigint support, which
    is almost certainly faster than the `long` library.
    
    1. microsoft/vscode#88386

Fixes #214


src/nodeDebugConfigurationProvider.ts
```diff
@@ -93,6 +93,22 @@ export class NodeDebugConfigurationProvider extends BaseConfigurationProvider<An
...
         config.internalConsoleOptions = 'neverOpen';
       }
+
+      // remove manual --inspect flags, which are no longer needed and interfere
+      if (config.runtimeArgs) {
+        const resolved: string[] = [];
+        for (const arg of config.runtimeArgs) {
+          if (arg.startsWith('--inspect-brk')) {
+            config.stopOnEntry = config.stopOnEntry ?? true;
+          } else if (arg.startsWith('--inspect')) {
```
roblourens: Maybe make these exact matches rather than `startsWith` because people might be matching passing flags to their own script or something like that.

src/common/hash/index.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  instance = fork(join(__dirname, 'hash.js'), [], { env: {}, silent: true });
+  instance.setMaxListeners(Infinity);
+  return instance;
+};
+
+const send = (req: HashRequest): Promise<string | undefined> => {
+  const cp = create();
+  cleanup();
```
EricCornelson: Does this call just start the 30s timer for the cleanup callback? It's not totally clear from the name alone.

src/nodeDebugConfigurationProvider.ts
connor4312: The CLI allows `--inspect[=adddress:port]` but I can regex that :)

src/common/hash/index.ts
connor4312: Yep, `debouce` will call the wrapped function 30 seconds after calls to it stop. [More explanation on debouncing](https://medium.com/@jamischarles/what-is-debouncing-2505c0648ff1)

EricCornelson: Got it. I'm familiar with debouncing, it just wasn't clear that calling `cleanup()` wasn't actually executing the cleanup method, but rather _starting the debounce timer for the cleanup method_.

connor4312: Ah, gotcha, I'll clarify that variable name on `master`!

Issue #223, opened 2020-01-12
Step into worker broken on macOS


roblourens:
- Debug the web-worker-ts demo
- Break on the `new Worker` line, press step in
- It doesn't break in the worker

Works on linux. This is why `cross thread constructor source map` is failing on mac.

On Linux there is an extra Debugger.paused event fired before the worker is loaded, which has the `asyncCallStackTraceId` that we need to pause when the worker is loaded. This isn't sent on mac. Chrome bug right, or do you see anything else?

[ww-mac.txt](https://github.com/microsoft/vscode-js-debug/files/4051124/ww-mac.txt)
[ww-linux.txt](https://github.com/microsoft/vscode-js-debug/files/4051125/ww-linux.txt)

connor4312:
Actually, it is breaking. But it looks like we depend on [`asyncCallStackTraceId`](https://chromedevtools.github.io/devtools-protocol/tot/Debugger#event-paused) in order to know whether the next pause is related to a previous async step, which is present on Linux and Windows but marked deprecated and removed in OSX. Looking into the proper alternative, docs are sparse.

connor4312:
I looked through it a bit. It doesn't seem like there were any public announcements or discussion on the deprecation. I have some ideas for alternative approaches, but I was mostly just throwing things and seeing what works. I pinged Dmitry to see if he has any recommendations.

roblourens:
One thing I noticed is that in Chrome, we get a pause event of type "other", and vscode is getting  type "instrumentation". I got rid of the code that sets the beforeScriptWithSourceMapExecution instrumentation breakpoint, and then it works.

connor4312:
That's because with an instrumentation breakpoint, we wan to continue evaluation if the user hasn't also requested to break at this location in some way. Because Chrome on OSX is missing the async call stack behavior, the adapter doesn't forecast that the user expects a break at the entrypoint of the webworker, so it continues onwards.

connor4312:
Chatting with @dgozman:

>Connor Peet
>    One kind of ugly idea I had was to stack the last call frame we saw before stepping in, and then looking for that frame in the async stack of pause events to see if we got here from an async step.
>
>​Dmitry Gozman
>    I see the problem, let me think a bit
>
> Dmitry Gozman
>    Your solution seems ok for now, but I'd suggest to compare all call frames up to asyncStackTraceId to be on the safe side.
>
> Dmitry Gozman
[...]

PR #224, opened 2020-01-12
Fix mac tests, enable for CI

roblourens:
- Add a test reporter to get partial golden text from tests that did not assert their golden text log, such as timed out tests
  - And golden text will now write out a `*.actual` file whenever the assert fails, for easier comparison or changing the expectation file
- Fix tests that fail because they loaded an old version of node
  - I set `'--disable-user-env-probe'` for this, I think nvm is not installed properly so when vscode tries to discover the user's environment, the nvm path ends up behind `/usr/local/bin` in PATH. I have no clue why it's not an issue on linux
- And fixing a test that fails because logging `Window` produces a different preview on platform, it makes sense to not depend on that
- Disabling a test that fails from #223


src/test/reporters/goldenTextReporter.ts
```diff
@@ -0,0 +1,36 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+import * as mocha from 'mocha';
+import { IGoldenReporterTextTest } from './goldenTextReporterUtils';
+
+class GoldenTextReporter extends mocha.reporters.Spec {
+  static alwaysDumpGoldenText = false;
```
connor4312: May be convenient to have this configurable via an environment variable

PR #230, opened 2020-01-14
Send console output as plain text when possible

### LINKED ISSUE #227: Send plain text in output events if we can
It would speed things up and mitigate #161 (+ linked issues) to mirror behavior of the existing adapter and immediately send simple text logs in the output events. I definitely would like this before we start shipping to insiders as it makes running any fast-exiting unit tests/programs pretty rough.

roblourens:
so that the client doesn't have to send a variables
request for every output. We will just use a variablesReference
when an object is logged, or when not all logged data gets
into the text representation
Fix #227

### DIFF
--- a/src/adapter/threads.ts
+++ b/src/adapter/threads.ts
@@ -902,26 +902,38 @@ export class Thread implements IVariableStoreDelegate {
       event.args[0].value = localize('console.assert', 'Assertion failed');
 
     let messageText: string;
+    let usedAllArgs = false;
     if (event.type === 'table' && event.args.length && event.args[0].preview) {
       messageText = objectPreview.formatAsTable(event.args[0].preview);
     } else {
       const useMessageFormat = event.args.length > 1 && event.args[0].type === 'string';
       const formatString = useMessageFormat ? (event.args[0].value as string) : '';
-      messageText = messageFormat.formatMessage(
+      const formatResult = messageFormat.formatMessage(
         formatString,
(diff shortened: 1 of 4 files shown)


src/adapter/threads.ts
```diff
@@ -902,26 +902,38 @@ export class Thread implements IVariableStoreDelegate {
...
         useMessageFormat ? event.args.slice(1) : event.args,
         objectPreview.messageFormatters,
       );
+      messageText = formatResult.result;
+      usedAllArgs = formatResult.usedAllSubs;
+    }
+
+    let output: string;
+    let variablesReference: number | undefined;
+    if (!usedAllArgs || event.args.some(arg => objectPreview.isObject(arg))) {
```
connor4312: `isObject` excludes several [subtypes](https://github.com/microsoft/vscode-js-debug/blob/master/src/adapter/objectPreview.ts#L17). You probably want a different check here.

roblourens: So this is what VariableStore uses to decide whether to send a variable for an arg, or just a text preview, and that's why I used it, but I think that logic is wrong because currently we can't inspect the properties of those types at all, and you can in chrome devtools. So I will leave this and file an issue for handing those types correctly.

https://github.com/microsoft/vscode-js-debug/blob/9af83b87b156f97cd70d99cd4742fdb04f56f893/src/adapter/variables.ts#L252-L260

image

connor4312: Ok

PR #233, opened 2020-01-14
script skipping after applying feedback from other round of PRs

Shenniey:
Updates after applying feedback from [here](https://github.com/microsoft/vscode-js-debug/pull/94)


src/adapter/sources.ts
```diff
@@ -320,6 +341,18 @@ export class SourceContainer {
     return undefined;
   }
 
+  sourceByReference(refNum: number): Source | undefined {
```
roblourens: It seems like the new sourceByReference and sourceByUrl are doing basically the same thing as source(), you can pass in a fake Dap.Source or change it to take a new type, but I don't think we need these new getters and the new sourceByUrl map when we have sourceByAbsolutePath.

src/adapter/sources.ts
```diff
@@ -320,6 +341,18 @@ export class SourceContainer {
...
+    return this._sourceByReference.get(refNum);
+  }
+
+  sourceByUrl(url: string): Source | undefined {
+    return this._sourceByUrl.get(url);
+  }
+
+  isScriptSkipped(url: string): boolean {
```
roblourens: I don't think this is called anywhere

src/adapter/sources.ts
```diff
@@ -275,16 +301,19 @@ export class SourceContainer {
...
     dap: Dap.Api,
     private readonly sourceMapCache: SourceMapCache,
     public readonly rootPath: string | undefined,
     public readonly sourcePathResolver: ISourcePathResolver,
     public readonly localSourceMaps: ISourceMapRepository,
+    scriptSkipper: ScriptSkipper,
   ) {
     this._dap = dap;
+    this._scriptSkipper = scriptSkipper;
+    scriptSkipper.addNewSourceContainer(this);
```
roblourens: There is a circular dependency between ScriptSkipper and the SourceContainer which feels awkward but I'm not sure what to suggest

src/adapter/scriptSkipper.ts
```diff
@@ -27,24 +44,18 @@ export class ScriptSkipper {
...
-        }
-
-        const match = nodeInternalRegex.exec(nodeInternal);
-        return match ? match[1] : nodeInternal;
+        return nodeInternalRegex.exec(nodeInternal)![1];
       });
 
-    if (!this.skipAllNodeInternals && nodeInternalPatterns.length > 0) {
-      this._nodeInternalsRegex = this._createRegexString(nodeInternalPatterns);
+    this._nodeInternalsRegex = this._createRegexString(nodeInternalPatterns);
```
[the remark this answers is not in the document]
Shenniey: It's meant to be a regex that doesn't match any string.

src/adapter/sources.ts
Shenniey: Woops, removed it

scripts/dap-custom.json
```diff
@@ -163,6 +163,44 @@
 			}]
     },
 
+    "ToggleSkipFileStatusRequest": {
```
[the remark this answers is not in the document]
connor4312: You can add more globs to the [prettier args here](https://github.com/microsoft/vscode-js-debug/blob/master/gulpfile.js#L311) and then run `gulp format` to apply the standard formatting

src/adapter/sources.ts
```diff
@@ -68,6 +69,35 @@ const defaultTimeouts: SourceMapTimeouts = {
   output: 1000,
 };
 
+export class UrlToSourceMap {
```
connor4312: I'd recommend using our `MapUsingProjection` here, I think that does what you're after

src/adapter/sources.ts
connor4312: Yea, elsewhere we just create partial dap sources

src/adapter/sources.ts
```diff
@@ -598,9 +633,18 @@ export class SourceContainer {
           undefined,
           undefined,
         );
+        await this._scriptSkipper.updateSkippingValueForScript(source);
```
connor4312: I'm not sure that I like calling this synchronously in source unwrapping. Verifying the absolute path involves reading and hashing the whole file, which may slow things down significantly when debugging large projects; in 951c8f68c0affe3f6c53e4cf4a77fc77ce327455 we saw a huge performance gain by moving hashing to a new thread and this change makes it somewhat synchronous again.

src/adapter/sources.ts
```diff
@@ -571,7 +604,9 @@ export class SourceContainer {
       // whether |sourceUrl| looks like a path and belongs to the workspace.
       const sourceUrl = utils.maybeAbsolutePathToFileUrl(this.rootPath, url);
       const baseUrl = sourceMapUrl.startsWith('data:') ? compiled.url() : sourceMapUrl;
-      const resolvedUrl = utils.completeUrlEscapingRoot(baseUrl, sourceUrl);
+      const resolvedUrl = utils.properlyResolveFileUrl(
+        utils.completeUrlEscapingRoot(baseUrl, sourceUrl),
+      );
```
connor4312: I think I may have fixed what you were seeing in master already (on my Windows machine wfh today). See if this fixes it: https://github.com/microsoft/vscode-js-debug/blob/fd559e0e49661d7c558c3a22e9502eb3114a7679/src/adapter/sources.ts#L575-L577

src/adapter/scriptSkipper.ts
```diff
@@ -65,37 +76,156 @@ export class ScriptSkipper {
...
+    });
+
+    this._updateAllScriptsToBeSkipped();
+  }
+
+  public isScriptSkipped(url: string): boolean {
+    return this._isUrlSkippedMap.get(pathUtils.forceForwardSlashes(url)) === true;
+  }
+
+  public initializeNodeInternals(nodeInternalsNames: string[]): void {
```
roblourens: Doesn't have to be public, probably doens't need to be a method

src/adapter/scriptSkipper.ts
```diff
@@ -65,37 +76,156 @@ export class ScriptSkipper {
...
+  public async updateSkippingValueForScript(source: Source) {
+    const url = source.url();
+    if (!this._isUrlSkippedMap.has(pathUtils.forceForwardSlashes(url))) {
+      // applying sourcemappathoverrides results in incorrect absolute paths
+      // for some sources that don't map to disk (e.g. node_internals), so here
+      // we're checking for whether a file actually corresponds to a file on disk
+      const pathOnDisk = await source.existingAbsolutePath();
+      if (pathOnDisk) {
+        // file maps to file on disk
+        this._setUrlToSkip(url, this._testRegex(this._nonNodeInternalRegex, pathOnDisk));
```
roblourens: Probably faster to store these as actual RegExps instead of dynamically constructing a RegExp each time this is run

src/adapter/scriptSkipper.ts
```diff
@@ -65,37 +76,156 @@ export class ScriptSkipper {
...
+  }
+
+  public initializeNodeInternals(nodeInternalsNames: string[]): void {
+    this._allNodeInternals = nodeInternalsNames.map(name => name + '.js');
+  }
+
+  private _isNodeInternal(url: string): boolean {
+    return (
+      (this._allNodeInternals && this._allNodeInternals.includes(url)) ||
+      this._testRegex('^internal/.+\\.js$', url)
```
roblourens: Faster and more readable to do this with a regex literal inline like `/^internal/.+\.js$/.test(url)`

roblourens:
When dealing with skipping sourcemapped scripts, in v1, we also have to deal with the case where a.ts and b.ts are built into a single app.js, and a.ts is skipped but b.ts is not. We use a different API for that, setBlackboxedRanges, you can see it here: https://github.com/microsoft/vscode-chrome-debug-core/blob/486dff43ba6f336122e20157c03dd13bba25e522/src/chrome/scriptSkipping.ts#L131

We can proceed without that part but we should file an issue to track it.

src/adapter/scriptSkipper.ts
```diff
@@ -65,37 +76,156 @@ export class ScriptSkipper {
...
       } else {
-        this._isUrlSkippedMap.set(url, this._testRegex(this._nodeInternalsRegex, url));
+        if (this._isNodeInternal(url)) {
+          this._setUrlToSkip(url, this._testRegex(this._nodeInternalsRegex, url));
+        } else {
+          this._setUrlToSkip(url, this._testRegex(this._nonNodeInternalRegex, url));
+        }
+      }
+      if (this.isScriptSkipped(url)) {
+        this._updateAllScriptsToBeSkipped();
```
roblourens: It looks like every loaded script (and sourcemap for that script?) will trigger a new call to the runtime to set blackbox patterns. I'm worried about that causing a perf issue as the app is loading and 1000s of scripts are being loaded. Maybe this should be debounced?

src/adapter/sources.ts
Shenniey: Oh ok, I see. Thanks!

src/adapter/scriptSkipper.ts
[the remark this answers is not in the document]
roblourens: I don't think we should rely on a trick like that, maybe just keep it as null and never run it in that case.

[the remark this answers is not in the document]
roblourens: Similarly, you have the skippingNodeInternals flag which just says whether nodeInternalsRegex is valid. You can just make nodeInternalsRegex `| null` or something to get rid of the extra flag.

src/adapter/scriptSkipper.ts
```diff
@@ -27,24 +44,18 @@ export class ScriptSkipper {
...
-
-        const match = nodeInternalRegex.exec(nodeInternal);
-        return match ? match[1] : nodeInternal;
-      });
-
-    if (!this.skipAllNodeInternals && nodeInternalPatterns.length > 0) {
-      this._nodeInternalsRegex = this._createRegexString(nodeInternalPatterns);
+    const nodeInternalRegex = /^<node_internals>[\/\\](.*)$/;
+
+    const nodeInternalPatterns = this._userSkipPatterns!.filter(pattern =>
```
roblourens: `_userSkipPatterns` is just used to initialize some other properties and isn't referenced after the constructor is done, so it shouldn't be a property, just pass it to these initialize methods as an argument. It will be more clear how it's used.

src/adapter/scriptSkipper.ts
```diff
@@ -27,24 +44,18 @@ export class ScriptSkipper {
   }
 
   private _preprocessNodeInternals(): void {
-    const nodeInternalRegex = /^<node_internals>[\/|\\\\](.*)$/;
-    const skipAllNodeInternalsRegex = /^<node_internals>[\/|\\\\]\\*\\*[\/|\\\\]\\*.js/;
+    const nodeInternalRegex = /^<node_internals>[\/\\](.*)$/;
 
-    const nodeInternalPatterns = this._userSkipPatterns
+    const nodeInternalPatterns = this._userSkipPatterns!
       .filter(pattern => pattern.includes('<node_internals>'))
```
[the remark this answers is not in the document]
roblourens: I don't know if this was supposed to be marked as resolved?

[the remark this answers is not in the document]
Shenniey: No, not yet! Sorry, I've got that in a commit I haven't pushed yet. I'll resolve it when that gets pushed.

src/dap/api.d.ts
```diff
@@ -22,10 +21,7 @@ export namespace Dap {
      * This can either be a normal result ('success' attribute true) or an error response ('success' attribute false and the 'message' set to 'cancelled').
      * Returning partial results from a cancelled request is possible but please note that a frontend client has no generic way for detecting that a response is partial or not.
      */
-    on(
-      request: 'cancel',
-      handler: (params: CancelParams) => Promise<CancelResult | Error>,
-    ): () => void;
+    on(request: 'cancel', handler: (params: CancelParams) => Promise<CancelResult | Error>): () => void;
```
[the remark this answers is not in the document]
Shenniey: Yeah, I think it's because I re-ran the "generate DAP API" task after I updated the custom-dap.json file (adding the toggle requests). I'll re-run gulp format and see if that fixes it

src/adapter/sources.ts
Shenniey: It's not exactly what I need 😔. That leaves the resolvedUrl looking like this:

image


But in order to do the url comparisons I need it to look like this:
image

src/dap/api.d.ts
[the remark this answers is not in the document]
Shenniey: I think it's formatted correctly now

src/ui/toggleSkippingFile.ts
```diff
@@ -0,0 +1,32 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import Dap from '../dap/api';
+import { fileURLToPath } from 'url';
+import { isFileUrl } from '../common/urlUtils';
+
+export function toggleSkippingFile(aPath: string): void {
```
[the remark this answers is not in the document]
Shenniey: Done!

src/adapter/sources.ts
Shenniey: I think this might be taken care of with the debouncer? If you could take another look

src/adapter/scriptSkipper.ts
Shenniey: Done!

src/adapter/sources.ts
Shenniey: Fixed!

roblourens:
Good progress in those 2 commits, don't forget to take a look at your merge conflicts too and let me know if you need help resolving them.

package.json
```diff
@@ -248,7 +248,7 @@
         {
           "command": "extension.NAMESPACE(node-debug).toggleSkippingFile",
           "group": "navigation",
-          "when": "inDebugMode && debugType == 'node' && callStackItemType == 'stackFrame'"
+          "when": "inDebugMode && debugType == 'NAMESPACE(node-debug)' && callStackItemType == 'stackFrame'"
```
roblourens: This should be contributed for Chrome too (you can just remove the debugType)

Shenniey: Done! Thanks

src/common/urlUtils.ts
```diff
@@ -157,6 +157,16 @@ export function isValidUrl(url: string): boolean {
   }
 }
 
+export function properlyResolveFileUrl(url: string): string {
```
roblourens: This isn't called anywhere, you can delete it

roblourens:
If you are looking at tests, there is a relevant commented out test in stacksTest.ts

src/adapter/debugAdapter.ts
```diff
@@ -322,6 +326,17 @@ export class DebugAdapter {
     return {};
   }
 
+  async _toggleSkipFileStatus(
+    params: Dap.ToggleSkipFileStatusParams,
+  ): Promise<Dap.ToggleSkipFileStatusResult | Dap.Error> {
+    if (!this._thread?.pausedDetails()) {
+      return errors.createSilentError(localize('error.threadNotPaused', 'Thread is not paused'));
```
connor4312: Does it need to be paused? Could I toggle skipping for a file even before I hit a BP?

src/common/urlUtils.ts
Shenniey: Done!

src/adapter/debugAdapter.ts
Shenniey: Actually, that's a good point. I put that in because of [this](https://github.com/microsoft/vscode-js-debug/pull/94#discussion_r350502705) comment, but I'm not sure why it needs to be paused. @roblourens ?

roblourens: Yeah I guess that's not required, the command is still in the command palette

PR #238, opened 2020-01-15
fix: breakpoint applying incorrectly in babel TS code

### LINKED ISSUE #231: BP goes to wrong line
- Debug the test repo from https://github.com/microsoft/vscode/issues/88532
- Set a breakpoint on line 3, `console.log('hello!')`
- It goes to line 5

connor4312:
Fixes #231

I made this change several months ago and noted and issue that required
us to use LEAST_UPPER_BOUND, however I'm no longer able to reproduce it
and this fixes babel positioning. I put together a couple test cases
to ensure the positions are correct; if we run into more problems, we
can add cases for them and fix them.


src/adapter/sources.ts
```diff
@@ -433,7 +433,7 @@ export class SourceContainer {
         source: sourceUrl,
         line: uiLocation.lineNumber,
         column: uiLocation.columnNumber - 1, // source map columns are 0-indexed
-        bias: SourceMapConsumer.LEAST_UPPER_BOUND,
+        bias: SourceMapConsumer.GREATEST_LOWER_BOUND,
```
roblourens: v1 first does GREATEST_LOWER, and if it doesn't return a result (there can be no mapping in that direction) it searches with LEAST_UPPER.

Also this can cause the breakpoint to shift up a line, I have wanted to check for that case and prefer a position that eventually maps to the same source line.

roblourens: I have been meaning to go through test repos in old v1 issues and maybe that will help find issues like that.

connor4312:
Hah, that's fun. Tests fail due to an off-by-one error that's been there for months. Fixing that also fixes this :)

image

to:

image

src/adapter/sources.ts
connor4312: Hm, okay. So basically the algorithm would be something like:

```py
positions = map.get_all_position_for(location).sort()
for position in positions:
  # prefer returning the first position on the source line
  if map.get_original_position(position).line == location.line:
    return location

return positions[0] # default to greatest_lower_bound
```

connor4312: Hm, okay. So basically the algorithm would be something like:

```py
positions = map.get_all_position_for(location).sort()
for position in positions:
  # prefer returning the first position on the source line
  if map.get_original_position(position).line == location.line:
    return location

return positions[0] # default to greatest_lower_bound
```

src/adapter/sources.ts
```diff
@@ -460,6 +493,48 @@ export class SourceContainer {
...
+   * bias passed to the method controls which of the matches we choose.
+   * Here, we will try to pick the position that maps back as closely as
+   * possible to the source line if we get an approximate match,
+   */
+  private getOptimalCompiledPosition(
+    sourceUrl: string,
+    uiLocation: IUiLocation,
+    sourceMap: SourceMapConsumer,
+  ): NullablePosition | undefined {
+    const nextLocation = sourceMap.generatedPositionFor({
```
roblourens: I think the "prev" and "next" names should be reversed

PR #242, opened 2020-01-16
fix: edge cases with vue projects

connor4312:
Vue is different than what we've seen so far in that, for each
component, it generates multiple 'files' to load CSS, HTML, and so on.
There might be the main file `webpack:///src/App.vue`, a CSS import
in `webpack:///src/App?BADF00D`, etc. The previous behavior of the
source collection was to always overwrite sources by their absolute
path when they got parsed, and as a result breakpoints would often
resolve into the wrong, stubbed file.

In this PR we add a heuristic to prefer the shortest URL if we see
multiple source mapped file mapping to the same file on disk. We can't
just discard any URLs with query strings, since often those are totally
valid, but this seems to make the Vue starter work well.


src/adapter/threads.ts
```diff
@@ -643,22 +643,23 @@ export class Thread implements IVariableStoreDelegate {
   rawLocation(
     location: Cdp.Runtime.CallFrame | Cdp.Debugger.CallFrame | Cdp.Debugger.Location,
   ): RawLocation {
-    // Note: cdp locations are 0-based, while ui locations are 1-based.
-    if ((location as Cdp.Debugger.CallFrame).location) {
+    // Note: cdp locations are 0-based, while ui locations are 1-based. Also,
+    // some we can *apparently* get negative locations; Vue's "hello world"
+    // project was observed to emit source locations at (-1, -1) in its callframe.
```
connor4312: ‾\\\_(ツ)\_/‾

roblourens:
I thought in Vue with hot module reloading, the "real" file was `webpack:///src/App.vue?something` and `webpack:///src/App.vue` was like a HMR stub or something, but maybe I have it backwards.

connor4312:
I think so :)

roblourens:
This is such ugly nonsense but the "real" .vue files are shorter because they've been pulled out of the folder structure, not because they are missing `?asdf` so we are both right

image

connor4312:
Oh, huh, that's different from what I was seeing with a fresh Vue starter: there the real files are missing the suffix (and also in a different folder). But, um, should work I guess 🙃 If this heuristic doesn't solve Vue then we would probably need to open an issue on them, maybe to put their fake files into an `__interna__` directory... this is a really hard scenario for us to solve correctly as a debug adapter.

roblourens:
This is a new project from the vue cli, and the files without the suffix are HMR stubs with a `webpack-internal://` URI, they are out of the screenshot

connor4312:
I did the same with this in my vue.config.js:

```
module.exports = {
  configureWebpack: {
    devtool: 'source-map'
  }
}
```

The default uses cheap-eval-source-map which neither the new or old adapters yet support very well as far as I know

PR #250, opened 2020-01-21
feat: allow restarting process attachment

### LINKED ISSUE #245: "restart" on attach config not working
- Debug vscode
- Run "Attach to Extension Host"
- Reload the window
- The EH config should be disconnected, but then reattach when the EH is reloaded, but it never reattaches

connor4312:
This borrows some code from the Node process launcher to implement
restarting process attachment. We'll print reconnecting messages to the
console while we're working on reconnection, and query the original
address/port to get a new inspector URL.

Fixes #245

### DIFF
--- a/src/build/generate-contributions.ts
+++ b/src/build/generate-contributions.ts
@@ -28,12 +28,14 @@ type OmittedKeysFromAttributes =
   | '__workspaceFolder'
   | '__workspaceCachePath';
 
+type DescribedAttribute<T> = JSONSchema6 &
+  Described & {
+    default: T;
+    enum?: Array<T>;
+  };
+
 type ConfigurationAttributes<T> = {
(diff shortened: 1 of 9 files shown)


src/build/generate-contributions.ts
```diff
@@ -290,6 +292,43 @@ const nodeAttachConfig: IDebugger<INodeAttachConfiguration> = {
...
+    restart: {
+      description: refString('node.attach.restart.description'),
+      default: true,
+      oneOf: [
+        {
+          type: 'boolean',
+        },
+        {
+          type: 'object',
+          required: ['exponential'],
```
roblourens: All these options kind of seem like overkill, but sure :)

connor4312: Eh probably 😛  We'll get telemetry on who set them so we can always yank them if we see no one cases about it.

PR #252, opened 2020-01-21
feat: make logpoint parsing more resilient

### LINKED ISSUE #226: logPoints don't work correctly with } in the expression part
After https://github.com/microsoft/vscode-js-debug/issues/113, logpoints will have a defect that also existed in v1. Our parsing of expressions out of logpoints is naive and won't work correctly when the expression contains a }.

One idea would be to look for the first {, then parse all text after that with TS. The first statement should be a block type. The end of the block would be the end of the embedded expression?

connor4312:
Fixes #226. As suggested in the original issue, we use the TS parser
to parse the code inside the {} 'blocks', which lets us handle more
complex expressions. We also now support multiple statements, and
(try to) return the last one, and log the stacks of any thrown errors
rather than silently failing. Finally, we return a verbose error to the
user in the event that their logpoint isn't valid JS, though there's
an issue with VS Code right now: https://github.com/microsoft/vscode/issues/89059

### DIFF
--- a/src/adapter/breakpoints/logPoint.ts
+++ b/src/adapter/breakpoints/logPoint.ts
@@ -1,0 +1,118 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+import * as ts from 'typescript';
+import { assert, logger } from '../../common/logging/logger';
+import { invalidLogPointSyntax } from '../../dap/errors';
+import { LogTag } from '../../common/logging';
+/**
+ * Statments which we should not prefix with `return {}`
+ */
+const unreturnable: ReadonlySet<ts.SyntaxKind> = new Set([
+  ts.SyntaxKind.ReturnStatement,
+  ts.SyntaxKind.IfStatement,
+  ts.SyntaxKind.DoStatement,
+  ts.SyntaxKind.WhileStatement,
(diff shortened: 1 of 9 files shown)


src/adapter/breakpoints/logPoint.ts
```diff
@@ -0,0 +1,118 @@
...
+  ts.SyntaxKind.ReturnStatement,
+  ts.SyntaxKind.WithStatement,
+  ts.SyntaxKind.SwitchStatement,
+  ts.SyntaxKind.LabeledStatement,
+  ts.SyntaxKind.TryStatement,
+  ts.SyntaxKind.ThrowStatement,
+  ts.SyntaxKind.DebuggerStatement,
+  ts.SyntaxKind.VariableDeclaration,
+  ts.SyntaxKind.VariableDeclarationList,
+]);
```
connor4312: The problem here is that if the breakpoint just has side-effect code like `while (foo) { bar(); }`, emitting `return while (foo)` is invalid. Instead of having a denylist of grammars, we could just `try { new Function('return ${stmt}') }` to see if it's valid. Not quite as fast or 'pure', but probably more correct. Could go either way, thoughts?

roblourens: I would probably go with the dynamic approach. At least it would be future-proof. But I could go either way.

PR #257, opened 2020-01-24
Handle skipFiles for multiple source files built into a bundle

### LINKED ISSUE #253: Can't individually skip source files that are compiled into a js bundle
- Debug an app with `a.ts, b.ts -> app.js`
- Set `skipFiles: ['a.ts']`
- Both a.ts and b.ts will be skipped

roblourens:
Fixes #253

The basic idea is that I keep track of authored source skip status separately from generated scripts', and instead of sending their paths to the runtime, send the ranges in the file that is skipped.

Since we have to send this separately per scriptID to each Cdp, it got a lot easier to change ScriptSkipper from one shared instance per root session, to one instance per Target. To get the instances to share toggle statuses under the same root session, they now fire events at each other with their root target id whenever a toggle happens.

Also some other minor tweaks and fixes.

### DIFF
--- a/src/adapter/scriptSkipper.ts
+++ b/src/adapter/scriptSkipper.ts
@@ -133,13 +133,9 @@ export class ScriptSkipper {
     );
   }
 
-  private _hasScript(url: string): boolean {
-    return this._isUrlSkipped.has(this._normalizeUrl(url));
-  }
-
-  private async _initializeSourceMappedSources(
+  private async _updateSourceWithSkippedSourceMappedSources(
     source: Source,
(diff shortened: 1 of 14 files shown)


src/adapter/scriptSkipper.ts
```diff
@@ -71,140 +91,154 @@ export class ScriptSkipper {
   }
 
   private _createRegexString(patterns: string[]): string {
+    // TODO this should use node-glob directly
```
roblourens: I'll do this

connor4312: We use minimatch already for some other stuff, you can grab that

PR #261, opened 2020-01-25
feat: handle source roots, pathmapping correctly

connor4312:
Previously, we didn't do much with path mappings, and were only
vaguely correct around dealing with source roots. (We were defaulting
to what the source-map library did for us, which it turns out is
sometimes destructive, resolving `../foo.ts` with a sourceRoot of `/`
to `/foo.ts`, as if it was its own filesystem).

This PR imports the bulk of its handling from chrome-debug-core, with
some tweaks to fit. One thing it adds is handling absolute source roots
inside of node modules; it defaults to mapping that relative to the root
of the module. We could expose pathMappings in the Node config as well
to let users configure this.


src/common/sourceMaps/sourceMapResolutionUtils.ts
```diff
@@ -0,0 +1,182 @@
...
+      // Technically not valid but it's easy to end up with paths like this.
+      absSourceRoot = (await resolver(sourceRoot, pathMapping)) || sourceRoot;
+
+      // If no pathMapping (node), use sourceRoot as is.
+      // But we also should handle an absolute sourceRoot for chrome? Does CDT handle that? No it does not, it interprets it as "localhost/full path here"
+    } else if (path.isAbsolute(generatedPath)) {
+      // sourceRoot is like "src" or "../src", relative to the script
+      absSourceRoot = resolveRelativeToFile(generatedPath, sourceRoot);
+    } else {
+      // generatedPath is a URL so runtime script is not on disk, resolve the sourceRoot location on disk.
```
roblourens: Can this branch happen now? It looks like generatedPath might always be an absolute path?

connor4312: We process the `sourceRoot` from the sourcemap like how we did before. The sourcemap spec is rather loose definition and I haven't personally seen sourceRoots that are URLs, but it's a code path in chrome-debug-core.

Issue #235, opened 2020-01-14
Support the Edge browser


roblourens:
Need to bring code for launching Edge from https://github.com/microsoft/vscode-edge-debug2/ to this repo.

dhanvikapila:
@roblourens - We will be bringing the code to launch the new Edge browser and not Legacy Edge.

connor4312:
Edge support work is complete at this time. Further improvements may be tracked in their own issues.

Testers: this is a meta-issue and does not need verification.

Issue #44, opened 2019-10-23
Integration test status


roblourens:
The current status of failing Chrome integration tests ported from the old codebase

Had to run some of these individually to get a clean run, since Chrome doesn't always shut down.

- [ ] "Attach without specifying a url"
  - Attach request
- [ ] 2 adapter.test.ts case sensitivity tests
  - Need to patch `dc.launchRequest` same as `dc.launch`
- [ ] "should stop on debugger statement"
  - "debugger" pause reason
- [ ] "Unknown request should produce error"
  - Timeout, looks for certain error response
- [ ] "Hit breakpoint on JavaScript when source map is invalid"
  - Crashes adapter, exception thrown from `source-map`
- [ ] 2 "Should launch successfully on port 0" tests
  - Not hooked up properly (?)
- [ ] "Unusual source-maps"
  - Column off by 1
- [ ] "Should stop on a breakpoint in an in-line script"
  - Stack frame format (?)
- [ ] "No unhandled exceptions when we parse invalid JavaScript code"
  - Looks for certain error output
- [ ] 3 "loaded sources" tests
  - "debugger" pause reason
  - And late "continue" events (DA sends "continue" after disconnect which Pinezorro will not like)
- [ ] 4 "stack traces" tests
  - Some difference in formatting
- [ ] Some others previously disabled as failing
- [ ] Many hit condition breakpoints
  - Not implemented
- [x] Break on load related tests
  - Needs investigation
- [ ] React framework tests
  - Needs investigation

Also only Chrome tests have been ported. There are a smaller number of node tests that I need to port as well.

PR #264, opened 2020-01-27
feat: WebView2 debugging support

johnemau:
* Minimum changed needed to support `msedge` debugger type, it is just an alias for `chrome`.
* This change is equivalent to the launch.json `{type: 'chrome', runTimeExecutable: 'path/to/msedge.exe'}`.
* Minimum change needed to support backwards compatibility with WebView2 launch
  debugging configurations (`useWebView`) from [vscode-edge-debug2][1].

Follow-ups to consider:
* Separate `chrome` and `msedge` debugger type configurations with Edge strings.
* Validate `useWebView` configuration is configured with `runTimeExecutable`.
* Tests for `useWebView` configuration.
* Add browser `version` support (`stable`, `beta`, `dev`, `canary`) for Microsoft Edge (Chromium)
  to match backwards compatibility with [vscode-edge-debug2][1].
* Deprecate `{useWebView: 'advanced'}` in favor of `{useWebView: true, urlFilter: 'bing.com'}`
* Deprecate `useWebView` in favor of `webview` debugger type.

[1]: https://github.com/microsoft/vscode-edge-debug
 
Related: #235

### DIFF
--- a/package.json
+++ b/package.json
@@ -164,7 +164,7 @@
     "onCommand:extension.NAMESPACE(chrome-debug).addCustomBreakpoints",
     "onCommand:extension.NAMESPACE(chrome-debug).removeAllCustomBreakpoints",
     "onCommand:extension.NAMESPACE(chrome-debug).removeCustomBreakpoint",
-    "onDebugResolve:NAMESPACE(edge)",
+    "onDebugResolve:NAMESPACE(msedge)",
     "onDebugResolve:NAMESPACE(extensionHost)"
   ],
   "contributes": {
--- a/package.json
+++ b/package.json
@@ -164,6 +164,7 @@
     "onCommand:extension.NAMESPACE(chrome-debug).addCustomBreakpoints",
     "onCommand:extension.NAMESPACE(chrome-debug).removeAllCustomBreakpoints",
     "onCommand:extension.NAMESPACE(chrome-debug).removeCustomBreakpoint",
+    "onDebugResolve:NAMESPACE(edge)",
     "onDebugResolve:NAMESPACE(extensionHost)"
   ],
   "contributes": {
--- a/src/build/strings.ts
+++ b/src/build/strings.ts
@@ -27,6 +27,9 @@ const strings = {
   'extensionHost.snippet.launch.description': 'Launch a VS Code extension in debug mode',
   'extensionHost.snippet.launch.label': 'VS Code Extension Development',
 
+  'edge.useWebView.description':
+    "(Edge only) When 'true', the debugger will treat the runtime executable as a host application that contains a WebView allowing you to debug the WebView script content. When set to 'advanced', the debugger will wait to connect to a WebView matching the urlFilter.",
+
   'chrome.address.description': 'TCP/IP address of debug port',
   'chrome.baseUrl.description':
     'Base URL to resolve paths baseUrl. baseURL is trimmed when mapping URLs to the files on disk. Defaults to the launch URL domain.',
--- a/src/build/strings.ts
+++ b/src/build/strings.ts
@@ -28,7 +28,7 @@ const strings = {
   'extensionHost.snippet.launch.label': 'VS Code Extension Development',
 
   'edge.useWebView.description':
-    "(Edge only) When 'true', the debugger will treat the runtime executable as a host application that contains a WebView allowing you to debug the WebView script content. When set to 'advanced', the debugger will wait to connect to a WebView matching the urlFilter.",
+    "(Edge (Chromium) only) When 'true', the debugger will treat the runtime executable as a host application that contains a WebView allowing you to debug the WebView script content. When set to 'advanced', the debugger will wait to connect to a WebView matching the urlFilter.",
 
   'chrome.address.description': 'TCP/IP address of debug port',
   'chrome.baseUrl.description':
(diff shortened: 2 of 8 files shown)


connor4312:
> Separate chrome and edge debugger type configurations with Edge strings.

👍 -- we already have an Edge-only configuration option, there'll surely be more eventually. We'll probably want to split BrowserLauncher into a BrowserLauncherBase and then a ChromeLauncher and EdgeLauncher, like we do for the Node launcher already. Firefox is becoming CDP-compatible as well so there may eventually be one of those as well. Would be happy to have that in this PR, but it could be a followup item for anyone to implement.

> Validate useWebView configuration is configured with runTimeExecutable.

Would be nice to allow simply entering `edge` as a launch runtime executable. I assume Edge has well-known location(s). `findBrowser.ts` has a bunch of platform logic for this; you could add a new Executable in win32() that discovers an `edge` type, and then it'll work.

> Add browser version support (stable, beta, dev, canary) for Microsoft Edge (Chromium)
to match backwards compatibility with vscode-edge-debug2.

Currently for Chrome we overload runtimeExecutable and allow users to enter `stable`, `canary`, etc. With multiple browsers this gets awkward. Maybe we introduce a separate `browserVersion` property for this.

johnemau:
I ran the Chrome integration tests locally (`npm run intTests`) and here are my results:

```cmd
  12 passing (5m)
  46 pending
  33 failing
```

The failures do not match what was reported in issues #44, however I am worried because the report is likely out of date and I assume to need undocumented local changes to prune out false-negatives.

johnemau:
> Would be happy to have that in this PR, but it could be a followup item for anyone to implement.

Lets do this in a follow-up PR for anyone to implement.

> I assume Edge has well-known location(s). `findBrowser.ts` has a bunch of platform logic for this; you could add a new Executable in win32() that discovers an `edge` type, and then it'll work.

I agree and propose we update `findBrowser.ts` in a follow-up PR.

> Currently for Chrome we overload runtimeExecutable and allow users to enter `stable`, `canary`, etc. With multiple browsers this gets awkward. Maybe we introduce a separate `browserVersion` property for this.

Sounds good to me, another good follow-up PR.

---

@connor4312, what is your preferred way to track the follow-up work?
Suggestions: I can file separate github issues or we can start a checklist on issue #235

connor4312:
I'd prefer separate Github issues that reference the main one; easier to track ongoing milestones, state, and assignments at a more granular level that way.

connor4312:
Let me know when you're happy merging this PR, and I'll get it in then. Note that this is endgame week for VS Code so we won't be as responsive until we normally are until Monday.

johnemau:
Thanks for the review @connor4312! This PR ready to land so please merge!

Per your suggestion I will file separate github issues for the follow-up work and reference the main Edge issue #235.

src/build/strings.ts
```diff
@@ -27,6 +27,9 @@ const strings = {
   'extensionHost.snippet.launch.description': 'Launch a VS Code extension in debug mode',
   'extensionHost.snippet.launch.label': 'VS Code Extension Development',
 
+  'edge.useWebView.description':
+    "(Edge only) When 'true', the debugger will treat the runtime executable as a host application that contains a WebView allowing you to debug the WebView script content. When set to 'advanced', the debugger will wait to connect to a WebView matching the urlFilter.",
```
dhanvikapila: Should this say "Edge (Chromium)" only?

package.json
```diff
@@ -164,6 +164,7 @@
     "onCommand:extension.NAMESPACE(chrome-debug).addCustomBreakpoints",
     "onCommand:extension.NAMESPACE(chrome-debug).removeAllCustomBreakpoints",
     "onCommand:extension.NAMESPACE(chrome-debug).removeCustomBreakpoint",
+    "onDebugResolve:NAMESPACE(edge)",
```
dhanvikapila: Can we rename "Edge" to "Msedge" everywhere? In the old debugger, Edge has meant Legacy Edge while msedge is for the new Edge Version.  Would be good to continue that terminology in this as well.

src/targets/browser/browserLauncher.ts
```diff
@@ -144,6 +144,18 @@ export class BrowserLauncher implements ILauncher {
     );
   }
 
+  prepareWebViewLaunch(params: IChromeLaunchConfiguration) {
+    // port 2015 is carried over from vscode-edge-debug2
+    params.port = params.port || 2015;
```
dhanvikapila: This should not be required for new Edge. 2015 was used because with Legacy Edge there is only one debug server for all Edge processes.  That is not the case with Chrome/MsEdge.

src/build/strings.ts
johnemau: This prefix should be removed as part of #266

johnemau: Done.

src/targets/browser/browserLauncher.ts
johnemau: Done.

package.json
johnemau: Done.

roblourens:
Is there any way to use "edge" for the new Edge and some other id for the legacy version? I think people would go for `edge` over `msedge` by default. And how many people want to use legacy edge when the new one is available?

dhanvikapila:
From a VS point of view there are going to be significant numbers of users for both versions of Edge.  For external customer facing strong sure we can say edge in v3 but we need to say msedge in telemetry to be able to differentiate between old edge and new one.

- Dhanvi

________________________________

roblourens:
I think we should use "edge" for user-facing things in VS Code and we can map it to msedge for telemetry, that's fine.

PR #276, opened 2020-01-29
adding script skipping tests

Shenniey:
Adding some tests for script skipping.

For a couple tests, I've added a short sleep in the tests to try to force the debouncer to execute before the stacktrace call (if we don't do this, the scripts aren't shown as hidden). This works on my machine locally but I feel like this has the potential to be flaky on CI. I'm not sure if there's a better way?


src/test/node/node-runtime-skipfiles-skip-node-internals.txt
```diff
@@ -0,0 +1,13 @@
...
+Module._compile @ internal/modules/cjs/loader.js:686:28 <hidden: Skipped by skipFiles>
+Module._extensions..js @ internal/modules/cjs/loader.js:734:10 <hidden: Skipped by skipFiles>
+Module.load @ internal/modules/cjs/loader.js:620:32 <hidden: Skipped by skipFiles>
+tryModuleLoad @ internal/modules/cjs/loader.js:560:12 <hidden: Skipped by skipFiles>
+Module._load @ internal/modules/cjs/loader.js:552:3 <hidden: Skipped by skipFiles>
+Module.runMain @ internal/modules/cjs/loader.js:776:12 <hidden: Skipped by skipFiles>
+executeUserCode @ internal/bootstrap/node.js:342:17 <hidden: Skipped by skipFiles>
+startExecution @ internal/bootstrap/node.js:276:5 <hidden: Skipped by skipFiles>
+startup @ internal/bootstrap/node.js:227:5 <hidden: Skipped by skipFiles>
+bootstrapNodeJSCore @ internal/bootstrap/node.js:743:3 <hidden: Skipped by skipFiles>
```
connor4312: I might actually make this assertion a regex, or otherwise test this output manually. The bottom portion of this stacktrace will vary between any different Node.js versions and make this test flake when we upgrade or if someone locally has a slightly different JS version.

src/test/node/node-runtime.test.ts
```diff
@@ -47,6 +48,22 @@ describe('node runtime', () => {
...
+    const cwd = join(testWorkspace, 'simpleNode');
+    const handle = await r.runScript(join(cwd, 'index.js'), { skipFiles: ['<node_internals>/**'] });
+    await handle.dap.setBreakpoints({
+      source: { path: join(cwd, 'index.js') },
+      breakpoints: [{ line: 1, column: 1 }],
+    });
+
+    handle.load();
+    const stoppedParams = await handle.dap.once('stopped');
+    await sleep(200);
```
connor4312: This delay shouldn't be necessary, were you hitting issues here?

connor4312: ah, I see the comment on the similar sleep below. Maybe comment this line too. Alternatively, maybe a better approach, should the client requesting a stacktrace caused any debounced blackboxing to apply immedately? If the client requests a stacktrace quickly after launch the user would see the lack of blackboxing within their ui.

roblourens: It's too late at that point - the blackbox patterns have not actually been sent to the runtime, and the stacktrace hints should match what the runtime knows.

src/test/node/node-runtime.test.ts
```diff
@@ -18,6 +18,7 @@ import { spawn, ChildProcess } from 'child_process';
 import Dap from '../../dap/api';
 import { delay } from '../../common/promiseUtil';
 import { nodeLaunchConfigDefaults, INodeLaunchConfiguration } from '../../configuration';
+import { sleep } from '../../int-chrome/testUtils';
```
roblourens: Don't use stuff from int-chrome, it might go away. This probably exists somewhere else

connor4312: There's `delay` in promise utils

connor4312: There's `delay` in promise utils

src/test/node/node-runtime.test.ts
connor4312: mm, that's unfortunate

src/test/stacks/stacks-skipfiles-single-authored-js.txt
```diff
@@ -0,0 +1,3 @@
+bar @ ${workspaceFolder}/web/script.js:6:3
```
roblourens: Why is the first frame not skipped?

src/test/stacks/stacks-skipfiles-multiple-authored-ts-to-js.txt
```diff
@@ -0,0 +1,6 @@
+foo @ ${workspaceFolder}/web/browserify/module1.ts:3:3
+bar @ ${workspaceFolder}/web/browserify/module2.ts:3:3
```
roblourens: Shouldn't these be marked as skipped?

Shenniey: Fixed!

src/test/stacks/stacks-skipfiles-single-authored-js.txt
Shenniey: This is also b/c of having to wait for the debouncer. Fixed

src/test/node/node-runtime.test.ts
Shenniey: Done

src/test/node/node-runtime-skipfiles-skip-node-internals.txt
Shenniey: Done!

src/test/node/node-runtime.test.ts
Shenniey: added comments

src/test/node/node-runtime-skipfiles-skip-node-internals.txt
```diff
@@ -0,0 +1,13 @@
+<anonymous> @ ${workspaceFolder}/simpleNode/index.js:1:1
```
roblourens: Should this file be deleted now that the custom assert is used?

Shenniey: Yep! Removed

PR #283, opened 2020-01-30
fix: filter should match against `file` property

johnemau:
### Problem
Users configuring `file` will not get their target filtered after launch.

### Solution
Add `file` property to filter check created for configs.

### Background
Users may provide a file instead of url, in that case the file path will be converted into a file url (`file://`) and loaded after launch.

The `'file' in config` was needed for TypeScript because `attach` configurations do not have `file` property by default.

### Bonus
Removed restrictive CDP type from filter, this will be helpful for an upcoming web view change.

### DIFF
--- a/src/common/urlUtils.ts
+++ b/src/common/urlUtils.ts
@@ -8,7 +8,6 @@ import * as path from 'path';
 import * as http from 'http';
 import * as https from 'https';
 import { fixDriveLetterAndSlashes, isUncPath } from './pathUtils';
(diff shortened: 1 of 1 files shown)


johnemau:
@connor4312 / @roblourens this change moves us closer to vscode-edge-debug2 parity and is needed for an upcoming change I have to fully match web view debugging parity.

Once this is merged I will open the PR for the web view change.

Thank you for your time looking at this!

Issue #266, opened 2020-01-27
Support Edge: Add unique launch/attach configurations (with strings) for `edge` debugger type


johnemau:
#### Background
This is part of the effort to achieve backwards compatibly with [vscode-edge-debug2] and [fully support debugging Microsoft Edge (Chromium)][1].

PR #264 added the `msedge` debugger type as a simple alias for `chrome` reusing all Chrome browser launch configurations.

#### What needs to happen
* Split the Edge and Chrome configurations (`configuration.ts`) into separate types, with shared properties in a common "Browser" configuration type (feel free to pick a better name). Ideally this type can be reused to onboard other Chromium/CDP compatible browsers.
* Add strings specific to the Edge browser to describe the Edge configuration values, the strings can be found in the [vscode-edge-debug2] project.
* Add an `EdgeDebugConfigurationProvider`
* Remove "Edge only" warnings in strings/comments, [1][2]

Related: #235

[vscode-edge-debug2]: https://github.com/microsoft/vscode-edge-debug2
[1]: https://github.com/microsoft/vscode-js-debug/issues/235
[2]: https://github.com/microsoft/vscode-js-debug/pull/264/files/de0ebbb285ca396816c8b5e599e3368a1b55e27c#r371569119

PR #285, opened 2020-01-30
feat: support multi-browser process web view debugging

johnemau:
In a [previous change][1] I introduced basic debugging support for WebView2 targets, that change lacked support for web views that were created under multiple browser processes because the web views would start under ports not known by our debugger.

In this change we add support for multiple browser processes by leveraging the web view environment variable: `WEBVIEW2_PIPE_FOR_SCRIPT_DEBUGGER ` which allows us to listen on a known pipe hosted by the OS for any web view process communication, the specific details can be [read here][2].

#### Here is the approach we take

1. Start listening to the pipe before the web view process is launched.
2. Wait to attach until after the matching target sends its port over the pipe.
3. Keep listening on the pipe for any web views to resume because all web views created under our debugger start as paused.

#### Notes

* This becomes the default code path for all web view debugging so the web view filter check has moved to the web view pipe server.

* Fixes the issue where the debugger may never attach to a web view if the port was not configured by the user.

#### Demo gif
vscode-js-debug-full-wv

Thanks to @jalissia for helping me understand web view's multi-browser process model.

Related: #264

[1]: https://github.com/microsoft/vscode-js-debug/pull/264
[2]: https://docs.microsoft.com/microsoft-edge/hosting/webview2/reference/webview2.idl

### DIFF
--- a/src/targets/browser/browserLauncher.ts
+++ b/src/targets/browser/browserLauncher.ts
@@ -2,12 +2,15 @@
  * Copyright (C) Microsoft Corporation. All rights reserved.
  *--------------------------------------------------------*/
 
+import { createServer } from 'net';
+import { randomBytes } from 'crypto';
+import { tmpdir } from 'os';
 import * as fs from 'fs';
 import * as path from 'path';
 import { CancellationToken } from 'vscode';
 import * as nls from 'vscode-nls';
 import CdpConnection from '../../cdp/connection';
-import { timeoutPromise } from '../../common/cancellation';
+import { timeoutPromise, NeverCancelled } from '../../common/cancellation';
 import { Contributions } from '../../common/contributionUtils';
 import { EnvironmentVars } from '../../common/environmentVars';
 import { EventEmitter, IDisposable } from '../../common/events';
@@ -20,6 +23,7 @@ import {
   ILaunchResult,
   IStopMetadata,
   ITarget,
+  IWebViewConnectionInfo,
 } from '../../targets/targets';
 import { TelemetryReporter } from '../../telemetry/telemetryReporter';
 import { baseURL } from './browserLaunchParams';
@@ -27,6 +31,8 @@ import { BrowserSourcePathResolver } from './browserPathResolver';
 import { BrowserTarget, BrowserTargetManager } from './browserTargets';
 import findBrowser from './findBrowser';
 import * as launcher from './launcher';
+import { WebSocketTransport } from '../../cdp/transport';
+import { getDeferred } from '../../common/promiseUtil';
 
 const localize = nls.loadMessageBundle();
 
@@ -80,6 +86,7 @@ export class BrowserLauncher implements ILauncher {
     cancellationToken: CancellationToken,
     telemetryReporter: TelemetryReporter,
     clientCapabilities: IDapInitializeParamsWithExtensions,
+    promisedPort?: Promise<number>,
   ): Promise<launcher.ILaunchResult> {
     let executablePath: string | undefined;
     if (executable && !chromeVersions.has(executable)) {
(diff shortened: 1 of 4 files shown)


src/targets/browser/browserLauncher.ts
```diff
@@ -140,25 +147,80 @@ export class BrowserLauncher implements ILauncher {
...
+    return promisedPort.promise;
   }
 
   async prepareLaunch(
     params: IChromeLaunchConfiguration,
     { dap, targetOrigin, cancellationToken, telemetryReporter }: ILaunchContext,
     clientCapabilities: IDapInitializeParamsWithExtensions,
   ): Promise<BrowserTarget | string> {
+    const filter = createTargetFilterForConfig(params);
+    const promisedPort =
```
EricCornelson: It looks like this has some compiler errors since if `params.useWebView` evaluates to false `promisedPort` will be a boolean, but if it's true `promisedPort` will be a `Promise<number>`.

In any case, does the webview logic belong here, since it's very specific to Edge? I'm thinking it would be cleaner and easier to maintain by maybe creating an `EdgeLauncher` class which extends from `BrowserLauncher` which can take care of Edge specific cases and leave `BrowserLauncher` more agnostic, and also avoid the Edge webview logic causing any unintended side effects when launching Chrome.

johnemau: >
> It looks like this has some compiler errors since if params.useWebView evaluates to false promisedPort will be a boolean, but if it's true promisedPort will be a Promise<number>.

This is actually OK in TypeScript, the `promised` port becomes a union type of `false | Promise<number>` which must be checked for truthiness before we can use it as a promise, [check out this example on the TS playground][1].

> In any case, does the webview logic belong here, since it's very specific to Edge?

I agree that a better home for the web view logic will eventually be another class, ideally it's own even separate from Edge, that said your suggestion of an `EdgeLauncher` would be a move in the right direction.

My intention with this change is to get the feature working with the existing classes as a first step and benefit to users of the extension and I created issue #266 to follow-up track splitting out Edge from Chrome debugger targets, addressing the concerns your mentioned.

[1]: https://www.typescriptlang.org/play/#code/CYUwxgNghgTiAEYD2A7AzgF3hkmBc8ARkkhCFCgNwBQ1y6WAlgAoxIC2jaCAvNrlgBkg+ChAB3eKw5cQAHkwxGKAOYA+ABQx4PNfBgaA5KTIALQwEoLNahqhoAnijDwNFnXoDe1eL-hksFTgQDGUVAkUwmz94IJAQsJ14KHEoRiZpTm5KGPgAejzRJCwkAGsfP0YAM1cWNiyQd29cvziE1SSUtIz62RyC+DKK3wBfWhHrWiA

EricCornelson: > ...and I created issue #266 to follow-up track splitting out Edge from Chrome debugger targets

Fair enough

> This is actually OK in TypeScript, the promised port becomes a union type of false | Promise<number>

While true, I'd still argue that it's confusing code. Also, I get compile errors on this line when I try to build your PR branch locally, but CI seems to build fine 🤷‍♂ (the error I get is actually on the filter param, not the promisedPort type)
image

brendyna:
@johnemau do we normally show an overlay on top of the WebView control when a breakpoint is hit? We do that in the F12 experience, I'm curious and it looks to be missing from the demo gif.

src/targets/browser/browserLauncher.ts
johnemau: > While true, I'd still argue that it's confusing code. 

Thanks for the feedback on this, I coded it for brevity but agree that overlying on short-circuits reduces readability. I will rewrite it with an `if` statement.

> I get compile errors on this line when I try to build your PR branch locally, but CI seems to build fine 🤷‍♂

This is because my change relies on #283 which I haven't merged in from master on my branch, sorry about that! CI passes because it is running on the merge of my branch + master.

I will rebase my change with your feedback and master so it will build locally (thanks for testing!).

johnemau: Done!

johnemau:
> @johnemau do we normally show an overlay on top of the WebView control when a breakpoint is hit? We do that in the F12 experience, I'm curious and it looks to be missing from the demo gif.

Good eye @brendyna I didn't notice this earlier and we do normally show an overlay over the WebView control just like the F12 experience in the browser.

I am able to reproduce the issue on my local installs of Edge (Chromium) and Chrome so this looks like a issue not specific to the vscode-js-debug extension.

To reproduced the issue used the F12 tools to debug a page with a `debugger` statement, after refreshing the page I expected to to be paused with an overlay but no overlay was present.

I will follow-up with a crbug after more investigation tomorrow.

johnemau:
> I will follow-up with a crbug after more investigation tomorrow.

Here is the bug: https://crbug.com/937561#c7 
FYI @brendyna 

To be clear the bug is not related to this PR.

src/targets/browser/browserLauncher.ts
```diff
@@ -215,16 +279,21 @@ export class BrowserLauncher implements ILauncher {
...
+    const mainTargetPromise = this._targetManager.waitForMainTarget(filterForMain);
+
+    if (params.useWebView === 'advanced') {
+      this._mainTarget = await mainTargetPromise;
+    } else {
+      this._mainTarget = await timeoutPromise(
+        mainTargetPromise,
+        cancellationToken,
+        'Could not attach to main target',
+      );
```
connor4312: I'm not sure this needs to be a separate codepath. People using webview, or the EdgeConfigurationProvider (once made) could just set this timeout to an arbitrarily high number. Or maybe we support `0` to be a never cancelled timeout, at the global level.

johnemau: That sounds good to me, I just pushed a commit that removes the timeout-free code path.

I originally added the timeout-free code path for advanced web view debugging to match vscode-edge-debug2 behavior but prefer this with the instruction to users to set the timeout value for now.

FYI @jalissia

johnemau:
> before we add too much more I'd like to get a split going.

Understood, this is the last change I see us needed to achieve full feature parity with vscode-edge-debug2 for web view debugging and agree the next step for web views and edge debugging is to refactor with issue #266.

Thanks everyone for the reviews! From my side this PR is ready to land.

EricCornelson:
> Here is the bug: https://crbug.com/937561#c7

Hey, that bug looks familiar! 😄

src/targets/browser/browserLauncher.ts
connor4312: I'll defer to you about whether removing the default timeout is generally desirable. If it is then I agree that it should be the default/not require user instruction, and be applied in the configuration resolver 🙂

PR #288, opened 2020-01-31
feat: use wasm content hash implementation

connor4312:
The hash speed was bugging me, particularly since we're soon going to
need to hash some files twice (as of #277) so last night I wanted to see if a
webassembly version would fare better. Turns out it does. Hashing 1KB:

```
              7,520 ops/sec > current (1x)
             72,500 ops/sec > wasm (9.64x)

  Benches: 2
  Fastest: wasm
  Elapsed: 10.9s
```

The current JS algorithm only is hasing 7.5MB/s, which may sound like
a decent amount, but remember some node_modules can be massive. This
brings up a bit.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -225,7 +225,9 @@ gulp.task('package:copy-extension-files', () =>
         base: buildDir,
       },
     ),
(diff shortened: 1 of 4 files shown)


EricCornelson:
Nice 👍

PR #292, opened 2020-02-03
feat: assign workspace folders for attaching to processes

connor4312:
Previously if you attached to a process in a multi-root workspace, we'd
fail because we needed a working directory. This adds functionality
(which only works on *nix, unfortunately) that automatically tries to
assign the workspace folder from the working directory of the process.

Also cleans up handling of process IDs a little bit; these can be made
simpler now that we no longer need to support
the legacy debugger protocol.

### DIFF
--- a/src/test/node/process-tree.test.ts
+++ b/src/test/node/process-tree.test.ts
@@ -45,6 +45,13 @@ describe('process tree', () => {
     expect(data[0].command).to.match(/./); // not empty string
   });
 
+  if (process.platform !== 'win32') {
+    it('gets the working directory', async () => {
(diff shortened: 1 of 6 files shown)


roblourens:
I don't quite understand, what was the issue before? You would have a workspace folder from the launch config right?

connor4312:
This is going through the "Attach to Node.js Process" command, which boots without a launch config. It works fine today if you have a single workspace folder open, but if you have multiple ones open we error out--it's needed in config to resolve source paths, do BP prediction, etc.

roblourens:
Oh, got it.

PR #293, opened 2020-02-04
fix: removed redundant useWebView: 'advanced' configuration

johnemau:
This update reflects that the code paths of `useWebView: 'true'` and `useWebView: 'advanced'`  have merged in #285 and now both support the `urlFilter` property and honor the `timeout` property.

### DIFF
--- a/src/build/strings.ts
+++ b/src/build/strings.ts
@@ -28,7 +28,7 @@ const strings = {
   'extensionHost.snippet.launch.label': 'VS Code Extension Development',
 
   'edge.useWebView.description':
-    "(Edge (Chromium) only) When 'true', the debugger will treat the runtime executable as a host application that contains a WebView allowing you to debug the WebView script content. When set to 'advanced', the debugger will wait to connect to a WebView matching the urlFilter.",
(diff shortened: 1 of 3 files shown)


src/build/strings.ts
```diff
@@ -28,7 +28,7 @@ const strings = {
   'extensionHost.snippet.launch.label': 'VS Code Extension Development',
 
   'edge.useWebView.description':
-    "(Edge (Chromium) only) When 'true', the debugger will treat the runtime executable as a host application that contains a WebView allowing you to debug the WebView script content. When set to 'advanced', the debugger will wait to connect to a WebView matching the urlFilter.",
+    "(Edge (Chromium) only) When 'true' or 'advanced', the debugger will treat the runtime executable as a host application that contains a WebView allowing you to debug the WebView script content.",
```
connor4312: If they're now the same, do we need two options or can it just be true/false?

johnemau: Good point, I was keeping `advanced` around for backwards compatibility with vcode-edge-debug2 configs but it will be less confusing for users if we just drop `advanced`.

I will update this PR to do that!

connor4312: Sounds good, thanks. For this extension we don't care about back compat with edge-debug (only for node and chrome debug)

johnemau: Done!

PR #294, opened 2020-02-04
fix: Move common js entrypoints to a single spot in the gulpfile.

EricCornelson:
VS was broken because it wasn't including hash.js in the bundled build. Side node, it might be nice to have some fail-fast way of knowing that the hashing subprocess is broken, as it's hard to diagnose at the moment (no error messages, things like stackTrace requests just start hanging because the absoluteFilePath promise never returns).

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -166,6 +166,14 @@ gulp.task('compile:static', () =>
 gulp.task('compile', gulp.series('compile:ts', 'compile:static', 'compile:dynamic'));
 
 async function runWebpack(packages) {
+
+  // add the entrypoints common to both vscode and vs here
+  packages.push(...[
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -166,6 +166,14 @@ gulp.task('compile:static', () =>
 gulp.task('compile', gulp.series('compile:ts', 'compile:static', 'compile:dynamic'));
 
 async function runWebpack(packages) {
+
+  // add the entrypoints common to both vscode and vs here
+  packages.push(...[
```
connor4312: I'd suggest this to avoid mutating the input, just in case anyone tries to use it later

```suggestion
  packages = [
      ...packages,
```

PR #298, opened 2020-02-04
fix: spaces in paths breaking electron apps

connor4312:
One more super annoying thing. _Apparently_ the debugger in electron doesn't url encode spaces in paths. This was causing the setup on my windows VM to never hit breakpoints, since it was looking for a path with %20 in it.

image

It looks like the old debugger didn't escape spaces at all. Maybe we should unescape everything/apply this same treatment to all characters. I don't want to modify URLs throughout the adapter to be unescaped, but this function could handle that for the purposes of matcing. Thoughts @roblourens?


roblourens:
Did this only affect file: URIs with the old debugger, because it definitely works with spaces in URLs in general

connor4312:
Yea it looked like it was only file URIs.

Issue #291, opened 2020-02-03
Can't set "program" to source file


roblourens:
- Create a node launch config where the "program" attribute points at a .ts file
- It tries to run the .ts file with node

The v1 DA tries to resolve the .ts file when `outFiles` is given, and otherwise will refuse to launch. We don't necessarily need the behavior to guess whether a file is a .js file because it can't be 100% accurate (https://github.com/microsoft/vscode/issues/89882) but we should at least resolve .ts files with the sourcemaps we found, for backcompat.

connor4312:
I'm not totally sure I agree with this. For instance someone could use `ts-node` as their runtime executable and be to set `program` to a `.ts` file.

roblourens:
In that case (ideally) there would be no sourcemaps present so we would pass the .ts file through.

This might be disabled in v1 if `runtimeExecutable` is set?

roblourens:
That's not the case. People use `args` with ts-node in v1 to work around the "program" restriction.

connor4312:
I feel like people using TypeScript would generally know that they can't generally run uncompiled code directly in Node. Not sure that auto-fixing the couple of cases where they don't is worth the need for a workaround + complexity in the debug adapter. Maybe one alternative would be providing an extra helpful message if we see a syntax error in the program's stdout combined with a non-`.js` extension?

roblourens:
Even though this has always been a little too "magical" for my taste, I think we should support it just to be compatible with old launch configs.

connor4312:
Mm. Maybe try to do this but just go ahead and run the program if we don't see a sourcemap, to drop the need for a workaround?

PR #304, opened 2020-02-05
Service tree

connor4312:
This PR adds a container for services that are meant to be shared between multiple debug sessions. The top-level factory takes the launch params and parent DAP for a debug session. The returned `IServiceFactory` allows `create`ing services and creating `.child` factories. This provides a nicer, standard way to share services and communication between debug sessions. I've just moved the breakpoint predictor into here for now, since that's what I need for #291. The script skipper is also a good candidate to add.

### DIFF
--- a/src/test/variables/variablesTest.ts
+++ b/src/test/variables/variablesTest.ts
@@ -106,7 +106,11 @@ describe('variables', () => {
       const worker = await r.worker();
       outputs.push({ output: await worker.dap.once('output'), logger: worker.logger });
       outputs.push({ output: await worker.dap.once('output'), logger: worker.logger });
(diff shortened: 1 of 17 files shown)


src/test/variables/variablesTest.ts
```diff
@@ -106,7 +106,11 @@ describe('variables', () => {
       const worker = await r.worker();
       outputs.push({ output: await worker.dap.once('output'), logger: worker.logger });
       outputs.push({ output: await worker.dap.once('output'), logger: worker.logger });
-      outputs.sort((a, b) => a.output.source!.name!.localeCompare(b.output.source!.name!));
+      outputs.sort((a, b) => {
```
connor4312: noticed this test had flaked on master with `cannot read property name of undefined`

Issue #300, opened 2020-02-05
Issues with invalid hit count bp


roblourens:
- Set a hit count bp with invalid hit count
- In the same file, set a normal bp
- The hit count bp is moved to the line of the normal bp but we do hit the normal bp as expected, seems like they are getting switched around

![image](https://user-images.githubusercontent.com/323878/73800256-0179fa80-476d-11ea-9f8a-786228d5ca9f.png)

![image](https://user-images.githubusercontent.com/323878/73800297-1eaec900-476d-11ea-9c60-6803dce85090.png)

![image](https://user-images.githubusercontent.com/323878/73800273-10f94380-476d-11ea-85b5-4efa7d79d8d9.png)

[vscode-debugadapter.json.txt](https://github.com/microsoft/vscode-js-debug/files/4156866/vscode-debugadapter.json.txt)

connor4312:
This is because when we fail to parse a hit condition, we simply skip the given breakpoint. If the breakpoint is first in the list, this causes the editor to think that the breakpoint we resolved at the given location was the hit count breakpoint. 

![image](https://user-images.githubusercontent.com/2230985/73889508-b7545000-4824-11ea-8b7d-5c4a24453730.png)

We already show a message in the console when we can't parse it, so I think the correct behavior here would be to keep the breakpoint around as unverified.

PR #305, opened 2020-02-05
fix: weirdness around invalid hit count breakpoints

connor4312:
See: https://github.com/microsoft/vscode-js-debug/issues/300#issuecomment-582654597

This simply uses a new subtype "never resolved breakpoint" which has
an always-false hit condition and overrides the resolved location to
always return undefined. We use that to keep the invalid hit count
breakpoints around. Perhaps not the most efficient way since the
breakpoint still exists in the debugee, but definitely the simplest way
that doesn't involve a ton of refactoring and special casing.

Also, noticed I never wrote unit test for hit count breakpoints, so I
added some of those.

### DIFF
--- a/src/test/breakpoints/breakpointsTest.ts
+++ b/src/test/breakpoints/breakpointsTest.ts
@@ -2,7 +2,7 @@
  * Copyright (C) Microsoft Corporation. All rights reserved.
  *--------------------------------------------------------*/
 
-import { testWorkspace, ITestHandle, createFileTree } from '../test';
+import { testWorkspace, ITestHandle, createFileTree, TestRoot, TestP } from '../test';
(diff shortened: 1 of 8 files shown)


src/test/breakpoints/breakpointsTest.ts
```diff
@@ -618,4 +618,77 @@ describe('breakpoints', () => {
       );
     });
   });
+
+  describe('hit count', () => {
```
roblourens: My PR has tests for these too 🤷‍♂

connor4312: If we test it twice that means it's twice as reliable!

I'll merge my test with yours once both our PRs are in master.

PR #306, opened 2020-02-06
Clean up int-chrome tests, port some

roblourens:
I went through all the tests in int-chrome and determined that they were either covered by normal tests that we already have or are not that valuable, except for a few that I brought over for conditional/hit count BPs. 

I added a test that will run with the latest create-react-app. Wanted to use a nightly/alpha/canary version but it seems like they don't consistently define it or keep their tags up to date:

image

And currently it runs as a separate task in the normal CI pipeline but we can see how slow/flaky it actually is and move it to another pipeline if needed.


roblourens:
I can't figure out why I can't spawn npm on windows, so I disabled it on windows.

src/test/framework/react-hit-breakpoint.txt
```diff
@@ -0,0 +1,28 @@
...
+legacyRenderSubtreeIntoContainer @ ${fixturesDir}/react-test/node_modules/react-dom/cjs/react-dom.development.js:27527:1
+render @ ${fixturesDir}/react-test/node_modules/react-dom/cjs/react-dom.development.js:27608:1
+./src/index.tsx @ ${fixturesDir}/react-test/src/index.tsx:7:1
+__webpack_require__ @ ${fixturesDir}/react-test/webpack/bootstrap:785:1
+fn @ ${fixturesDir}/react-test/webpack/bootstrap:150:1
+1 @ localhost꞉3000/static/js/main.chunk.js:376:18
+__webpack_require__ @ ${fixturesDir}/react-test/webpack/bootstrap:785:1
+checkDeferredModules @ ${fixturesDir}/react-test/webpack/bootstrap:45:1
+webpackJsonpCallback @ ${fixturesDir}/react-test/webpack/bootstrap:32:1
+<anonymous> @ localhost꞉3000/static/js/main.chunk.js:1:73
```
connor4312: I would make this a substring test and just assert the `App @` line since internals will change with React updates

roblourens: Good point

PR #312, opened 2020-02-06
refactor: get rid of global logger, move to service tree

### LINKED ISSUE #299: Log lines lost after first debug session
I guess I haven't noticed this before but maybe you are aware of it. On the second and later debug sessions, we lose messages from the debug log because we call `logger.setup` a second time but the logger is not "reset" to queue messages after the first session.

Also need to think about how it works when two sessions run simultaneously.

### LINKED ISSUE #311: Adapter occasionally writes lots of nulls (several mb worth) to the logs
Repro steps: Unknown. Not able to consistently repro

See attached log

[nulls-vscode-debugadapter.log](https://github.com/microsoft/vscode-js-debug/files/4167429/nulls-vscode-debugadapter.log)

connor4312:
This is _mostly_ skim-able. The biggest changes are around some of the
breakpoint condition stuff to move them to be more class-based where we
could inject a logger at instantiation time. Also now, hit count
breakpoints, log points, and conditional breakpoints are represented by
the same interface, which streamlines some code.

- Fixes #311 -- Pretty sure this was weird stuff with multiple consumers
  appending/closing/opening a single file concurrently.
- Fixes #299 -- Log sessions have their own log files. We'll in fact
  never overwrite previous files, incrementing a log number each time.
  I wanted to check with @roblourens to see/make sure that VS Code
  prunes its logs storage folders occasionally, otherwise we'll be
  stashing tons of logs that never get deleted.

Tomorrow I'll probably play with inversify here to get an automatic IOC
system going. The wiring of services everywhere, particularly the
telemetry and logging services, is getting somewhat messy.


src/adapter/breakpointPredictor.ts
```diff
@@ -61,8 +61,13 @@ export class BreakpointsPredictor {
     );
   }
 
-  @logPerf<BreakpointsPredictor>(m => ({ type: m.repo.constructor.name }))
   private async createInitialMapping(): Promise<MetadataMap> {
+    return logPerf(this.logger, `BreakpointsPredictor.createInitialMapping`, () =>
```
connor4312: Since logging is now dependency injected, we can't use a cute decorator to log perf metrics. Oh well.

roblourens:
> to see/make sure that VS Code
prunes its logs storage folders occasionally, otherwise we'll be
stashing tons of logs that never get deleted.

We are writing these to tmpdir though right?

Issue #302, opened 2020-02-05
[Blazor] allow inspectorUri in Chrome launch configurations


connor4312:
To debug Blazor apps, we need to use its CDP-speaking proxy service, which runs at a separate address than whatever browser we launch. Therefore, the inspector uri needs to be configurable. We suggest adding an additional option to the Chrome launch configuration:

```ts
export interface IChromeLaunchConfiguration {
  /**
   * WebSocket (`ws://`) address of the inspector. It's a template string that
   * interpolates keys in `{curlyBraces}`. Available keys are:
   *
   *  - `url.*` is the parsed address of the running application. For instance,
   *    `{url.port}`, `{url.hostname}`
   *  - `port` is the debug port that Chrome is listening on.
   *  - `browserInspectUri` is the inspector URI on the launched browser
   *  - `wsProtocol` is the hinted websocket protocol. This is set to `wss` if
   *    the original URL is `https`, or `ws` otherwise.
   */
  inspectUri?: string;
```

Upon launch, if we see that the inspectUri is provided, we will launch chrome with a `--remote-debugging-port` (instead of a pipe), render that string, then create a WebSocket connection to it. If the connection fails, we will immediately abort the launch with an appropriate error.

```ts
const templateString = (input, replacement) =>
  input.replace(
    /{(.*?)}/g,
    (match, key) => replacement.hasOwnProperty(key) ? replacement[key] : match,
  );
```

One convenient thing about this approach over the original one (below) is that additional options can be passed to the proxy without needing changes in VS code. For example, if we wanted to pass whether the user has enabled telemetry as a parameter, that could be done without a js-debug change. It may also be used in case other X-to-JavaScript/WASM frameworks/tools do similar things in the future, as they may have their own setups.

*Open question:* this partially duplicates some options in the Chrome attach scenario. Should we try to unify these options?

<details><summary>Original proposal (email)</summary>

* VS sees that `launchSettings.json` contains a value for `debugEndpoint`. Example:
  `"debugEndpoint": "/_framework/debug/ws-proxy"`
  This causes it *not* to connect the JS debugger to <browserCdpEndpoint> directly as it does today,
  but instead follows this alternate flow.
 
* VS defines <applicationCdpEndpoint> = `ws(s)://<applicationUrl><debugEndpoint>?browser=<browserCdpEndpoint>`
[...]

PR #313, opened 2020-02-06
Add support for custom inspect uri format for Blazor

digeff:
Add support for custom inspect uri format for Blazor

See https://github.com/microsoft/vscode-js-debug/issues/302 for more details

Added some extra logging that I used while debugging this feature

cc @connor4312 @SteveSandersonMS @roblourens

### DIFF
--- a/src/cdp/transport.ts
+++ b/src/cdp/transport.ts
@@ -29,6 +29,7 @@ export class PipeTransport implements ITransport {
   constructor(pipeWrite: NodeJS.WritableStream, pipeRead: NodeJS.ReadableStream);
 
   constructor(pipeWrite: net.Socket | NodeJS.WritableStream, pipeRead?: NodeJS.ReadableStream) {
+    logger.info(LogTag.Internal, `Using pipe as transport layer`);
     this._pipeWrite = pipeWrite as NodeJS.WritableStream;
     if (!pipeRead) this._socket = pipeWrite as net.Socket;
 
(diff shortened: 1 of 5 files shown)


connor4312:
Changes look good to me. Could we get an integration test here? Could be something super synthetic like standing up a `ws` server in the test, passing a config like

```
inspectUri: `ws://127.0.0.1:${testServer.port}?port={url.port}&...
```

and then just waiting for any connection and checking the query params.

digeff:
Can we merge this change in now so it's easier for people to start testing it, and I'll work on the integration tests meanwhile?

roblourens:
Would also just like a basic unit test of the parsing logic in `constructInspectorWSUri`. I can merge it now though.

EricCornelson:
Should we also add usage notes in the package.json? Or is this meant to be a "hidden" launch parameter?

roblourens:
Oh, they will get a warning if it isn't registered in the package.json, which we probably don't want.

SteveSandersonMS:
@digeff @connor4312 How does one signal that the browserInspectUri needs to be uri-component encoded when being inserted into the query string part of a URL?

SteveSandersonMS:
Maybe there should be a {browserInspectUriEncoded} parameter to use for this.

connor4312:
An `{browserInspectUriEncoded}` parameter would work for me.

PR #314, opened 2020-02-06
Delay the thread started event until we get the 'initialized' request from vscode

### LINKED ISSUE #281: Thread events coming in before the response to initialize crashes the debug session in Visual Studio
Thread events are send out immediately when a new session attaches, which often means they go out before the IDE host has sent the initialize request. This can often crash the debug session in VS, as the thread's created event comes in before the session is actually started in VS which means it gets missed, then later when the adapter sends the thread exit event VS crashes because it can't find the exiting thread.

Logs:
[vscode-debugadapter-early-thread-events.log](https://github.com/microsoft/vscode-js-debug/files/4131983/vscode-debugadapter-early-thread-events.log)

roblourens:
@connor4312 you are more familiar with the startup than I am, but basically the `acquireDap` call is resolved as soon as the client tries to attach, then there is no guarantee that we will wait for 'initialize' before the Thread sends its started event. This is an easy fix to delay that until 'initialize' happens, but you might have a better idea?

Fix #281

### DIFF
--- a/src/adapter/threads.ts
+++ b/src/adapter/threads.ts
@@ -4,7 +4,7 @@
 
 import * as nls from 'vscode-nls';
 import Cdp from '../cdp/api';
-import { delay, getDeferred } from '../common/promiseUtil';
+import { delay, getDeferred, IDeferred } from '../common/promiseUtil';
 import * as sourceUtils from '../common/sourceUtils';
 import * as urlUtils from '../common/urlUtils';
 import { AnyLaunchConfiguration, OutputSource } from '../configuration';
@@ -85,10 +85,27 @@ export type RawLocation = {
   scriptId?: Cdp.Runtime.ScriptId;
 };
 
+class DeferredContainer<T> {
+  private _dapDeferred: IDeferred<T> = getDeferred();
+
+  constructor(private readonly _obj: T) {}
+
+  resolve(): void {
+    this._dapDeferred.resolve(this._obj);
+  }
+
+  with<Return>(callback: (obj: T) => Return): Return | Promise<Return> {
+    if (this._dapDeferred.hasSettled()) {
+      return callback(this._obj);
+    } else {
+      return this._dapDeferred.promise.then(obj => callback(obj));
+    }
+  }
+}
+
 export class Thread implements IVariableStoreDelegate {
   private static _lastThreadId = 0;
   public readonly id: number;
-  private _dap: Dap.Api;
   private _cdp: Cdp.Api;
   private _name: string;
   private _pausedDetails?: IPausedDetails;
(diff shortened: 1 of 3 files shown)


roblourens:
Maybe we could delay calling `debugAdapter.createThread` at all until 'initialize'

EricCornelson:
This would solve specifically the thread events going out before the init response, but not all cases of events going out before init. I'm also seeing output events go out early.

Does it make sense to buffer messages at the connection level and block them from going out until after the init response?

roblourens:
Actually, I think I just need to move all of `_initialize` into this delayed method rather than just the one line.

roblourens:
You're right that in theory something somewhere else could still send something over the Dap earlier but I'd rather line it up in predictable order rather than buffer/resend events.

EricCornelson:
That makes sense. What about adding an assert to verify and catch any possible future instances?

src/adapter/threads.ts
```diff
@@ -403,7 +402,7 @@ export class Thread implements IVariableStoreDelegate {
     }
   }
 
-  private _initialize() {
```
connor4312: Delaying attachment of these event listeners seems dangerous to me. E.g. is there a case where the user has a debugger statement in their script which gets hit almost immediately and we don't listening for the `paused` event in time?

roblourens: Those events should be send when we send `Debugger.enable` and not before.

roblourens: Well, if it hits a debugger statement before `Debugger.enable` will it still pause? I guess I don't know for sure.

roblourens: If we want to do that ASAP how about we make `this.dap` a deferred which will be resolved on 'initialize'? Then we can hook up handlers and they can send events as normal and they will effectively be queued and released after initialize.

EricCornelson: Won't that be caught by the fact that we start off at about:blank, send all the config (such as Debugger.enable) and _then_ navigate to the user's actual page?

Not as confident about what happens on the Node side of things, but I believe we always start paused there anyway right?

connor4312: Yea, V8 won't pause unless there's a debugger attached and enabled. I think you can simulate races pretty well by putting `await delay(1000)` at the top of `initialize`. For me this does cause us to miss breakpoints early on in the execution.

roblourens:
Pushed a different option that implements what I proposed above. This maintains sync access to the Dap after initialize. Partly out of paranoia about ordering things, partly from breaking tests due to the telemetry events coming before other output events. But then I disabled telemetry during tests which fixed that issue.

If you are uncomfortable with code that is maybe async and maybe sync, I think it would be safe to make it actually async all the time.

PR #316, opened 2020-02-07
refactor: move to inversify for dependency injection

connor4312:
Following #312, we realized service wiring was getting rather messy
and burdensome. This PR brings Inversify in to provide dependency
injection. Inversify support hierarchal containers, which works nicely
for our case where we have global services (e.g. delegate launcher),
top-level session services (e.g. the logger and launchers) as well as
session-specific services.

This PR mostly just moves the services that were previously in our
ad-hoc services.ts into Inversify. It's my plan that as we touch things
and expand on code, we can move more services into the container.

---

Also, I'm not super set on Inversify and am open to pushback if we think this is not a good direction to take the codebase.


src/ioc.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  return container;
+};
+
+export interface IRootOptions {
+  delegateLauncher: DelegateLauncherFactory;
+}
+
+/**
+ * Creates the
```
roblourens: the what???

connor4312: we may never know 👻

PR #320, opened 2020-02-10
Add tests and documentation for inspect uri

digeff:
Added an integration tests, some unit tests, encoded the nested uri and documentation for inspect uri
More details at https://github.com/microsoft/vscode-js-debug/issues/302

cc @connor4312 @roblourens @SteveSandersonMS


package.json
```diff
@@ -118,6 +118,7 @@
     "http-server": "^0.11.1",
     "husky": "^3.1.0",
     "lodash": "^4.17.15",
+    "long": "^4.0.0",
```
connor4312: accidental?

src/build/generate-contributions.ts
```diff
@@ -578,6 +578,12 @@ const chromeBaseConfigurationAttributes: ConfigurationAttributes<IChromeBaseConf
     description: refString('edge.useWebView.description'),
     default: false,
   },
+  inspectUri: {
+    type: ['string', 'null'],
+    description: refString('chrome.inspectUri.description'),
+    default:
+      '{wsProtocol}://{url.hostname}:{url.port}/_framework/debug/ws-proxy?browser={browserInspectUri}', // The default value used to debug Blazor
```
connor4312: I might just leave this out since it'll be autogenerated by e.g. the C# extension generally

package.json
digeff: At a point I was getting a runtime failure because this was missing. Maybe it's not needed any more

src/build/generate-contributions.ts
digeff: You mean the whole parameter or only the default value?

package.json
connor4312: Yea we used it for hashing before but that moved to WebAssembly

src/build/generate-contributions.ts
connor4312: The default value

package.json
digeff: Removed

src/build/generate-contributions.ts
digeff: Removed

Issue #267, opened 2020-01-27
Support Edge: Support finding Edge (Chromium) executables via `findBrowser.ts`


johnemau:
#### Background
This is part of the effort to achieve backwards compatibly with [vscode-edge-debug2] and [fully support debugging Microsoft Edge (Chromium)][1].

Edge (Chromium) is installed in well-known locations similar to Chrome.

#### What needs to happen
*  Update `findBrowser.ts` to check if the debug type is `edge` and search for the installed versions. See [this comment for some additional details][2].

Related: #235

[vscode-edge-debug2]: https://github.com/microsoft/vscode-edge-debug2
[1]: https://github.com/microsoft/vscode-js-debug/issues/235
[2]: https://github.com/microsoft/vscode-js-debug/pull/264#issuecomment-578868423

PR #321, opened 2020-02-10
refactor: split edge into separate config, its own launcher

### LINKED ISSUE #266: Support Edge: Add unique launch/attach configurations (with strings) for `edge` debugger type

connor4312:
Also, moves configuration provides to inversify as well. The split right
now is pretty basic, just doing the minimium to move the
webview-specific code out of the browser launcher. With #267 they will
diverge a little more.

There's a decent amount of churn around configs here. But the way we
have them set up they're well typed-checked, so it's fairly safe.

Fixes #266

@johnemau I don't have a webview setup (and integration tests are still on the way 🙂 ) if you could give this PR a sanity check some time before Friday that would be awesome.


johnemau:
:shipit: LGTM, I verified this works with the webview demo.

image

I did get some typing issues and a timeout when I merge in my webview test PR (#325) but I am ignoring those for now because that PR has not landed and the demo is the better test.

Thanks for doing this!

PR #325, opened 2020-02-11
add test for webviews

### LINKED ISSUE #303: Tests: Add integration test for WebView launch scenario
### Steps
1. Create a simple WebView2 application and include it in the project.
2. Have the application launch and hit a break-point as an integration test, similar to the int-chrome tests.

### Platforms
Windows 10 is currently the only supported platform for WebView2. 

### Background
WebView2 debugging support landed in #264, and #285.

johnemau:
This change adds an integration test file `webview.breakpoints.test.win.ts` for launching and debugging a script file in a webview.

- WebViews are currently only supported on the [Windows platform][1] `testRunner.ts` was changed to only add tests with the suffix `.test.win.js` when run on the Windows platform.

- WebViews require Edge (Chromium) to be installed on the host VM a step was added to the pipeline that downloads and installs the latest Edge Canary from https://www.microsoftedgeinsider.com/en-us/download/ Once the test VM image include Edge (Chromium) this step can be removed.

- Updated `demos\webview` to build a portable release version of WebView2Sample.exe which is used in the test, this change was based on [updates to the WebView2Sample project][2] .

- Updated `binder.ts` and `test.ts` to pass the current `ILauncher` down during attach in tests instead of assuming to use the `ChromeLauncher`. This was needed to support the `EdgeLauncher` in tests.

### Follow-ups to consider
- Update the test helpers to [support `AnyChromiumConfigurations`][3].
- [Snapshot the Edge install version][4] or switch from canary to stable to limit the chance of a bad Edge build bringing down the CI.
- Add demo and test for multi-browser process webview launch scenario (#285)
- Move the webview tests into their own suite that builds the WebView2Sample.exe from source and nuget packages.

Fixes #303 

[1]: https://docs.microsoft.com/en-us/microsoft-edge/hosting/webview2#supported-platforms
[2]: https://github.com/MicrosoftEdge/WebView2Samples/pull/16
[3]: https://github.com/microsoft/vscode-js-debug/pull/325/files#diff-fec9b7364e826cfd2a863605fbec1829R27
[4]: https://github.com/microsoft/vscode-js-debug/pull/325/files#diff-3db134b5ae894e84d07e08d5893eb937R25


johnemau:
I am having a hard time reading the [check results][1] and confirming the new test ran on the Windows pipeline.

[1]: https://github.com/microsoft/vscode-js-debug/pull/325/checks

connor4312:
It doesn't look like it did. The best way is to just hit "view raw log" and find in there or open that in your favorite editor. You can also use VSO's [fancy tests view](https://dev.azure.com/vscode/VS%20Code%20debug%20adapters/_build/results?buildId=36999&view=ms.vss-test-web.build-test-results-tab). 

Maybe the easiest way would be to just wrap the describe block in `if (process.platform === 'win32')`. We do that for some other tests already. Not the coolest way but it does the job :)

johnemau:
The test it fails with a timeout only in CI, I suspect the WebView2Sample.exe is not launching successfully due to the VM setup but I can not reproduce locally and will need either better logging or to recreating the test VM locally to move forward with getting it running under CI.

In the mean time I would like to get the test merged in so it can at least be run locally.

I set the test to `.skip` and removed the CI step to install Edge canary with the intention of someone (me if possible) re-enabling the test the future (we can keep the issue open and update it to tracking CI support for the test).

connor4312:
thanks, I'll see if I can play around and get CI working

connor4312:
The tests time out on my (relatively fresh) windows VM, maybe for the same reason as the test runner is failing.

image

johnemau:
Thanks @connor4312, that is the clue I needed!

These are missing dependencies from VS 2019, I will look into getting them checked in under the `testWorkspace/webview/win` directory and hopefully that will resolve the CI issue.

johnemau:
[Screen Shot 2020-02-11 at 11 16 49 PM][1]

[1]: https://dev.azure.com/vscode/VS%20Code%20debug%20adapters/_build/results?buildId=37139&view=ms.vss-test-web.build-test-results-tab&runId=1720528&resultId=100279&paneView=debug

connor4312:
Nice! Is this PR ready to be merged then?

johnemau:
> Nice! Is this PR ready to be merged then?

@connor4312, yes this PR is ready to merge! :shipit:

johnemau:
@connor4312, @EricCornelson, @roblourens - Now that this PR and #328 have merged we can add Edge (Chromium) tests on the Windows platform by creating test files with the suffix `*.test.win.ts` and using `DebugType.Edge`.

Devs will need to have Edge on their machine to get the tests to pass locally: https://www.microsoftedgeinsider.com

PR #330, opened 2020-02-12
Refactor SessionManager to extract vscode specific code

EricCornelson:
This change introduces the following interfaces to abstract implementation specific behavior in regards to launching/attaching to new DAP debug sessions:

IConnectionStrategy - defines a strategy for obtaining a new DAP connection for a debug session

IDebugSessionLike - abstraction around the platform specific session (e.g. vscode.DebugSession)

SessionLauncher - specifies how to launch a new platform specific debug session (e.g. vscode.debug.startDebugging for vscode)

This change also makes SessionManager and Session generic to the debug session type.


src/sessionManager.ts
```diff
@@ -0,0 +1,238 @@
...
+import { Container } from 'inversify';
+import { createTopLevelSessionContainer } from './ioc';
+import { IChromeAttachConfiguration } from './configuration';
+
+/**
+ * Interface defining a strategy for obtaining a new DAP connection for
+ * a given session.
+ */
+export interface IConnectionStrategy {
+  init(telemetryReporter: TelemetryReporter, logger: ILogger): void;
```
connor4312: It looks like `init()` on the connection strategy can be reused between multiple/nested debug sessions. I think how it exists now might break nested sessions in VS Code. Maybe we should have an `IConnectionStrategyFactory` with an `.obtain()` call that returns something like

```ts
export interface IConnectionStrategy extends IDisposable /* to clean up any server */ {
  connection: Promise<DapConnection>;
}
```

src/cdp/connection.ts
```diff
@@ -71,18 +72,34 @@ export default class Connection {
...
   }
 
+  private GOOD_LOGGER = {
+    log(tag: string, msg: any) {
+      fs.appendFileSync(
+        'C:\\tempz\\goodlog.txt',
+        JSON.stringify({ timestamp: Date.now(), tag, metadata: msg }) + '\n',
+      );
+    },
+  };
```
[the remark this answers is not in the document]
EricCornelson: Dang, thought I was quick enough that you wouldn't see that 😄

src/dap/flatSessionConnection.ts
```diff
@@ -52,11 +55,14 @@ export class ChildConnection extends DapConnection {
   }
 
   _send(msg: Message) {
-    msg.sessionId = this.sessionId;
+    if (this.sessionId !== '') {
```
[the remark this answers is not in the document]
EricCornelson: Good point. That's the way I had it initially (sessionId undefined for the top level connection), but I didn't want to also make `IDebugSessionLike`'s `id` be possibly undefined since we use that as the key for the session map.

Should be able to use undefined for the connection but still pass the `IDebugSessionLike` id as "root" or something.

src/sessionManager.ts
EricCornelson: I don't think it would get reused since we're always creating a new instance in `createDebugAdapterDescriptor`, and the only place `init` is called is in the constructor of the session, which is only ever called from `createSession`. Am I missing something?

I do agree with making the connection strategy extend `IDisposable` though so we don't leak anything.

EricCornelson: I do think it's probably better to add another level of abstraction though to get away from the `init` method, which I'm not a fan of.

EricCornelson: @connor4312 I did a bit of refactoring, but still not fully happy with the connection strategy in general.

Is it ok if we merge this PR as is, and I'll have a follow-up coming shortly to focus on the connection piece?

src/flatSessionLauncher.ts
```diff
@@ -9,128 +9,127 @@ import 'reflect-metadata';
...
-    this.connection.dispose();
-    this._nameChangedSubscription.dispose();
+  private _name: string;
+  set name(newName: string) {
+    this.childConnection
+      .then(x => x.dap())
+      .then(dap => {
+        dap.process({ systemProcessId: this.mockProcessId, name: newName });
+      });
+  }
```
connor4312: Probably update `_name` here too

src/ui/vsCodeSessionManager.ts
```diff
@@ -0,0 +1,96 @@
...
+   * @inheritdoc
+   */
+  public createDebugAdapterDescriptor(
+    debugSession: vscode.DebugSession,
+  ): vscode.ProviderResult<vscode.DebugAdapterDescriptor> {
+    debugSession.workspaceFolder;
+
+    const debugServer = net.createServer().listen(0);
+    const connectionStrat = new SocketServerConnectionStrategy(debugServer);
+    this.disposables.push(connectionStrat);
```
connor4312: Probably want to dispose of this when the session ends, otherwise disposables will keep growing until the extension host is kileld

EricCornelson: Good point, done (also just took the server dispose responsibility out of the connection strategy)

EricCornelson:
Superseded by #338

PR #331, opened 2020-02-12
fix: keep <node_internals> and path data in module names

### LINKED ISSUE #329: js-debug: loaded scripts inconsistencies
VS Code stable

- debug express "hello world" app by setting a breakpoint in app.js:31

js-debug shows these loaded scripts:

whereas node2 shows these loaded scripts:

(the latter is correct)

The main problem is that js-debug does not prefix internal modules (e.g. _http_agent.js, http.js, tty.js) in a way that groups them together under one node in the Loaded Scripts view). As a result they appear intertwined with the user's project files (app.js. bin/www).

Node2 uses the artificial prefixes `<node_internals>` and `<eval>`.

I suggest that js-debug (at least for now) does the same as node2.

The other problem is the bogus debug session "Launch Program" that appears without ever having any scripts under its node. We can address this as part of https://github.com/microsoft/vscode/issues/90518.

connor4312:
Previously we didn't prefix node_internals in the DAP source, and we
also only included the basename in the source `name`, which led to
pretty ugly file trees. Now, the name will be the relative path from
the rootPath, keeping the `<eval>` or `<node_internals>` name if present.

Fixes #329

### DIFF
--- a/src/adapter/breakpointPredictor.ts
+++ b/src/adapter/breakpointPredictor.ts
@@ -103,6 +103,10 @@ export class BreakpointsPredictor implements IBreakpointsPredictor {
       return new Map();
     }
 
+    if (!this.rootPath) {
+      return new Map();
+    }
+
     const sourcePathToCompiled: MetadataMap = new MapUsingProjection(
       urlUtils.lowerCaseInsensitivePath,
     );
--- a/src/adapter/sources.ts
+++ b/src/adapter/sources.ts
@@ -3,7 +3,6 @@
  *--------------------------------------------------------*/
 
 import * as nls from 'vscode-nls';
-import * as path from 'path';
 import { URL } from 'url';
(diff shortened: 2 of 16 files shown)


connor4312:
Huh, I didn't realize that was a thing. One potential quirk with the way the script skipper does it would be if the user is running a different Node version than the version the extension host is running, which could have different files.

Here I've preserved the assumption the old adapters took (which _seems_ correct) that any non-absolute paths we get when running Node code are internals.

roblourens:
It actually evals that inside the debuggee. It is technically possible, but not common, to have non-absolute paths for some other script, probably just when using `//#sourceURL` in an eval script.

connor4312:
Ok, okay.

This code path is only hit if there isn't a sourcemap, so that should not be a problem here.

roblourens:
`sourceURL` lets you set the name of an eval script, and this will be reported as the url, and you won't have a sourcemap when using this

PR #338, opened 2020-02-18
Refactor connection to use a "transport" interface

EricCornelson:
Building on #330, this breaks out a separate "transport" interface for dap connections instead of the "connectionStrategy".

This allows us much more control over the transport layer for DAP, as well as making the ownership/lifecycle of the transport layer more explicit, and further reduces the footprint for potential differences between different modes (e.g. flat session vs. socket), as it does away with the "flatSessionConnection" and "messageEmitterConnection" classes.


src/ioc.ts
```diff
@@ -217,6 +210,13 @@ export const createGlobalContainer = (options: {
   container.bind(Execa).toConstantValue(execa);
   container.bind(FS).toConstantValue(fsPromises);
 
+  // Core services:
+  container
+    .bind(ILogger)
+    .to(Logger)
+    .inSingletonScope()
+    .onActivation(trackDispose);
```
connor4312: The logger is configured by and bound to the parameters in a single session; I don't think it should be created at a global level.

EricCornelson: I suppose we can manually inject the logger in to the transport when we create the top level session, does that work for you?

connor4312: Alternately, maybe the transport could not need a logger and the DAP Connection could do all the necessary logging?

EricCornelson: considered that, but I like keeping it in the transport since that's where we're adding the sessionId for flat session, and if we move it up to connection we won't get the sessionId in logs

connor4312:
Feel free to merge when you're ready

EricCornelson:
Not authorized 😃

PR #348, opened 2020-02-24
feat: pretty printing

connor4312:
This kind of existed before, but it broke a while ago and had some
limitations and bugs for a while. This commit improves the experience,
makes it work pretty well for me locally:

- Only show the pretty print command when a printable file is open in
  a js-debug session.
- Previously, we only suggested printing ephemeral files. Now we run the
  minified check on files that are in the stacktrace when the user
  opens them, which I think is a good compromise between being apparent
  and being too annoying.
- The ability to turn off pretty print suggestions is now a user setting.
- Previously if you were on a breakpoint and then pretty-printed a file,
  continuing to step through would return to the original file, and
  any breakpoints in the minified file would not be updated. We fix that
  by moving the "original location" of breakpoints into the beauitified
  file.
- Added tests for pretty printing.
- There were several edge cases around how we handled pretty print
  commands for file on disk and for multiple debug sessions,
  those should be fixed here.

Something that doesn't work yet is the preservation of breakpoints in
beauitified sources on reload. To do this, I think we need to have some
extra logic to watch for breakpoints set in *-pretty.js files, and then
generate the beautified source automatically when we see a source
matching those breakpoints. For a followup PR.

Demo: https://memes.peet.io/img/bc0c485e-323b-44a7-bc2d-bd274ba2344c.mp4


package.json
```diff
@@ -229,6 +230,13 @@
       }
     ],
     "menus": {
+      "commandPalette": [
+        {
+          "command": "extension.NAMESPACE(node-debug).prettyPrint",
+          "title": "%pretty.print.script%",
+          "when": "inDebugMode && debugType == NAMESPACE(chrome) && editorLangId == javascript"
```
roblourens: Why just `debugType == NAMESPACE(chrome)`?

src/test/sources/pretty-print.test.ts
```diff
@@ -0,0 +1,101 @@
...
+import { itIntegrates } from '../testIntegrationUtils';
+
+describe('pretty print sources', () => {
+  async function waitAndStayPaused(p: ITestHandle) {
+    const { threadId } = p.log(await p.dap.once('stopped'));
+    await p.logger.logStackTrace(threadId);
+    return () => p.dap.continue({ threadId });
+  }
+
+  itIntegrates.only('base', async ({ r }) => {
```
roblourens: `.only`s

package.json
connor4312: The child sessions use chrome as their 'fake' debug type in the attachment process. So that check will return true any time there's at least one debug target :)

https://github.com/microsoft/vscode-js-debug/blob/2f8893afbcd9b1f67f42cbaa20d7b96d7969b337/src/sessionManager.ts#L187-L192

src/test/sources/pretty-print.test.ts
connor4312: whops

PR #357, opened 2020-02-26
Fix wss

digeff:
Accept self-signed certificates on localhost on wss connections
cc @EricCornelson @connor4312

### DIFF
--- a/src/cdp/transport.ts
+++ b/src/cdp/transport.ts
@@ -8,6 +8,7 @@ import * as events from 'events';
 import { CancellationToken } from 'vscode';
 import { timeoutPromise } from '../common/cancellation';
 import { LogTag, ILogger } from '../common/logging';
+import { isLoopback } from '../common/urlUtils';
 
 export interface ITransport {
   send(message: string): void;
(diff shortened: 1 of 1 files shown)


src/cdp/transport.ts
```diff
@@ -108,10 +109,17 @@ export class WebSocketTransport implements ITransport {
   onmessage?: (message: string, receivedTime: bigint) => void;
   onend?: () => void;
 
-  static create(url: string, cancellationToken: CancellationToken): Promise<WebSocketTransport> {
+  static async create(
+    url: string,
+    cancellationToken: CancellationToken,
+  ): Promise<WebSocketTransport> {
+    const isSecure = !url.startsWith('ws://');
```
EricCornelson: any reason this couldn't be
```ts
const isSecure = url.startsWith('wss://')
```

src/cdp/transport.ts
```diff
@@ -108,10 +109,17 @@ export class WebSocketTransport implements ITransport {
...
+    const targetAddressIsLoopback = await isLoopback(url);
+
     const ws = new WebSocket(url, [], {
       perMessageDeflate: false,
       maxPayload: 256 * 1024 * 1024, // 256Mb
+      rejectUnauthorized: !(isSecure && targetAddressIsLoopback),
```
connor4312: Shouldn't this be

```suggestion
      rejectUnauthorized: isSecure && targetAddressIsLoopback,
```

?

src/cdp/transport.ts
connor4312: I prefer to treat things as secure unless explicitly specified otherwise. In the current use it doesn't matter that much, but if the code around it changes it might.

PR #359, opened 2020-02-26
Fix blazor bps

digeff:
When we set a breakpoint without sources being available, we also assume that it can be a file path instead of a file hosted on a web service

### DIFF
--- a/src/adapter/breakpoints/breakpointBase.ts
+++ b/src/adapter/breakpoints/breakpointBase.ts
@@ -8,7 +8,7 @@ import { Thread, Script } from '../threads';
 import Cdp from '../../cdp/api';
 import { LogTag } from '../../common/logging';
 import { IUiLocation, base1To0 } from '../sources';
-import { urlToRegex } from '../../common/urlUtils';
+import { urlToRegex, absolutePathToFileUrl } from '../../common/urlUtils';
 
 export type LineColumn = { lineNumber: number; columnNumber: number }; // 1-based
 
@@ -348,7 +348,7 @@ export abstract class Breakpoint {
       ? this._manager._sourceContainer.sourcePathResolver.absolutePathToUrl(this.source.path)
       : undefined;
     if (!url) return;
-    await this._setByUrl(thread, url, lineColumn);
+    await this._setByUrl(thread, url, lineColumn, this.source.path);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/adapter/breakpoints/breakpointBase.ts
```diff
@@ -371,14 +371,23 @@ export abstract class Breakpoint {
...
+  ): Promise<void> {
     lineColumn = base1To0(lineColumn);
     if (this.hasSetOnLocation({ url }, lineColumn)) {
       return;
     }
 
+    const filePathRegexp = filePath
+      ? '|' + urlToRegex(absolutePathToFileUrl(filePath)?.toString()!)
+      : '';
+    const urlRegexp = urlToRegex(url) + filePathRegexp;
```
connor4312: `urlToRegex` already does this if a file uri is presented:

https://github.com/microsoft/vscode-js-debug/blob/7818330c650465c4e62b4d61cd9d7ed40101b284/src/common/urlUtils.ts#L321-L325

connor4312:
How about taking this approach instead, and passing the filePath until urlToRegex there. Keeps the logic in one place and avoids adding duplicate path segments.

```diff
diff --git a/src/common/urlUtils.ts b/src/common/urlUtils.ts
index 0fb8cf0..bd6e8d3 100644
--- a/src/common/urlUtils.ts
+++ b/src/common/urlUtils.ts
@@ -315,15 +315,24 @@ export function isDataUri(uri: string): boolean {
 }
 
 /**
- * Converts and escape the file URL to a regular expression.
+ * Converts and escape the file URL to a regular expression. If the path
+ * is a file URI or an explicit `filePath` is set, the regex will match that
+ * disk path as well as the URI.
  */
-export function urlToRegex(aPath: string) {
+export function urlToRegex(aPath: string, filePath?: string) {
   const absolutePath = fileUrlToAbsolutePath(aPath);
   aPath = escapeRegexSpecialChars(aPath);
   if (absolutePath) {
     aPath += `|${escapeRegexSpecialChars(absolutePath)}`;
   }
 
+  if (
+    filePath &&
+    (!absolutePath || lowerCaseInsensitivePath(filePath) !== lowerCaseInsensitivePath(absolutePath))
+  ) {
+    aPath += `|${escapeRegexSpecialChars(filePath)}`;
+  }
+
   // If we should resolve paths in a case-sensitive way, we still need to set
   // the BP for either an upper or lowercased drive letter
   if (isCaseSensitive) {

```

connor4312:
Please add a test for this as well, if possible. Doesn't have to be in this PR.

PR #360, opened 2020-02-26
fix process names in vs

EricCornelson:
Debugged process names were blank in Visual Studio. Added fixes to make sure we send the initial target name, and subsequent name updates.


src/dap/connection.ts
```diff
@@ -28,6 +29,9 @@ export default class Connection {
   private _dap: Promise<Dap.Api>;
   private disposables: IDisposable[] = [];
 
+  private _initialized = getDeferred<Connection>();
```
digeff: Please give one of these a different name

PR #377, opened 2020-03-03
Use ripgrep for outFiles search

roblourens:
Since this requires bumping the vscode engine version, we will want to wait a bit before merging this, otherwise people using stable will not be able to use the extension. I guess we can wait a whole month or dynamically figure out whether Insiders is being used.


jasonwilliams:
stable is on 1.45.0 now, whats the update with this PR?

roblourens:
Thanks for the reminder. @connor4312 you can merge this if you are ready for it

PR #379, opened 2020-03-05
Return entry for stop on entry breakpoints

digeff:
Return entry for stop on entry breakpoints.
VS relies on the reason to auto-resume this breakpoint when needed

### DIFF
--- a/src/adapter/threads.ts
+++ b/src/adapter/threads.ts
@@ -71,6 +71,7 @@ export interface IThreadDelegate {
   scriptUrlToUrl(url: string): string;
   executionContextName(description: Cdp.Runtime.ExecutionContextDescription): string;
   initialize(): Promise<void>;
+  entryBreakpointId: string | undefined;
 }
 
 export type ScriptWithSourceMapHandler = (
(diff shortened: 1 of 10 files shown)


src/adapter/threads.ts
```diff
@@ -845,11 +846,14 @@ export class Thread implements IVariableStoreDelegate {
         };
       default:
         if (event.hitBreakpoints && event.hitBreakpoints.length) {
+          const isStopOnEntry =
+            event.hitBreakpoints.length === 1 &&
+            this._delegate.entryBreakpointId === event.hitBreakpoints[0];
```
connor4312:
```suggestion
          const isStopOnEntry =
           this._delegate.entryBreakpointId &&
           event.hitBreakpoints.includes(this._delegate.entryBreakpointId);
```

in case we stop there for some other random reason

src/targets/node/nodeAttacherBase.ts
```diff
@@ -39,6 +39,7 @@ export abstract class NodeAttacherBase<T extends AnyNodeConfiguration> extends N
       initialized: async () => {
         toDispose = this.onFirstInitialize(cdp, run);
         await toDispose;
+        return undefined;
```
connor4312: unrelated?

digeff: This is because I changed the signature of NodeTarget initialized method to:   
`initialized?(target: NodeTarget): Promise<string | undefined>;`

src/adapter/threads.ts
digeff: My thinking was that if the user sets a breakpoint on this same place, then we don't want VS to auto-resume, so we want the breakpoint to take precedence and say "breakpoint" as a reason instead of "entry"

src/targets/node/nodeAttacherBase.ts
digeff: This is because I changed the signature of NodeTarget initialized method to:   
`initialized?(target: NodeTarget): Promise<string | undefined>;`

src/adapter/threads.ts
connor4312: ah, right, makes sense

PR #380, opened 2020-03-05
adding Edge as an option to allow debugging edge with node

Shenniey:
Adding Edge to the check that enables debugging node with a browser

### DIFF
--- a/src/targets/node/nodeLauncher.ts
+++ b/src/targets/node/nodeLauncher.ts
@@ -66,11 +66,7 @@ export class NodeLauncher extends NodeLauncherBase<INodeLaunchConfiguration> {
     let config: INodeLaunchConfiguration | undefined;
     if (params.type === DebugType.Node && params.request === 'launch') {
       config = { ...params };
-    } else if (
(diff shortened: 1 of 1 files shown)


src/targets/node/nodeLauncher.ts
```diff
@@ -66,7 +66,11 @@ export class NodeLauncher extends NodeLauncherBase<INodeLaunchConfiguration> {
     let config: INodeLaunchConfiguration | undefined;
     if (params.type === DebugType.Node && params.request === 'launch') {
       config = { ...params };
-    } else if (params.type === DebugType.Chrome && params.server && 'program' in params.server) {
+    } else if (
+      (params.type === DebugType.Chrome || params.type === DebugType.Edge) &&
+      params.server &&
+      'program' in params.server
+    ) {
```
connor4312: We can actually use `in` and TS will narrow and type check for us 🙂 

```suggestion
    } else if ('server' in params && params.server && 'program' in params.server) {
```

Shenniey: Updated -- thank you!!

PR #389, opened 2020-03-10
fix: source handling with remoteRoot

### LINKED ISSUE #383: Node Debug not working via Docker Container
## Desc 
When using `pwa-node` locally it works fine, however using `pwa-node` via docker it fails to bound a breakpoint.

## Steps (Docker)
- Clone https://github.com/jasonwilliams/nextjs9-typescript-server-vscode-example
- npm install
- docker-compose up
- in VSCode "launch server (Docker)" debugger
- add a breakpoint in server/index.ts (line 12 const expressApp = express())
- it should say unbound breakpoint

## Steps local
- Clone https://github.com/jasonwilliams/nextjs9-typescript-server-vscode-example
- npm install
- docker-compose up
- in VSCode "launch server" debugger
- add a breakpoint in server/index.ts (line 12 const expressApp = express())
- it will work.

## Trace
https://gist.github.com/jasonwilliams/a185ee98ce15d9773a47dd688a867dc4
The mapping seems correct.

connor4312:
Fixes #383. There were a two things that caused issues attaching to
remote processes:

1. We requested sourcemaps from their remote location, which doesn't
   exist locally. This PR rebases the path correctly.
2. We used url length as the heuristic for which to narrow the 'correct'
   file at an absolute path. With a different remote root, though, the
   urls might be totally different. We now also check whether the
   original path is the sourcemap parent of the generated path. This was
   the problem in the linked issue, where ts-node was used.

### DIFF
--- a/src/test/breakpoints/breakpointsTest.ts
+++ b/src/test/breakpoints/breakpointsTest.ts
@@ -621,6 +621,21 @@ describe('breakpoints', () => {
       handle.assertLog({ substring: true });
     });
 
+    itIntegrates.only('works in remote workspaces', async ({ r }) => {
+      await r.initialize;
+
+      const cwd = join(testWorkspace, 'tsNode');
+      const handle = await r.runScriptAsRemote(join(cwd, 'index.js'));
+      await handle.dap.setBreakpoints({
+        source: { path: join(cwd, 'double.ts') },
+        breakpoints: [{ line: 5, column: 1 }],
+      });
+
+      handle.load();
+      await waitForPause(handle);
+      handle.assertLog({ substring: true });
+    });
+
     itIntegrates('does not adjust already correct', async ({ r }) => {
       await r.initialize;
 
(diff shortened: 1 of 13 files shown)


jasonwilliams:
Should there be tests for this?
Btw thanks soo much for fixing it!

connor4312:
Yea, I was thinking about how to do this, I think I can do it well enough by having a test with a duplicate of the workspace running the process out of `os.tmpdir()`. Won't be able to test cross-platform things (e.g. Windows running Linux containers) but that would work for this, at least.

jasonwilliams:
I think that should be good enough

connor4312:
Yep, good reasoning.

I unintentionally practiced proper TDD, by adding a test and not realizing I had changed back to `master`, then spending a while debugging why it was failing.

src/test/breakpoints/breakpointsTest.ts
```diff
@@ -621,6 +621,21 @@ describe('breakpoints', () => {
       handle.assertLog({ substring: true });
     });
 
+    itIntegrates.only('works in remote workspaces', async ({ r }) => {
```
jasonwilliams: I’m guessing you don’t want to keep the `.only`

jasonwilliams:
Seems to be working great today :D
Thank you

connor4312:
Awesome thanks for confirming!

Issue #345, opened 2020-02-21
[Feature request] Add inspectUri argument to Node debugger


SounD120:
Hey guys, [React Native Tools](https://github.com/microsoft/vscode-react-native) team here. During migration of the extension on js-debug we found out that Node debugging scenarios doesn't support [`inspectUri`](https://github.com/microsoft/vscode-js-debug/blob/09bffbb12174fbd85d95ab9f30eeef49e2457f9b/src/build/strings.ts#L77) argument for discovering debugging target websocket via [debugging proxy](https://github.com/microsoft/vscode-cdp-proxy) as browser scenarios does. Is it possible to add such an argument for Node scenarios?

connor4312:
Node debugging hooks up differently than Chrome, so it's not a super trivial change--there's no existing pathway where we go out and make a websocket connection to a debug target in Node. Are you trying to do something that doesn't work with the inspectUri shim present in the Chrome configs?

SounD120:
@connor4312 In React Native extension there is a special debugger worker process that communicates with the application JS bridge. The whole debugging workflow is tied to this [pure Node process](https://github.com/microsoft/vscode-react-native/blob/0379522089f47369dd1c3c934f3bf705d82b4bb1/src/debugger/forkedAppWorker.ts#L76-L128). And since our debugging case in React Native extension is about debugging Node, I think Chrome debugging scenarios will not work for it.

RedMickey:
@connor4312, we tried to [add a support of `inspectUri` parameter to Node Attach scenario](https://github.com/RedMickey/vscode-js-debug/pull/1). `inspectUri` parameter was added to Node Attach scenario configuration and handled almost the same way as for Chrome cases. The parameter was passed to `watchDog` process for establishing a connection with CDP proxy over Node debugger. With this configuration we can debug React Native applications, but it turned out that for some reasons breakpoints processing time increased. Now it takes a second or even longer to handle a breakpoint. Is this an expected behavior? Do you know what could cause this?

Also, we saw an update in [vscode-cdp-proxy](https://github.com/microsoft/vscode-cdp-proxy), and it seems to be very useful for us. Could you please publish a new version of that package so we could add it to the extension as a dependency? Thanks.

roblourens:
Have you tried this with a "chrome" type config using inspectUri? Even though you are actually attaching to a Node process, since we are just attaching to some websocket using CDP, it shouldn't matter.

[...]

PR #391, opened 2020-03-11
Add inspectUri parameter to Node target

RedMickey:
#345 
This is a probable implementation of support of `inspectUri` parameter for Node attach scenario.
It can be used the following way:
```
{
    "type": "pwa-node",
    "request": "attach",
    "name": "Attach",
    "port": 40324,
    "inspectUri": "ws://127.0.0.1:13602?browser={browserInspectUri}"
}
```


RedMickey:
According to this [issue](https://github.com/microsoft/vscode-js-debug/issues/345), it seems that implementation of `inspectUri` parameter won't make much sense for the extension in this case. So we are going to implement inspector socket lookup on RN extension side ourselves.

PR #408, opened 2020-03-30
fix: adjust line breakpoints to the first character in their line

### LINKED ISSUE #407: The breakpoint appears to wrong line
This is my test project, you can git clone and test:
[https://github.com/njzydark/vscode-debug-test](https://github.com/njzydark/vscode-debug-test)

When use chrome config to debug, every thing works right, bug when switch to pwa-chrome, there has two problems:

1. lanuch is slow
2. when I set breakpoint, it appear to wrong line

wrong-line

connor4312:
This fixes #407. TypeScript is generally fine, but in Babel sourcemaps
they don't include the leading whitespace as part of the first
statement. This can cause us to place the breakpoint on the start of the
previous line, and in _some_ cases the "variance" calculation will
result in the breakpoint being placed on the next statement, since it
sees the original line had a >0 variance.

In this change, we attempt to scan the source files and automatically
move line breakpoints to the first non-whitespace character in the line
they appear on (if any). This causes the location to be resolved
correctly in all cases.


src/adapter/breakpoints.ts
```diff
@@ -367,6 +374,9 @@ export class BreakpointManager {
 
       for (let index = 0; index < params.breakpoints.length; index++) {
         const bpParams = params.breakpoints[index];
+        if (!bpParams.column) {
+          bpParams.column = columnChars[bpParams.line];
```
roblourens: So you get a breakpoint request with no column or column 0, and try to get the sourcemapped position, and the problem is that babel's sourcemap doesn't have an exact match for column 0, so you get a position in the previous line or the next position in the line? Is that right? Would it be the same as saying "when bpParams.column isn't set, always take the next sourcemap position on the same source line" or something?

connor4312: Yes, exactly. Here's a visualization of the mapping in Sokra's source-map-visualization. Previously line breakpoints were getting set at column 1, which was before the source range in app.tsx, so with GREATEST_LOWER_BOUND, it was placing the breakpoint at the opening brace of the function.



Looking at it some more, I think we can actually avoid the file reading by using `allGeneratedPositionsFor` and then using the first-lowest-variance location from the candidates. I'll give that a spin...

roblourens: Cool, and I don't know whether js-debug considers the source line of the mapped position at all when evaluating breakpoint locations, v1 never did, but I imagined that we could do something like

- Get setBreakpoint request with source location
- Pick a mapping, get the generated location from sourcemap
- Run that location back through to get the source location
- If it doesn't map back to the same line as the requested source location, score that against it

connor4312: Yep, you actually suggested doing that a little while ago, that's what we do :) I pushed a followup commit that adjusts our logic there. Previously we only looked at a single location in the `LEAST_UPPER_BOUND`, but as you can see in the screenshot babel mapped `console` to multiple locations. It happened that the position at the semicolon was the first-generated position in the map, so we would set a breakpoint there which Chrome would resolve to the next line.

The update I pushed gets the first location with the least variance, that maps correctly.

roblourens: Oh ok cool!

connor4312:
OSX tests have an existing issue since the build agents were updated

PR #439, opened 2020-04-17
Split telemetry into multiple properties to avoid it being cut and dropped

digeff:
We are currently dropping a lot of telemetry because we are hitting the property size limits. Splitting the operations into multiple properties will avoid this issue.

Telemetry properties with this change look like:
vs.bpt.diagnostics.debugadapter.setbreakpoints	{
  "operation": "setBreakpoints",
  "totalTime": 1.4,
  "max": 1.4,
  "avg": 1.4,
  "stddev": null,
  "count": 1,
  "failed": 0
}
vs.bpt.diagnostics.debugadapter.setbreakpoints.errors	[]
vs.bpt.diagnostics.debugadapter.setexceptionbreakpoints	{
  "operation": "setExceptionBreakpoints",
  "totalTime": 0.7,
  "max": 0.7,
  "avg": 0.7,
  "stddev": null,
  "count": 1,
  "failed": 0
}
vs.bpt.diagnostics.debugadapter.target.attachedtotarget	{
  "operation": "Target.attachedToTarget",
  "totalTime": 0.7,
  "max": 0.6,
  "avg": 0.3,
  "stddev": 0.3,
  "count": 2,
  "failed": 0
}
vs.bpt.diagnostics.debugadapter.target.attachedtotarget.errors	[]

@connor4312 @roblourens @EricCornelson


src/telemetry/classification.ts
```diff
@@ -55,7 +55,10 @@ export interface IGlobalMetrics {
 }
 
 interface IRPCOperationClassification {
-  performance: { classification: 'SystemMetaData'; purpose: 'PerformanceAndHealth' };
+  [key: string]: {
```
connor4312, line 58: The GDPR classification tool doesn't support generic keys unfortunately--that's why I'd moved it to an array initially

scripts/generate-lib.js
```diff
@@ -0,0 +1,34 @@
...
+  const fileName = path.join(__dirname, relativeFilePath);
+  fs.writeFileSync(
+    fileName,
+    code,
+    prettier.format(result.join('\n'), {
+      parser: 'typescript',
+      ...require('../package.json').prettier,
+    }),
+    { encoding: 'utf-8' },
+  );
```
connor4312, line 19: You're passing the "code" as the second argument and the prettier result as the "options", which won't do anything--surprised it doesn't throw an error

```suggestion
  fs.writeFileSync(
    fileName,
    prettier.format(result.join('\n'), {
      parser: 'typescript',
      ...require('../package.json').prettier,
    }),
    { encoding: 'utf-8' },
  );
```

src/telemetry/classification.ts
connor4312: We should be able to delete the arbitrary string key here so that everything gets fully typed.

PR #444, opened 2020-04-22
Avoid duplicated this variables in Blazor

digeff:
Avoid duplicated this variables in Blazor
This will fix: https://github.com/mono/mono/issues/19306

### DIFF
--- a/src/adapter/variables.ts
+++ b/src/adapter/variables.ts
@@ -154,14 +154,20 @@ export class VariableStore {
 
     const variables = await this._getObjectProperties(object);
     if (object.scopeRef) {
+      const existingVariables = new Set(variables.map(v => v.name));
       for (const extraProperty of object.extraProperties || [])
-        variables.push(
-          this._createVariable(
-            extraProperty.name,
-            object.wrap(extraProperty.name, extraProperty.value),
-            'propertyValue',
-          ),
-        );
+        /* Normally we add the "this" variable as en extra propertie because it's not included in the variables
+         * that come from v8. Blazor does include it, and we don't know what other CDP debuggers will do, so we
+         * avoid adding duplicated variables.
+         */
+        if (!existingVariables.has(extraProperty.name))
+          variables.push(
+            this._createVariable(
+              extraProperty.name,
+              object.wrap(extraProperty.name, extraProperty.value),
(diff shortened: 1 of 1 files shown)


connor4312:
Since this is a behavior difference in Blazor, should the tweak be applied on Blazor's end rather than the adapter?

digeff:
Chrome DevTools shows a single "this" variable for Blazor, so they have some logic to figure this case out. I think we should try to match it, and implement our own logic to make this work.

connor4312:
I'm OK with that then. Thanks for the fix.

dgozman:
> Chrome DevTools shows a single "this" variable for Blazor, so they have some logic to figure this case out. I think we should try to match it, and implement our own logic to make this work.

DevTools has the similar logic, because of source maps "names" section. I assume it also accidentally works for Blazor.

lewing:
Thanks for taking care of this on your end for now, we would love feedback on the expectations we are violating from your perspective. We're in the uncomfortable position of trying to represent a very different execution environment and accidentally working in DevTools is much better than not working at all from our perspective.

connor4312:
From my POV -- there's two parts of CDP

 - the protocol documentation (i.e. chromedevtools.github.io/devtools-protocol)
 - the behavior that Chrome exhibits

The protocol documentation almost exclusively describes the structure of the protocol, and alone is not enough to implement CDP. So therefore I view what Chrome does as the normative behavior for CDP, which defines js-debug. For translated protocols, like Blazor and (to a lesser extent) React Native, it's my expectation that the CDP proxy will translate in such a fashion that the debug target behaves identically to Chrome for the subset of functionality that we require.

[](https://github.com/microsoft/vscode-cdp-proxy)

This separation keeps js-debug from gathering a charcuterie of patches and workaround for various targets, which increase complexity and are hard for us to test--and therefore maintain. Also, because there is not a single, IETF-style spec for CDP, patches can become especially 'sticky' because there is no single source that can readily indicate whether a behavior is part of the 'CDP spec', some deprecated functionality, or a patch for a certain debug target. Stickiness can be partially mitigated through comments, as Diego did in this PR, but comments have an tendency to become outdated, irrelevant, or missing over years.

Hope that makes sense.

PR #446, opened 2020-04-22
Send commit id in telemetry

digeff:
For VS we'll send the exact commit id in the telemetry messages


src/telemetry/classification.ts
```diff
@@ -205,6 +209,7 @@ export const createLoggers = (sendEvent: (event: Dap.OutputEventParams) => void)
   const browserVersion = (metrics: IBrowserVersionMetrics) => {
     globalMetrics.browser =
       (metrics.targetProject || metrics.targetProject) + '/' + metrics.targetVersion;
+    setJsDebugCommitId(globalMetrics);
```
connor4312, line 212: I would just run this code synchronously when `createLogger()` is first called

PR #458, opened 2020-05-06
fix: major performance issues with webpack eval bundles

### LINKED ISSUE #431: Slow starting/refreshing of debug session in Chrome
When I am using debug.chrome.useV3: true then debug session took several minutes to just start Angular based application. Any update of code means reload of application which means another wait session. So I am forced to not use V3 for debugging even to try it. I noticed that break points also is not bound so for now this new debugger seems completely usefulness for browser debugging. If you have any suggestions I could try. Official version of debugger works normally for me.

connor4312:
Webpack has a couple modes where each module in the source file is
emitted in its own `eval` statement. As these modes are generally the
fastest for webpack to produce, they're fairly popular.

The issue is that js-debug pauses on each sourcemapped script it
encounters in order to parse their sourcemaps and set and breakpoints.
Usually on the web, we get one sourcemap per JavaScript file, which we
can parse pretty easily. But in eval mode we get thousands of indiviual,
tiny scripts and sourcemaps that we pause, parse, and resume in series.
Even for a fairly simple app of mine, load times were around 20-30s.

My workaround here is to add a change that detects if the code we're
looking at seems to be from a webpack eval bundle (if we see a data-uri
sourcemap, coming from a script in an eval statement, whose sourceURL is
set and starts with "webpack"). In this case we drop into a slightly
less-correct mode: the `beforeScriptWithSourceMapExecution` breakpoint
is removed and instead we set aggressive entrypoint breakpoints on the
basename of any file the user has breakpoints in. For example, if they
have set a breakpoint in `foo/bar.ts`, we'll set a breakpoint to pause
at the entrypoint of any script containing `bar`.

The downside is that this doesn't take into account any custom source
map overrides, so they could map `webpack://./potato.ts` to `foo/bar.ts`
and we would not pick up on it.

Alternative solutions I thought about were either (a) not supporting it
or (b) doing some magical override of the `eval` statement on the page.
However, I think this is a better approach and is actually a pretty
minimal amount of code.

(You might ask "what does Chrome devtools do here?" The answer is:
nothing. They don't seem to have special logic to ensure we set
breakpoints before evaluating code, they just work as fast as they can
and hope the breakpoints get set in time.)

Fixes #431

### DIFF
--- a/src/adapter/breakpoints/entryBreakpoint.ts
+++ b/src/adapter/breakpoints/entryBreakpoint.ts
@@ -2,15 +2,40 @@
  * Copyright (C) Microsoft Corporation. All rights reserved.
  *--------------------------------------------------------*/
 
-import { Breakpoint } from './breakpointBase';
-import { BreakpointManager } from '../breakpoints';
+import { Breakpoint, LineColumn } from './breakpointBase';
+import { BreakpointManager, EntryBreakpointMode } from '../breakpoints';
 import Dap from '../../dap/api';
+import { basename, extname } from 'path';
+import { Thread } from '../threads';
 
 /**
  * A breakpoint set automatically on module entry.
  */
 export class EntryBreakpoint extends Breakpoint {
-  constructor(manager: BreakpointManager, source: Dap.Source) {
+  public static getModeKeyForSource(mode: EntryBreakpointMode, path: string) {
+    return mode === EntryBreakpointMode.Greedy ? basename(path, extname(path) || undefined) : path;
+  }
+
+  constructor(
+    manager: BreakpointManager,
+    source: Dap.Source,
+    private readonly mode: EntryBreakpointMode,
+  ) {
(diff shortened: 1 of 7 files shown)


src/adapter/breakpoints/entryBreakpoint.ts
```diff
@@ -2,15 +2,40 @@
...
+  constructor(
+    manager: BreakpointManager,
+    source: Dap.Source,
+    private readonly mode: EntryBreakpointMode,
+  ) {
     super(manager, source, { lineNumber: 1, columnNumber: 1 });
   }
+
+  protected _setPredicted() {
+    return Promise.resolve();
```
connor4312, line 28: Since I was here, I stubbed this out -- there's no point to looking at sourcemaps while setting entry BPs, since these BPs are exclusively used for intercepting scripts that don't yet have sourcemaps.

PR #459, opened 2020-05-06
Added posibility to run with chosen package manager instead of `npm` only.

cbmd:
Was fully extracted from bundled `npm` extension.

Ideally it should consume function exposed by by `npm` extension bundled within VS Code, although function exposed from `tasks.ts` but not from whole extension and can't be consumed by `vscode.extensions.getExtension('vscode.npm')` notation.


connor4312:
Looks good, thanks! Test flake is not related.

PR #473, opened 2020-05-15
Fix hitting breakpoints on first line on VS

digeff:
* Make entryBreakpoint id unique by adding (?:entryBreakpoint){0} to it, so CDP will let us add 2 breakpoints in the same location
* Remove our logic that prevents adding an entry breakpoint if there is an user breakpoint in the same location
* When we hit the entryBreakpoint, if there is a user breakpoint in the same location, consider it a normal breakpoint. If there isn't, consider it an entry breakpoint (VS needs this to process the stop on entry behavior properly)


src/targets/targets.ts
```diff
@@ -48,7 +48,12 @@ export interface ITarget {
   scriptUrlToUrl(url: string): string;
   sourcePathResolver(): ISourcePathResolver;
   executionContextName(context: Cdp.Runtime.ExecutionContextDescription): string;
-  entryBreakpointId: string | undefined;
+  entryBreakpoint: IBreakpointPathAndId | undefined;
+}
+
+export interface IBreakpointPathAndId {
+  path: string;
+  cdtpId: string;
```
connor4312, line 56: Typo?

```suggestion
  cdpId: string;
```

src/adapter/threads.ts
```diff
@@ -885,9 +887,32 @@ export class Thread implements IVariableStoreDelegate {
...
+            });
+
+            if (entryBreakpointSource !== undefined) {
+              const entryBreakpointLocations = this._sourceContainer.currentSiblingUiLocations({
+                lineNumber: 1,
+                columnNumber: 1,
+                source: entryBreakpointSource,
+              });
+
+              // But if there is a user breakpoint on the same location that the stop on entry breakpoint, then we consider it an user breakpoint
```
connor4312, line 910: If there is a user breakpoint at the same spot, will `event.hitBreakpoints` contain all of them?

src/adapter/breakpoints.ts
```diff
@@ -629,12 +638,6 @@ export class BreakpointManager {
       return;
     }
 
-    // Don't apply a custom breakpoint here if the user already has one.
-    const byPath = this._byPath.get(source.path) ?? [];
-    if (byPath.some(isSetAtEntry)) {
-      return;
-    }
```
connor4312, line 636: Why did we need to remove this?

src/targets/targets.ts
digeff: I'll do the rename on the workspace

src/adapter/breakpoints.ts
digeff: Sorry, this is not needed any more now that I'm using a unique id for the entry stop breakpoint. Restoring

src/adapter/threads.ts
digeff: In the cases I tested it just contains the entry breakpoint. I wrote the code as-is in case we run into a case where we find more than 1 breakpoint here.

src/adapter/breakpoints.ts
```diff
@@ -94,6 +94,15 @@ export class BreakpointManager {
     urlUtils.lowerCaseInsensitivePath,
   );
 
+  public hasAtLocation(location: IUiLocation) {
+    const breakpointsAtPath = this._byPath.get(location.source.absolutePath()) || [];
```
connor4312, line 98: I would suggest checking _byRef on the source as well. That code path isn't used in these changes, but that would avoid a footgun if someone tries to use this later and wonders why it isn't working. Something like

```ts
return [this._byRef.get(location.source.sourceReference(), this._byPath.get(location.source.absolutePath)]
    .some(breakpoints => breakpoints?.some(bp => ...
```

connor4312:
npm run fmt and you're set

connor4312:
🚀

PR #493, opened 2020-06-01
Fix possible race condition

digeff:
In telemetry we got:

TypeError: Invalid value used as weak map key
    at WeakMap.set (<anonymous>)
    at k._onPaused (threads.js:511:34)

My guess is that while we await this._smartStepper.shouldSmartStep(pausedDetails)
For some reason the app resumes (maybe the user reloads the page, or uses Chrome DevTools), so this._pausedDetails becomes undefined and this._pausedDetailsEvent.set(pausedDetails, event); fails because undefined can't be a key of a WeakMap.

Assuming that's happening, this fix should address that issue

### DIFF
--- a/src/adapter/threads.ts
+++ b/src/adapter/threads.ts
@@ -666,24 +666,25 @@ export class Thread implements IVariableStoreDelegate {
       return;
     }
 
-    if (await this._smartStepper.shouldSmartStep(pausedDetails)) {
-      if (this._pausedDetails === pausedDetails) {
-        this.stepInto();
-      }
+    const shouldSmartStep = await this._smartStepper.shouldSmartStep(pausedDetails);
+    if (this._pausedDetails !== pausedDetails) {
       return;
     }
 
-    if (this._pausedDetails === pausedDetails) {
-      this._pausedDetailsEvent.set(pausedDetails, event);
-      this._pausedVariables = new VariableStore(
-        this._cdp,
-        this,
-        this.launchConfig.__autoExpandGetters,
-      );
-      scheduledPauseOnAsyncCall = undefined;
-
-      await this._onThreadPaused(pausedDetails);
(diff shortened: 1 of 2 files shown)


src/adapter/threads.ts
```diff
@@ -659,25 +659,26 @@ export class Thread implements IVariableStoreDelegate {
...
 
-    this._pausedDetails = this._createPausedDetails(event);
+    // We store pausedDetails in a local variable to avoid race conditions while awaiting this._smartStepper.shouldSmartStep
+    const pausedDetails = (this._pausedDetails = this._createPausedDetails(event));
     if (!shouldPause) {
       this.resume();
       return;
     }
 
-    if (await this._smartStepper.shouldSmartStep(this._pausedDetails)) {
```
connor4312, line 668: You could double check, after awaiting, that `this._pausedDetails === pausedDetails`, so that we don't smart step or trigger onThreadPaused if a race occurs here. (if they changed, just bail out)

digeff: Changed

src/adapter/threads.ts
```diff
@@ -659,25 +659,31 @@ export class Thread implements IVariableStoreDelegate {
...
+      this._pausedDetailsEvent.set(pausedDetails, event);
+      this._pausedVariables = new VariableStore(
+        this._cdp,
+        this,
+        this.launchConfig.__autoExpandGetters,
+      );
+      scheduledPauseOnAsyncCall = undefined;
+
+      await this._onThreadPaused(pausedDetails);
+    }
```
connor4312, line 686: Something like

```js
const shouldSmartStep = await this._smartStepper.shouldSmartStep(pausedDetails);
if (this._pausedDetails !== pausedDetails) {
  return;
}

...
```

would be less repetitive :)

PR #501, opened 2020-06-03
fix: Blazor attach (add support for inspectUri in attach)

EricCornelson:
Adding support for inspectUri (for Blazor) in attach.

### DIFF
--- a/src/targets/browser/launcher.ts
+++ b/src/targets/browser/launcher.ts
@@ -194,7 +194,7 @@ export async function attach(
     const connectionURL = await retryGetWSEndpoint(browserURL, cancellationToken);
 
     const inspectWs =
-      options.inspectUri && options.pageURL
+      options.inspectUri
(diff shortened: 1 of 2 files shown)


src/targets/browser/launcher.ts
```diff
@@ -189,7 +192,13 @@ export async function attach(
     return new CdpConnection(connectionTransport, logger, telemetryReporter);
   } else if (browserURL) {
     const connectionURL = await retryGetWSEndpoint(browserURL, cancellationToken);
-    const connectionTransport = await WebSocketTransport.create(connectionURL, cancellationToken);
+
+    const inspectWs =
+      options.inspectUri && options.pageURL
```
connor4312, line 197: Technically we don't require a page URL in the general sense to use constructInspectorWSUri

```suggestion
      options.inspectUri
```

Issue #486, opened 2020-05-26
Error: Unknown session id: <session-id> while processing: Inspector.targetReloadedAfterCrash


digeff:
We got telemetry that the debug adapter is seeing a lot of this error:
```
Error: Unknown session id: <session-id> while processing: Inspector.targetReloadedAfterCrash
    at t.default._onMessage (connection.js:47:19)
    at call (connection.js:19:61)
    at t.EventEmitter.fire (events.js:36:27)
    at T.call (webSocketTransport.js:21:33)
    at T.r (event-target.js:120:16)
    at T.emit (events.js:210:5)
    at e.exports.R (websocket.js:801:20)
    at e.exports.emit (events.js:210:5)
    at e.exports.dataMessage (receiver.js:436:14)
    at e.exports.getData (receiver.js:366:17)
```

This also happens while processing these other events, and some request responses too:

- Debugger.scriptParsed
- Mono.runtimeReady
- Runtime.consoleAPICalled
- Inspector.targetCrashed
- Inspector.detached

Can this be causing a detrimental experience to the users?

digeff:
A lot of our users are hitting this issue. Raising to vs-p1

connor4312:
Found one big case I was hitting locally -- if the user manually detached from a target, we didn't tell Chrome about that, so were still getting messages from a tab that we didn't care about.

digeff:
Awesome! :)

connor4312:
@digeff / @EricCornelson do you still see this coming up in your telemetry since my initial fix above?

connor4312:
Closing and marking verified -- please let me know if this is not appropriate.

digeff:
@connor4312 Sorry for the delay. At the moment all the stack traces in telemetry are blank, so we don't know what's happening yet. We are looking into this.

PR #511, opened 2020-06-05
Partial fix for unknown session

digeff:
I was able to reproduce the unknown session issue #486 . We were getting messages about a session after it has been disposed.
I don't know if this solves all cases of unknown session or not.

### DIFF
--- a/src/cdp/connection.ts
+++ b/src/cdp/connection.ts
@@ -48,6 +48,7 @@ export default class Connection {
   private _lastId: number;
   private _transport: ITransport;
   private _sessions: Map<string, CDPSession>;
+  private _disposedSessions = new Map<string, Date>();
   private _closed: boolean;
   private _rootSession: CDPSession;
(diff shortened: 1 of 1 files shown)


src/cdp/connection.ts
```diff
@@ -98,7 +99,20 @@ export default class Connection {
...
     if (!session) {
-      throw new Error(`Unknown session id: ${object.sessionId} while processing: ${object.method}`);
+      const disposedDate = this._disposedSessions.get(object.sessionId);
+      if (!disposedDate) {
+        throw new Error(
+          `Unknown session id: ${object.sessionId} while processing: ${object.method}`,
+        );
+      } else {
+        const secondsAgo = (Date.now() - disposedDate.getTime()) / 1000.0;
+        this.logger.verbose(
```
connor4312, line 109:
```suggestion
        this.logger.warn(
```

PR #517, opened 2020-06-09
Kill only the tab by default

digeff:
Now we only kill the tab when the debugging sessions ends by default.
I added an option to keep the existing functionality too in case someone prefers that one.

### DIFF
--- a/src/build/generate-contributions.ts
+++ b/src/build/generate-contributions.ts
@@ -793,8 +793,9 @@ const chromeLaunchConfig: IDebugger<IChromeLaunchConfiguration> = {
     },
     cleanUp: {
       type: 'string',
+      enum: ['wholeBrowser', 'onlyTab'],
       description: refString('browser.cleanUp.description'),
-      default: null,
+      default: 'onlyTab',
     },
     browserLaunchLocation: {
       description: refString('browser.browserLaunchLocation.description'),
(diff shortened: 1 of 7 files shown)


src/configuration.ts
```diff
@@ -503,6 +503,11 @@ export interface IChromiumLaunchConfiguration extends IChromiumBaseConfiguration
    * If true, will start profiling soon as the page launches.
    */
   profileStartup: boolean;
+
+  /**
+   * Close whole browser or just the tab when cleaning up
+   */
+  cleanUp: 'wholeBrowser' | 'onlyTab' | null;
```
connor4312, line 510: I don't think null is a valid value here

src/build/generate-contributions.ts
```diff
@@ -791,6 +791,11 @@ const chromeLaunchConfig: IDebugger<IChromeLaunchConfiguration> = {
       description: refString('browser.profileStartup.description'),
       default: true,
     },
+    cleanUp: {
+      type: 'string',
+      description: refString('browser.cleanUp.description'),
+      default: null,
+    },
```
connor4312, line 798: Should have an `enum` attribute

src/configuration.ts
digeff: Removed

src/build/generate-contributions.ts
digeff: Done

connor4312:
Looks like the new default of killing only the tab is causing tests to time out

PR #531, opened 2020-06-16
fix: cdp message perf issue

EricCornelson:
The current method we use to ensure cdp handler task ordering is
extremely slow (e.g. limited to ~65 messages/sec) and can cause major
issues when debugging certain projects.

We now use the same fix for this as playwright which is much more
performant, and also corrected the check for node version, since this is
only a problem in < 11.0.0 so we don't use the fix on stable 12.18.0
(which doesn't need it)

### DIFF
--- a/src/cdp/connection.ts
+++ b/src/cdp/connection.ts
@@ -174,9 +174,9 @@ export default class Connection {
   }
 }
 
-// Node versions before 12 do not guarantee relative order of tasks and microstasks.
+// Node versions before 11.0.0 do not guarantee relative order of tasks and microstasks.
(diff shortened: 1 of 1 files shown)


src/cdp/connection.ts
```diff
@@ -309,7 +309,8 @@ class CDPSession {
   }
 
   _processQueue() {
-    setTimeout(() => {
+    const waitWrapper = makeWaitForNextTask();
```
dgozman, line 312: I think it is important to call makeWaitForNextTask() once, and use the same wrapper all the time. Otherwise, `spinning` flag will be a new variable all the time.

PR #532, opened 2020-06-17
Improve user experience when content validation fails

digeff:
Show an error message when content validation fails. Return undefined so other code that depends on this can work as expected.

This is in response of having several reports of this error from telemetry:
```
Error [ERR_IPC_CHANNEL_CLOSED]: Channel closed
    at ChildProcess.send (internal/child_process.js:678:16)
    at send (index.js:37:8)
    at Object.verifyFile (index.js:56:66)
    at Object.checkContentHash (sourceUtils.js:253:24)
    at new A (sources.js:92:50)
    at O.source [as addSource] (sources.js:511:24)
    at threads.js:912:19
```

### DIFF
--- a/src/adapter/sources.ts
+++ b/src/adapter/sources.ts
@@ -27,7 +27,7 @@ import { Script } from './threads';
 import { IScriptSkipper } from './scriptSkipper/scriptSkipper';
 import { once } from '../common/objUtils';
 import { IResourceProvider } from './resourceProvider';
-import { sourceMapParseFailed } from '../dap/errors';
+import { sourceMapParseFailed, contentHashValidationFailed } from '../dap/errors';
 
 const localize = nls.loadMessageBundle();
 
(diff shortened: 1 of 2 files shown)


connor4312:
Sorry for the late review on this.

I think this could happen if the [forked hash process](https://github.com/microsoft/vscode-js-debug/blob/master/src/common/hash/index.ts#L25-L32) is closed or killed. I'm not sure what would cause this, maybe an OOM killer? I think the better solution is to have an `exit` listener on the process to reset the `instance = undefined`. Maybe that alone will solve the error.

I hesitate showing a user message here since it's not clear what they need to do to fix the problem.

digeff:
@connor4312 If I understand correctly your suggestion, here it is: https://github.com/microsoft/vscode-js-debug/pull/539
Do you know how can I test that code?

connor4312:
@digeff you should be able to test it by:

 1. opening a webpage/node program, the killing the node.exe process running the hash function
 2. _within 30 seconds_ reload the page or do something else that causes more scripts to load in an need hashing
 3. if the change worked, the hasher should be restarted automatically without errors

connor4312:
Thanks for the PR! #539 should fix the issue, so I'll close the PR (but keep the branch in case we end up needing this)

PR #533, opened 2020-06-18
chore: make waitWrapper a single constant

EricCornelson:
Addressing @dgozman's [feedback](https://github.com/microsoft/vscode-js-debug/pull/531#discussion_r441134020) about this so that we don't create separate
state for each callback.

### DIFF
--- a/src/cdp/connection.ts
+++ b/src/cdp/connection.ts
@@ -311,7 +311,6 @@ class CDPSession {
   }
 
   _processQueue() {
-    const waitWrapper = makeWaitForNextTask();
     waitWrapper(() => {
       if (this.paused) {
         return;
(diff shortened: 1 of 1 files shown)


src/cdp/connection.ts
```diff
@@ -377,6 +376,8 @@ class CDPSession {
   }
 }
 
+const waitWrapper = makeWaitForNextTask();
```
connor4312, line 379: This can be stored as a property on the Connection instance I believe. Avoids extra waiting in case we have multiple active CDP connections (e.g. multiple debug sessions going)

EricCornelson: 👍

PR #539, opened 2020-06-24
Handler hasher process exit

digeff:
Feedback from: https://github.com/microsoft/vscode-js-debug/pull/532

### DIFF
--- a/src/common/hash/index.ts
+++ b/src/common/hash/index.ts
@@ -27,7 +27,6 @@ const create = () => {
     return instance;
   }
 
-  instanceFailureCount = 0;
   instance = fork(join(__dirname, 'hash.bundle.js'), [], { env: {}, silent: true, execArgv: [] });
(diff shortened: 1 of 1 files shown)


src/common/hash/index.ts
```diff
@@ -30,6 +30,14 @@ const create = () => {
     delete deferredMap[msg.id];
     deferred?.resolve(msg);
   });
+  instance.on('exit', () => {
+    Object.keys(deferredMap).forEach(msgId => {
+      delete deferredMap[msgId];
+      const deferred = deferredMap[msgId];
+      deferred?.reject(new Error('hash.bundle.js process unexpectedly exited'));
+    });
```
connor4312, line 38: Alternately, we could actually issue those deferred requests again to the new process to avoid any kind of interruption

digeff: Added some retrying with a max retry count

PR #547, opened 2020-07-01
Do not block on exception pause before configurationDone

digeff:
Chrome cannot answer some APIs when the target is paused. Do not block on exception pause before configurationDone to avoid the API getting stuck.

### DIFF
--- a/src/adapter/debugAdapter.ts
+++ b/src/adapter/debugAdapter.ts
@@ -173,7 +173,37 @@ export class DebugAdapter implements IDisposable {
     this._pauseOnExceptionsState = 'none';
     if (params.filters.includes('caught')) this._pauseOnExceptionsState = 'all';
     else if (params.filters.includes('uncaught')) this._pauseOnExceptionsState = 'uncaught';
-    if (this._thread) await this._thread.setPauseOnExceptionsState(this._pauseOnExceptionsState);
+    if (this._thread) {
+      const result = this._thread.setPauseOnExceptionsState(this._pauseOnExceptionsState);
+      const logger = this._services.get<ILogger>(ILogger);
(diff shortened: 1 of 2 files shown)


src/adapter/debugAdapter.ts
```diff
@@ -173,7 +173,37 @@ export class DebugAdapter implements IDisposable {
...
+          })
+          .catch(rejection => {
+            const output = localize(
+              '`errors.setExceptionBreakpoints.async.failed`',
+              'Failed to configure the exceptions for which to pause due to: {0}',
+              rejection.message || localize('errors.unknown', 'Unknown error'),
+            );
+            this.dap.output({ category: 'stderr', output: output });
+            const telemetry = this._services.get<ITelemetryReporter>(ITelemetryReporter);
+            telemetry.reportOperation(
```
connor4312, line 198: This operation data is already reported by the CDP connection itself, I don't think we need extra bits here

src/adapter/debugAdapter.ts
```diff
@@ -173,7 +173,37 @@ export class DebugAdapter implements IDisposable {
...
+        logger.verbose(LogTag.Internal, 'setExceptionBreakpoints: awaiting blocking response');
+        await result;
+      } else {
+        result
+          .then(() => {
+            logger.verbose(
+              LogTag.Internal,
+              'setExceptionBreakpoints: succesful non-blocking response',
+            );
+          })
```
connor4312, line 189: We get this info from the CDP response in the log (setPauseOnExceptionsState is a one-line function that runs the CDP call) method don't think we need extra stuff here.

src/adapter/debugAdapter.ts
digeff: You are right

PR #549, opened 2020-07-02
Now we can configure the timeouts used for waiting for source-maps

digeff:
In VS we were not hitting breakpoints because the 1 sec source-map timeout wasn't enough for the source-maps to be processed.

Added the "cumulative extra timeout" concept, to address this up to 10 seconds. Also made it configurable if it's not enough for all users.


src/adapter/threads.ts
```diff
@@ -1265,17 +1266,49 @@ export class Thread implements IVariableStoreDelegate {
    */
   async _handleSourceMapPause(scriptId: string, brokenOn: Cdp.Debugger.Location): Promise<boolean> {
     this._pausedForSourceMapScriptId = scriptId;
-    const timeout = this._sourceContainer.sourceMapTimeouts().scriptPaused;
+    const perScriptTimeout = this._sourceContainer.sourceMapTimeouts().scriptPaused;
+    const timeout =
+      perScriptTimeout + this._sourceContainer.sourceMapTimeouts().extraCumulativeScriptPaused;
```
connor4312, line 1271: I make this a single `pauseForSourcemap` option that defaults to `1000ms` rather than adding an extra property that adds on top of the default; easier to reason about.

connor4312: Clarified in chat

src/adapter/threads.ts
```diff
@@ -1265,17 +1266,49 @@ export class Thread implements IVariableStoreDelegate {
...
+    const remainingExtraCumulativeScriptPaused =
+      this._sourceContainer.sourceMapTimeouts().extraCumulativeScriptPaused -
+      Math.max(wallClockTimeBlockedInMs - perScriptTimeout, 0);
+    this._sourceContainer.setSourceMapTimeouts({
+      ...this._sourceContainer.sourceMapTimeouts(),
+      extraCumulativeScriptPaused: remainingExtraCumulativeScriptPaused,
+    });
+    this.logger.verbose(LogTag.Internal, `Blocked execution waiting for source-map`, {
+      timeSpentWallClockInMs: wallClockTimeBlockedInMs,
+      remainingExtraCumulativeScriptPaused: remainingExtraCumulativeScriptPaused,
```
connor4312, line 1295:
```suggestion
      remainingExtraCumulativeScriptPaused,
```

PR #562, opened 2020-07-09
Do not block on source-map if there are no breakpoints

digeff:
Do not block on source-map processing if there are no breakpoints


connor4312:
Nice, this is great!

PR #652, opened 2020-07-23
fix: exchange zh-hans and zh-hant

justjavac:


connor4312:
Thanks for the contribution

@digeff do you still use these translations / do you need to take this change in VS?

connor4312:
We actually no longer need these translations (https://github.com/microsoft/vscode-js-debug/pull/654) since they're in https://github.com/microsoft/vscode-loc now.

Thanks for the PR!

PR #654, opened 2020-07-23
Remove hardcoded localization

digeff:
Remove hardcoded localization.

Also my repository really didn't like the line endings of those files, and keeps fixing them :/


connor4312:
Thanks!

PR #666, opened 2020-07-28
Added customDescriptionGenerator feature to customize the descriptions of the debugger

digeff:
The customDescriptionGenerator feature lets you customize the textural representation that the debugger uses for objects/variables. This is a simple example of how a Fraction object could look (taken from the tsNodeApp demo):
image


rbuckton:
🎉 I was just asking for something like this a few weeks ago: https://github.com/microsoft/vscode/issues/102181

src/adapter/variables.ts
```diff
@@ -570,20 +580,60 @@ export class VariableStore {
...
+    context?: string,
+  ): Promise<string> {
+    const defaultValueDescription =
+      (name === '__proto__' && object.description) ||
+      objectPreview.previewRemoteObject(object, context);
+
+    if (this.customDescriptionGenerator) {
+      const customValueDescription = await this._cdp.Runtime.callFunctionOn({
+        objectId: object.objectId,
+        functionDeclaration: this.customDescriptionGenerator,
```
connor4312, line 612: You could reuse the infrastructure we use for [log statements](https://github.com/microsoft/vscode-js-debug/blob/master/src/adapter/breakpoints/conditions/logPoint.ts#L54-L83) for this description. This would let users write something like `JSON.stringify(x)` instead of `function (x) { return JSON.stringify(x) }`, which is more pleasant to deal with.

I also think capturing errors inline, like the logpoints do, is preferable -- I will be able to see that an error happened if I'm looking for my value, but I don't get a wall of errors in my console whenever I hit a breakpoint. It's also not always clear where the failure happened out of context, even with the name, since variable are collected for all kinds of things that may not be obvious to the user.

src/build/strings.ts
```diff
@@ -226,6 +226,8 @@ const strings = {
     'If set, attaches to the process via the given port. This is generally no longer necessary for Node.js programs and loses the ability to debug child processes, but can be useful in more esoteric scenarios such as with Deno and Docker launches. If set to 0, a random port will be chosen and --inspect-brk added to the launch arguments automatically.',
   'enableContentValidation.description':
     'Toggles whether we verify the contents of files on disk match the ones loaded in the runtime. This is useful in a variety of scenarios and required in some, but can cause issues if you have server-side transformation of scripts, for example.',
+  'customDescriptionGenerator.description':
+    'Customize the textual representation that the debugger shows for objects (local variables, etc...)',
```
connor4312, line 230: It'd be nice if this was a little more helpful. As an end user I wouldn't be sure how to use this.

src/adapter/something.js
```diff
@@ -0,0 +1,9 @@
+function _generatedCode(defaultValue) {
+  try {
+    return (function (def) {
+      return global.customDebuggerDescription(this, def)
+    })(defaultValue);
+  } catch (e) {
+    return e.stack || e.message || String(e);
+  }
+}
```
connor4312, line 9: did you mean to check this and src/adapter/notes.js in?

src/adapter/variables.ts
```diff
@@ -573,21 +584,81 @@ export class VariableStore {
...
+    name: string,
+    value: RemoteObject,
+    object: Cdp.Runtime.RemoteObject,
+    context?: string,
+  ): Promise<string> {
+    const defaultValueDescription =
+      (name === '__proto__' && object.description) ||
+      objectPreview.previewRemoteObject(object, context);
+
+    if (this.customDescriptionGenerator) {
```
connor4312, line 613: I'd invert this `if` to reduce nesting

src/common/sourceCodeManipulations.ts
```diff
@@ -0,0 +1,108 @@
...
+    return codeToFunctionExecutingCodeAndReturningErrors(parameterNames, callFunctionCode, true);
+  } else {
+    return statementToFunctionReturningErrors(parameterNames, statements, true);
+  }
+}
+
+/**
+ * code => (parameterNames) => return catchAndReturnErrors(code)
+ * */
+export function codeToFunctionExecutingCodeAndReturningErrors(
```
connor4312, line 30: This is only used internally

```suggestion
function codeToFunctionExecutingCodeAndReturningErrors(
```

testWorkspace/tsNodeApp/.vscode/launch.json
```diff
@@ -15,7 +15,8 @@
             "program": "${workspaceFolder}\\app.ts",
             "outFiles": [
                 "${workspaceFolder}/**/*.js"
-            ]
+            ],
+            "customDescriptionGenerator": "const lal = 4; lal + global.customDebuggerDescription(this, defaultValue)",
```
connor4312, line 19: Probably undo the changes to this demo -- or move it to a different demo for your work

src/adapter/something.js
digeff: Removed

src/adapter/variables.ts
digeff: Done

src/common/sourceCodeManipulations.ts
digeff: Done

testWorkspace/tsNodeApp/.vscode/launch.json
digeff: Moved to it's own demo

PR #667, opened 2020-07-29
feat: support node remote debug through ws

denghongcai:
when specify a ws addr, skip discover process

### DIFF
--- a/src/cdp/webSocketTransport.ts
+++ b/src/cdp/webSocketTransport.ts
@@ -7,7 +7,7 @@ import WebSocket from 'ws';
 import { timeoutPromise } from '../common/cancellation';
 import { EventEmitter } from '../common/events';
 import { HrTime } from '../common/hrnow';
-import { isLoopback } from '../common/urlUtils';
+import { isLoopback, isIpHostname } from '../common/urlUtils';
 import { ITransport } from './transport';
 
 export class WebSocketTransport implements ITransport {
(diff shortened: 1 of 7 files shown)


src/cdp/webSocketTransport.ts
```diff
@@ -28,13 +28,21 @@ export class WebSocketTransport implements ITransport {
...
       rejectUnauthorized: !(isSecure && targetAddressIsLoopback),
       followRedirects: true,
-    });
+    };
+
+    if (targetAddressIsLoopback) {
+      options.headers = {
+        host: 'localhost',
+      };
+    }
```
connor4312, line 43: This is intentionally added all the time -- by default chrome will not respond to connections made remotely (i.e. to a non-local Host) as a security measure against attacks from the browser.

src/configuration.ts
```diff
@@ -468,6 +468,11 @@ export interface INodeAttachConfiguration extends INodeBaseConfiguration {
   type: DebugType.Node;
   request: 'attach';
 
+  /**
+   * Inspector webSocket address
+   */
+  wsAddr?: string;
```
connor4312, line 474: Can you rename this to `websocketAddress` and add it to the `generate-contributions.ts` as well? The generation file is what is used to create the package.json contributions that contains the schema for the launch.json file.

src/cdp/webSocketTransport.ts
denghongcai: in private network, websocket may proxied by NGINX or other reverse proxy.

change it to add `localhost` header only when url hostname is ip.

src/configuration.ts
denghongcai: fixed.

src/cdp/webSocketTransport.ts
connor4312: I think I'd rather keep doing it all the time.

Setting the host to localhost _could_ break in the reverse proxy scenario and require changes to the proxy to allow it to work.

Not doing so _will_ break lookups that use DNS and don't use a reverse proxy to correct the "host" header sent to the debuggee.

denghongcai: if add `wss=` or `ws=` query to url for Chrome devtools
`devtools://devtools/bundled/inspector.html?experiments=true&ws=demo:41000`
it will pass `demo` as Host header, differed from js-debug.

I hava another idea, create a new extension to simulate discover service (`/json`) and reverse proxy websocket instead of modify js-debug.

connor4312: > I hava another idea, create a new extension to simulate discover service (/json) and reverse proxy websocket instead of modify js-debug.

Alternately, it could be a config option for `attach` configuration, like `remoteHostHeader`

connor4312:
I've merged the custom websocket part of this in https://github.com/microsoft/vscode-js-debug/commit/b2853581edf91a9cb95e7dbd2bc3f91f602b1e85. I'm willing to take another PR to make the `Host` header configurable 🙂

PR #683, opened 2020-08-03
fix: don't launch chrome with a cwd that doesn't exist

EricCornelson:


src/targets/browser/launcher.ts
```diff
@@ -98,7 +99,7 @@ export async function launch(
     const cp = childProcess.spawn(executablePath, browserArguments.toArray(), {
       detached: true,
       env: env.defined(),
-      cwd,
+      cwd: existsSync(cwd) ? cwd : process.cwd(),
```
connor4312, line 102: Since this function is async already, you can use `canAccess` from the `fsUtils` file to avoid blocking the thread :)

EricCornelson: done!

PR #693, opened 2020-08-10
Add support for old CDP APIs for back compatibility with ios-webkit-debug-proxy

### LINKED ISSUE #647: Add backward compatibility for Console Domain and support of the Debugger.setBreakpointsActive method
Hello, this is [Cordova Tools](https://github.com/microsoft/vscode-cordova) team. We are adapting debugging of Ionic and Cordova apps on iOS to `js-debug` approach. We use [ios-webkit-debug-proxy](https://github.com/google/ios-webkit-debug-proxy) tool to be able to debug iOS apps via Chrome Debug Protocol. IWDP uses quite old version of Chrome Dev Tools and requires deprecated [`Console Domain`](https://chromedevtools.github.io/devtools-protocol/tot/Console/) to process console messages. But `js-debug` doesn't support this domain. It would be awesome if it had backward compatibility for old `Console Domain`.
We also found out that IWDP requires the debugger to send the [`Debugger.setBreakpointsActive`](https://chromedevtools.github.io/devtools-protocol/tot/Debugger/#method-setBreakpointsActive) message in order to activate breakpoints, but `js-debug` doesn't use it.
According to these unsupported methods we could try to implement them on our [vscode-cdp-proxy](https://github.com/microsoft/vscode-cdp-proxy) side, but since the debugger counts messages ids, we can break the ids order when we inject additional messages to the original sequence. We think that adding of additional messages from our side is fairly unreliable.
We'd like to discuss how it is possible to implement the support of the `Console Domain` (at least [`Console.enable`](https://chromedevtools.github.io/devtools-protocol/tot/Console/#method-enable) method at the beginning of debugging) and `Debugger.setBreakpointsActive` method.
At the current version of the Cordova Tools extension we use `vscode-chrome-debug-core` and it debugs iOS apps without problems. I attach logs of CDP messages for `vscode-chrome-debug-core` and `js-debug` to make it easier to find difference between these approaches:
- [vscode-chrome-debug-core logs](https://github.com/microsoft/vscode-js-debug/files/4953379/logCor.txt)
- [js-debug logs](https://github.com/microsoft/vscode-js-debug/files/4953381/logCor3042.txt)

SounD120:
Fixes #647
Hi @connor4312 . Ported some handlers from [vscode-chrome-debug-core](https://github.com/microsoft/vscode-chrome-debug-core) for compatibility with updated Cordova Tools iOS debugger and React Native Tools iOS direct debugger with [ios-webkit-debug-proxy](https://github.com/google/ios-webkit-debug-proxy) usage.
Please let me know if any additional checks or tests for these scenarios are needed.

### DIFF
--- a/src/adapter/threads.ts
+++ b/src/adapter/threads.ts
@@ -497,6 +497,7 @@ export class Thread implements IVariableStoreDelegate {
       this._executionContextsCleared();
     });
     if (this.launchConfig.outputCapture === OutputSource.Console) {
+      this._cdp.Console.on('messageAdded', async event => this._onMessageAdded(event));
       this._cdp.Runtime.on('consoleAPICalled', async event => {
         const slot = this._claimOutputSlot();
         slot(await this._onConsoleMessage(event));
@@ -622,6 +623,25 @@ export class Thread implements IVariableStoreDelegate {
     this._cdp.Debugger.enable({}).then(response => {
       if (response) Thread._allThreadsByDebuggerId.set(response.debuggerId, this);
     });
+  }
+
+  /**
+   * For backcompat, also listen to Console.messageAdded, only if it looks like the old format.
+   */
+  // eslint-disable-next-line
+  private async _onMessageAdded(event: any) {
+    // message.type is undefined when Runtime.consoleAPICalled is being sent
+    if (event && event.message && event.message.type) {
+      const onConsoleAPICalledParams: Cdp.Runtime.ConsoleAPICalledEvent = {
+        type: event.message.type,
+        timestamp: event.message.timestamp,
+        args: event.message.parameters || [{ type: 'string', value: event.message.text }],
+        stackTrace: event.message.stack,
+        executionContextId: 1,
+      };
+      const slot = this._claimOutputSlot();
+      slot(await this._onConsoleMessage(onConsoleAPICalledParams));
+    }
   }
 
   private async _onPaused(event: Cdp.Debugger.PausedEvent) {
--- a/src/adapter/threads.ts
+++ b/src/adapter/threads.ts
@@ -515,6 +515,10 @@ export class Thread implements IVariableStoreDelegate {
     this._cdp.Debugger.on('paused', async event => this._onPaused(event));
     this._cdp.Debugger.on('resumed', () => this.onResumed());
     this._cdp.Debugger.on('scriptParsed', event => this._onScriptParsed(event));
+    this._cdp.Console.enable({}).catch(() => {
+      /* Specifically ignore a fail here since it's only for backcompat */
+    });
+    this._cdp.Debugger.setBreakpointsActive({ active: true }); // This call is used for some non-standard runtimes like: https://github.com/google/ios-webkit-debug-proxy
     this._cdp.Runtime.enable({});
 
     if (!this.launchConfig.noDebug) {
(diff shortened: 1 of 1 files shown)


connor4312:
Do you run your own proxy between js-debug and the ios-webkit-debug-proxy? If so, could these additional messages/transforms be added there?

SounD120:
The problem is that this may break the sequence of messages (messages id) sent to js-debug, do you think that it is better to implement that this way?

connor4312:
In js-debug we don't care whether IDs are sequential -- we just use it as an easy way to have unique message IDs. Though you could also capture and ignore the response to those method calls within the proxy if you wanted.

SounD120:
@connor4312 I think we can also capture IDs but I'm not sure if it will not break something in js-debug if we send some commands from the proxy.  How are IDs calculated in js-debug?
How can we be sure that the IDs we send are unique in js-debug context?

connor4312:
IDs in js-debug are an incrementing number. I just pushed a change to start the js-debug IDs at 1000. So to generate IDs you can use safely in the proxy you can do something like

```
let idCounter = 0;

// ...

const callId = (idCounter ++) % 1000;
```

If an ID comes to js-debug for which there's no corresponding request, the message will be ignored.

SounD120:
We discussed that with @EricCornelson , and we are not sure that it will be safe to modify IDs that are sent to js-debug. I think that inside of the proxy we don't have a safe mechanism to control IDs consistency, so we cannot be sure that the IDs that we will try to use are not yet busy in js-debug context, also we cannot control numbering of js-debug's IDs. It might lead to breakages when numbering on js-debug side is changed.
Also on the proxy side that feature will be harder to implement since in the proxy we cannot just call a few functions with CDP API, but we have to gather such requests ourselves and then send it to the debugger. So this might lead to duplicating functions  from js-debug inside the proxy.
So maybe it makes sense to consider the method proposed in that PR? Let us know if you have any concerns about it.

SounD120:
@connor4312 Please let us know if you have any concerns on this PR. It is blocking us a bit, so we'd like to pick the best solution for that situation ASAP.

connor4312:
>  we are not sure that it will be safe to modify IDs that are sent to js-debug. I think that inside of the proxy we don't have a safe mechanism to control IDs consistency, so we cannot be sure that the IDs that we will try to use are not yet busy in js-debug context, also we cannot control numbering of js-debug's IDs.

js-debug will never use CDP IDs <1000, so that space is safe for you to use. It was an intention of the proxy design to allow adding extra calls like this, in addition to transcription, so if this doesn't work please let me know!

> Also on the proxy side that feature will be harder to implement since in the proxy we cannot just call a few functions with CDP API, but we have to gather such requests ourselves and then send it to the debugger.

It seems like you can reliably make these changes in the proxy by:

1. Sending `Console.enable({})` and `Debugger.setBreakpointsActive({ active: true })` after you see a `Debugger.enable()` call come through the connection.
1. Transcribe `Console.messageAdded` events to `Runtime,consoleAPICalled` using the logic you have in `_onMessageAdded`

SounD120:
@connor4312 Hm, we'll give it a try, thanks. I didn't know that we can send CDP commands inside of the proxy via the Connection object 😄  . Let this PR stay open for a while, we'll close it if this solution will work for us.

connor4312:
Sounds good, let me know if you need help with anything

RedMickey:
Hi @connor4312, we implemented this fix on our CDP proxy side. Thank you for your help. Please close this PR, it isn't needed anymore.

connor4312:
Okay, good to hear you got that working! Ping me if you need anything else. Apologies for the delay on this and the other issue, they came in as js-debug became is VS Code default and my triage list was quite full.

RedMickey:
That's ok, we understand that there were a lot of work at that time. 😄

PR #694, opened 2020-08-10
don't smart step entry breakpoint

Shenniey:
We run into the issue of the debugger pausing on the first line when we have smart stepping turned on for the entry breakpoint.

### DIFF
--- a/src/adapter/smartStepping.ts
+++ b/src/adapter/smartStepping.ts
@@ -37,7 +37,7 @@ export class SmartStepper {
   constructor(
     private readonly launchConfig: AnyLaunchConfiguration,
     private readonly logger: ILogger,
-  ) { }
+  ) {}
 
   private resetSmartStepCount(): void {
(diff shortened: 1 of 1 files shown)


connor4312:
Actually, I'm a little confused. The `entry` breakpoint is used when launching with `stopOnEntry: true` or an --inspect-brk. In this case the user wants to stay paused on the first line.

I actually agree that we shouldn't smartStep here, but 'pausing on the first line' is intended behavior for entry BPs.

Shenniey:
The problem we're running into is that when we hit the entry breakpoint, we step into the smartstepper, who then changes the reason of the bp from 'entry' to 'step'. Then when that gets passed back to the VS side of things and we want to auto-resume the entrypoint bp, we can't recognize the bp as the entrypoint bp.

connor4312:
Ah, gotcha. So this actually fixes multiple things 🙂

PR #699, opened 2020-08-12
feat: Create the VS Debug Server mode

EricCornelson:
This change is part of an effort for the VS implementation to move away from flat-session mode, and to as closely as possible follow the VS Code implementation, while sharing as much code as possible to avoid breakages due to divergent implementations.

This change moves most of the debug server creation logic out of vsCodeSessionManager and in to its own class (serverSessionManager), and adds the vsDebugServer entry-point which hosts a debug server without any VS Code API dependencies.

### DIFF
--- a/src/serverSessionManager.ts
+++ b/src/serverSessionManager.ts
@@ -100,6 +100,7 @@ export class ServerSessionManager<T extends IDebugSessionLike> {
       const session = sessionCreationFunc(transport);
       deferredConnection.resolve(session.connection);
     });
+    debugServer.on('error', deferredConnection.reject);
     debugServer.listen(debugServerPort || 0);
     this.servers.set(debugSession.id, debugServer);
 
--- a/src/serverSessionManager.ts
+++ b/src/serverSessionManager.ts
@@ -1,0 +1,124 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+import { SessionManager, ISessionLauncher, IDebugSessionLike, Session } from './sessionManager';
+import { IDisposable } from './common/disposable';
+import * as net from 'net';
+import { Container } from 'inversify';
+import { StreamDapTransport, IDapTransport } from './dap/transport';
+import { IPseudoAttachConfiguration } from './configuration';
+import { getDeferred, IDeferred } from './common/promiseUtil';
+import DapConnection from './dap/connection';
+import { Readable, Writable } from 'stream';
+
+interface IDebugServerCreateResult {
+  server: net.Server;
+  connectionPromise: Promise<DapConnection>;
+}
+
+/**
+ * A class for handling specifically server-based sessions in js-debug
+ */
+export class ServerSessionManager<T extends IDebugSessionLike> {
+  private readonly sessionManager: SessionManager<T>;
+  private disposables: IDisposable[] = [];
+  private servers = new Map<string, net.Server>();
+
+  constructor(globalContainer: Container, sessionLauncher: ISessionLauncher<T>) {
+    this.sessionManager = new SessionManager(globalContainer, sessionLauncher);
+    this.disposables.push(this.sessionManager);
+  }
+
+  /**
+   * Create the appropriate debug server type from the configuration passed in the debug session
+   * @param debugSession The IDE-specific debug session
+   * @param debugServerPort Optional debug port to specify the listening port for the server
+   */
+  public createDebugServer(debugSession: T, debugServerPort?: number) {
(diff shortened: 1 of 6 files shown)


src/vsDebugServer.ts
```diff
@@ -0,0 +1,137 @@
...
+    } else {
+      this.launchRoot();
+    }
+  }
+
+  launchRootFromExisting(inputStream: Readable, outputStream: Writable) {
+    const deferredConnection: IDeferred<DapConnection> = getDeferred();
+    const session = new VSDebugSession(
+      'root',
+      'JavaScript debugger root session',
```
digeff, line 70: Is this shown anywhere? Should we localize it?

src/vsDebugServer.ts
```diff
@@ -0,0 +1,137 @@
...
+    if (inputStream && outputStream) {
+      this.launchRootFromExisting(inputStream, outputStream);
+    } else {
+      this.launchRoot();
+    }
+  }
+
+  launchRootFromExisting(inputStream: Readable, outputStream: Writable) {
+    const deferredConnection: IDeferred<DapConnection> = getDeferred();
+    const session = new VSDebugSession(
```
digeff, line 68: Is it there an easy way to merge this code with lines 83-84?

EricCornelson: whoops! meant to clean that up before I submitted. Good catch.

src/vsDebugServer.ts
```diff
@@ -0,0 +1,137 @@
...
+    private readonly childConnection: Promise<DapConnection>,
+    public readonly configuration: DebugConfiguration,
+  ) {
+    this._name = name;
+  }
+
+  private _name: string;
+  set name(newName: string) {
+    this._name = newName;
+    this.childConnection
```
digeff, line 40: Can we run into any issues because we do not await this operation?

src/vsDebugServer.ts
```diff
@@ -0,0 +1,137 @@
...
+    const childAttachConfig = { ...config, sessionId: target.id, __jsDebugChildServer: '' };
+    const deferredConnection: IDeferred<DapConnection> = getDeferred();
+    const session = new VSDebugSession(
+      target.id(),
+      target.name(),
+      deferredConnection.promise,
+      childAttachConfig,
+    );
+    const result = this.sessionServer.createChildDebugServer(session);
+    result.connectionPromise.then(x => deferredConnection.resolve(x));
```
digeff, line 109: Can result.connectionPromise be rejected?

src/vsDebugServer.ts
EricCornelson: It's potentially shown in the process window and process drop-down in VS (though it isn't being shown right now). I'm actually inclined to just pass an empty string rather than trying to localize this, as it's more an implementation detail than something users care about.

src/vsDebugServer.ts
EricCornelson: We specifically _don't_ want to await this, since it would block init code. We just want to send the process event with the target name once we are initialized.

src/vsDebugServer.ts
EricCornelson: It will never be rejected. It could fail to ever resolve if VS never connects to that session's server.

EricCornelson:
> Do we have enough telemetry to understand if things are working as expected on production?
> Do we have enough logging to diagnose issues that may happen to customers?

I believe so. Any issues with launching/attaching to a session should be captured by PineZorro events, same with sessions disconnecting unexpectedly.

src/serverSessionManager.ts
```diff
@@ -0,0 +1,124 @@
...
+    sessionCreationFunc: (transport: IDapTransport) => Session<T>,
+    debugServerPort?: number,
+  ): IDebugServerCreateResult {
+    const deferredConnection: IDeferred<DapConnection> = getDeferred();
+    const debugServer = net.createServer(socket => {
+      const transport = new StreamDapTransport(socket, socket);
+      const session = sessionCreationFunc(transport);
+      deferredConnection.resolve(session.connection);
+    });
+    debugServer.listen(debugServerPort || 0);
```
connor4312, line 103:
```suggestion
    debugServer.on('error', deferredConnection.reject);
    debugServer.listen(debugServerPort || 0);
```

src/serverSessionManager.ts
```diff
@@ -0,0 +1,124 @@
...
+    this.sessionManager.terminate(debugSession);
+    this.servers.get(debugSession.id)?.close();
+    this.servers.delete(debugSession.id);
+  }
+
+  /**
+   * @inheritdoc
+   */
+  public dispose() {
+    this.disposables.forEach(d => d.dispose());
```
connor4312, line 122: you only have one disposable in here

```suggestion
    this.sessionManager.dispose();
```

PR #706, opened 2020-08-17
fix: race condition which leaves dangling debug sessions

EricCornelson:
Fixing a race condition that can leave dangling debug sessions (i.e. debug sessions which have no active CDP target, but don't yet know it).

### DIFF
--- a/src/binder.ts
+++ b/src/binder.ts
@@ -14,7 +14,7 @@ import { CancellationTokenSource } from './common/cancellation';
 import { EventEmitter, IDisposable } from './common/events';
 import { ILogger, LogTag, resolveLoggerOptions } from './common/logging';
 import { mapValues } from './common/objUtils';
-import { delay } from './common/promiseUtil';
+import { delay, getDeferred, IDeferred } from './common/promiseUtil';
 import * as urlUtils from './common/urlUtils';
 import {
   AnyLaunchConfiguration,
@@ -45,10 +45,12 @@ export interface IBinderDelegate {
   releaseDap(target: ITarget): void;
 }
 
+type ThreadData = { thread: Thread; debugAdapter: DebugAdapter };
+
 export class Binder implements IDisposable {
   private _delegate: IBinderDelegate;
   private _disposables: IDisposable[];
-  private _threads = new Map<ITarget, { thread: Thread; debugAdapter: DebugAdapter }>();
+  private _threads = new Map<ITarget, IDeferred<ThreadData>>();
   private _terminationCount = 0;
   private _onTargetListChangedEmitter = new EventEmitter<void>();
   readonly onTargetListChanged = this._onTargetListChangedEmitter.event;
@@ -349,7 +351,7 @@ export class Binder implements IDisposable {
     return result;
   }
 
-  public async attach(target: ITarget, launcher: ILauncher) {
+  public async attach(target: ITarget, threadData: IDeferred<ThreadData>, launcher: ILauncher) {
     if (!this._launchParams) {
       throw new Error('Cannot launch before params have been set');
     }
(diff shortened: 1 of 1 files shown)


src/binder.ts
```diff
@@ -382,7 +382,17 @@ export class Binder implements IDisposable {
...
     const startThread = async () => {
       await debugAdapter.launchBlocker();
-      cdp.Runtime.runIfWaitingForDebugger({});
+      const result = await cdp.Runtime.runIfWaitingForDebugger({});
+      if (!result) {
+        // There is a race condition here (especially in VS) where the user may have refreshed the page
+        // or the target may have otherwise been killed before we connected the DAP session.
+        // if the result of runIfWaitingForDebugger comes back undefined, it means the target connection
+        // is dead, BUT, we don't want to error the launch because technically the target already launched
+        // and terminated, so we send a successful launch response but immediately follow it with a terminated event
```
connor4312, line 391: Is there some way we can detect this in e.g. the browser launcher and send a termination event there? This sends the right response to the UI but internally the binder will still be waiting for a terminated event from the launcher.

EricCornelson: Do child targets go through the browser launcher? For this fix, I was looking particularly at web workers

connor4312: Yea, they're tracked in the browserTargetManager that keeps a list of them. The state update should happen there and propagate down to the binder

EricCornelson: So, I'm a little confused about the flow here, and how things _should_ be working. The binder will attach to a target when it's created, and then `await` the DAP connection for the corresponding debug session, at which point it wires up the DAP listeners. When the target is destroyed, the binder will be notified via the launcher's `onTargetListChanged` event, and it will call `_detachOrphanThreads` then `_releaseTarget` which disposes the debugAdapter object.

But the problem is that `_detachOrphanThreads` works off of the current threads map, and the thread for the target doesn't get added to the map until _after_ the DAP connection is established, which means that `_releaseTarget` doesn't even get called for that target.

So the launcher is doing everything properly, and knows that the targets are detached, but the binder's state gets out of sync because it's not listening for events on a thread that hasn't yet established a connection. One solution might be putting yet another "launch blocker" style promise which we can await in the `_releaseTarget` method, and then just add the thread to the thread's map earlier so that it will trigger that event but wait until the launch event is received to fire, but I feel like the more of those promises we add, the harder the flow is to understand end-to-end.

Would it work to just call `_releaseTarget` manually in launch if we find that it's no longer connected?

connor4312: If I understand, correctly, maybe a more graceful approach would be to:

- Have `this._threads.set(target` be set from within `_attachToNewTargets`, where currently it's only set after the DA is acquired which I believe leads to this race
- Have the thread store a `Promise<DebugAdapter | undefined>` instead of the `DebugAdapter` directly (which it could get from calling `attach()`)
- (if needed?) have it also store a CancellationTokenSource which can be triggered on detach,

EricCornelson: The problem with that approach is that if we abort the attach, we'll get an error dialog in VS that the debug session launch has failed. In order to not get that error, we need to "go through the motions" and clean up the debug session _after_ the launch event.

connor4312: I think in this scenario the `Promise<DebugAdapter | undefined>` would not resolve until `startThread` has been run. (We should probably also Promise.race the launchBlocker against the CancellationToken in case it's being slow and the user asks to terminate.) So the termination logic would not begin until that point.

I think I'd favor that approach over what's in the PR because, although the binder is not reused between sessions and in the VS case neither is the process, keeps the setup and teardown logic unified and avoids a special race-detection code path.

EricCornelson: Got it, that makes sense. Updated the PR, please take another look.

connor4312 (approved):

PR #707, opened 2020-08-18
fix: block DAP messages from being sent after responding to the disco…

EricCornelson:
…nnect request, and add error handling on transport stream writes


connor4312:
Can you take a look at the failed test?

EricCornelson:
Looks like all the tests are passing. Not sure what's going on with MacOS. It looks like the tests are passing but the npm run tests task is timing out?

connor4312:
Yea, I need to look at that

PR #709, opened 2020-08-18
Support for using remote FileSystem

digeff:
* Add support to use a remote FileSystem. If __remoteFilePrefix is specified, when we do a fs.exists operation, if the filePath starts with __remoteFilePrefix, the operation will be performed by asking the DAP client to perform this operation.
* Modified the code using fsUtils.exist to consume fsUtils via injection

**This is the first iteration of the remote file system, we'll need further changes in the future (such as reading source-maps using remote file operations too)** I'm sending this now because we already have a lot of changes, so as to keep the PR to a reasonable level


src/binder.ts
```diff
@@ -118,15 +120,20 @@ export class Binder implements IDisposable {
...
         ),
       );
-      dap.on('launch', params =>
-        this._boot(
+      dap.on('launch', params => {
+        const remoteFilePrefix = (params as { __remoteFilePrefix: string | undefined })
+          .__remoteFilePrefix;
+        this._rootServices
+          .bind(FSUtils)
+          .toConstantValue(LocalAndRemoteFsUtils.create(remoteFilePrefix, fsPromises, dap));
```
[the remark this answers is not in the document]
connor4312: You should be able to get rid of this now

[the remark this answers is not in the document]
digeff: Done!

connor4312:
Looks like a build failure, otherwise lgtm

src/common/sourceUtils.ts
```diff
@@ -241,7 +241,7 @@ export async function checkContentHash(
   }
 
   if (!contentHash) {
-    const exists = await fsUtils.exists(absolutePath);
+    const exists = fs.existsSync(absolutePath);
```
connor4312, line 244: why was this change made? We don't want to block the main thread of FS operations if it can be helped

digeff: That was my mistake. Good catch!

src/ioc.ts
```diff
@@ -58,13 +59,15 @@ import {
...
   FS,
+  FSUtils,
   IContainer,
   IsVSCode,
   ProcessEnv,
   SessionSubStates,
   StoragePath,
   trackDispose,
   VSCodeApi,
+  //  FSUtils,
```
connor4312, line 70:
```suggestion
```

src/test/goldenText.ts
```diff
@@ -50,8 +50,20 @@ export class GoldenText {
       const match = frame.match(/^(.*):(\d+):(\d+)$/);
       if (!match) return null;
       const filePath = match[1];
-      if (filePath === __filename) continue;
-      return filePath;
+
+      // We need to consider .ts files because we may have source-map-support to convert stack traces to .ts
```
connor4312, line 54: This seems dangerous in tests, where precise behavior js-debug's of source maps handling is tested

src/ui/configuration/nodeDebugConfigurationResolver.ts
```diff
@@ -45,6 +47,8 @@ export class NodeConfigurationResolver extends BaseConfigurationResolver<AnyNode
   constructor(
     @inject(ExtensionContext) context: vscode.ExtensionContext,
     @inject(INvmResolver) private readonly nvmResolver: INvmResolver,
+    // We don't need Node to support remote file system at the moment, so we hardcode fsUtils
```
connor4312, line 50: What I would do instead for these cases is inject the local FS utils in `createGlobalContainer`, since we know that for services that operate independently of a DAP connection we will never be able to do anything else. Then in these cases you can inject the fs utils normally.

src/ioc-extras.ts
```diff
@@ -53,6 +53,11 @@ export const Execa = Symbol('execa');
  */
 export const FS = Symbol('FS');
 
+/**
+ * Injection for the `FsUtils` module.
+ */
+export const FSUtils = Symbol('FsUtils');
```
connor4312, line 59: Since the FS utils are an interface we own, please rename this to `IFSUtils` and place it in the file where that interface is declared. This will merge the declarations, e.g.

https://github.com/microsoft/vscode-js-debug/blob/25d128bf4749da1c301bb62e9cd3a7cd143bc503/src/adapter/performance/index.ts#L12-L19

digeff: Done

src/ui/configuration/nodeDebugConfigurationResolver.ts
digeff: Done

src/test/goldenText.ts
digeff: As discussed offline, this only affect golden test files

connor4312 (approved):

PR #715, opened 2020-08-20
Ensure source.rootPaths being defined when Node is a child session of Chrome

digeff:
We were having some issues on Node sessions that were a child of a Chrome session because they were started without the rootSession or cwd configured, and so we couldn't configure the source.rootPath so we ended up showing absolute paths for the loaded script files. With this fix, we show relative file paths instead

### DIFF
--- a/src/adapter/sources.ts
+++ b/src/adapter/sources.ts
@@ -490,7 +490,8 @@ export class SourceContainer {
   ) {
     this._dap = dap;
 
-    const mainRootPath = 'webRoot' in launchConfig ? launchConfig.webRoot : launchConfig.rootPath;
+    const mainRootPath =
+      'webRoot' in launchConfig ? launchConfig.webRoot : launchConfig.rootPath || launchConfig.cwd;
     if (mainRootPath) {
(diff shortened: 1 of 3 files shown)


connor4312:
When are Node sessions children of Chrome sessions?

digeff:
@connor4312  We use the server launch.json argument to launch a Chrome session that also has a node session

connor4312:
Ah, yea. In retrospect that was a bad idea, done since `vscode-pwa` had that originally, but a lot of the path resolution stuff isn't the same between them...

digeff:
@connor4312 What would you suggest as the best way to handle this scenario? Is there a different feature/parameter we can use?

connor4312:
Not totally sure. I was just looking at our data and wondering whether this is something we want to keep. The number of VS Code users using the `server` is extremely small and, at least in VS Code, this can be accomplished with compound launch configs or prelaunch tasks.

Thoughts? Do you guys have many people using this feature?

digeff:
@connor4312 We'll try to change our implementation to use two different sessions started independently. Let's hold off on this PR to see if that approach will work without issues.

PR #716, opened 2020-08-20
Skip Bootloader to hide it from users

digeff:
This fixes all the bootloader sources appearing on the "Loaded scripts" list (which the user doesn't care about)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -273,6 +273,11 @@ gulp.task('flatSessionBundle:webpack-bundle', async () => {
   return runWebpack({ packages, devtool: 'nosources-source-map' });
 });
 
+gulp.task('markBootloaderAsCDPFile', done => {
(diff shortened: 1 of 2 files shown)


connor4312:
Thanks for the PR. I've thought about hiding internal stuff in general within loaded scripts. Logpoints use the known format `logpoint-<hash>.cdp` so that we can hide them in the stacktraces ([ref](https://github.com/microsoft/vscode-js-debug/blob/32244ec6b0c979f79a1808cbb506e6ef5900d6b2/src/adapter/breakpoints/conditions/logPoint.ts#L27)). Maybe we should generalize and just say we ignore all `*.cdp` scripts.

With that in place, we can add a comment like `//# sourceURL=bootloader.cdp` at the end of the bootloader bundler, which would be captured by that. Doing it this way would be more reliable, since there are cases where the bootloader might move from its default location (auto attach, or if there's a space in the bootloader path and the user is using Node <12.)

PR #747, opened 2020-09-10
fix: broken VS error telemetry

EricCornelson:
Visual Studio's telemetry APIs reject properties with exclamation points. Adding back the regular error property name for VS only.


src/telemetry/unhandledErrorReporter.ts
```diff
@@ -15,9 +15,20 @@ export const enum ErrorType {
 export function installUnhandledErrorReporter(
   logger: ILogger,
   telemetryReporter: ITelemetryReporter,
+  isVsCode?: boolean,
```
connor4312, line 18: Can you make this required, or default to `true` in order to be conservative about the data we send?

EricCornelson: done

PR #750, opened 2020-09-11
await nyc.report() or no test coverage gets recorded

dcermak:
`report` is an asynchronous function and should be awaited otherwise the test coverage might not get written to the disk


connor4312:
thanks!

PR #791, opened 2020-10-07
Debugger does not work when running Node.js on an unsupported operating system

hp8wvvvgnj6asjm7:
Fixes https://github.com/microsoft/vscode/issues/108222


connor4312:
Thanks!

This will be in the next [nightly build](https://github.com/microsoft/vscode-js-debug#nightly-extension)

PR #793, opened 2020-10-07
Add support for customDebuggerProperties

digeff:
Lets user configure which properties to show for an object in the locals view, watch view, etc...

Known issues:
 * Customizing the properties of an array show the properties both flattened, and inside [0..99][100..199] etc...


connor4312:
Can you describe a little more your use case for this?

digeff:
I want to change what variables the debugger show for objects. Some examples:
  - Map currently show an [[Entries]] variable, and you need to dive inside that to see which elements the map has. With this feature I can modify what variables we print, to show the elements directly
  - If I have a Decorator, Wrapper or Proxy pattern, I can merge the variables of the original objects, and the decorated or wrapped or proxied objects, and show everything in a single level, instead of needing to dig multiple levels to find them

In case it's helpful, this page describes a similar feature for the C# debugger: https://docs.microsoft.com/en-us/dotnet/framework/debug-trace-profile/enhancing-debugging-with-the-debugger-display-attributes

digeff:
@connor4312 What do you think of using this as the first version of this feature, so we can start delivering value and getting feedback, and we can iterate on top of this to improve the UI/UX for how to configure it?

connor4312:
Something is weird with this diff -- it's including all the changes made in master already

digeff:
@connor4312 I just redid the commit/branch to fix the issue

PR #796, opened 2020-10-08
Refactoring to create absolutePathToUrlRegexp

digeff:
Start using absolutePathToUrlRegexp instead of absolutePathToUrl so we can customize the Regexp we use for Blazor scenarios  

@connor4312 If you like this approach, I'll add some unit tests for BrowserSourcePathResolver.absolutePathToUrlRegexp

### DIFF
--- a/src/adapter/breakpoints/breakpointBase.ts
+++ b/src/adapter/breakpoints/breakpointBase.ts
@@ -438,15 +438,21 @@ export abstract class Breakpoint {
   protected hasSetOnLocation(script: Partial<Script>, lineColumn: LineColumn) {
     return this.cdpBreakpoints.find(
       bp =>
-        (script.url &&
-          isSetByUrl(bp.args) &&
-          (new RegExp(bp.args.urlRegex ?? '').test(script.url) ||
-            (script.urlRegexp && bp.args.urlRegex === script.urlRegexp)) &&
-          lcEqual(bp.args, lineColumn)) ||
-        (script.scriptId &&
-          isSetByLocation(bp.args) &&
-          bp.args.location.scriptId === script.scriptId &&
-          lcEqual(bp.args.location, lineColumn)),
+        script.scriptId &&
+        isSetByLocation(bp.args) &&
+        bp.args.location.scriptId === script.scriptId &&
+        lcEqual(bp.args.location, lineColumn),
+    );
+  }
+
+  /**
+   * Returns whether a breakpoint has been set on the given line and column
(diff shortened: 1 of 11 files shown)


src/adapter/threads.ts
```diff
@@ -92,6 +92,7 @@ export class ExecutionContext {
 
 export type Script = {
   url: string;
+  urlRegexp?: string;
```
connor4312, line 95: This should not be a property of Scripts

src/adapter/breakpoints/breakpointBase.ts
```diff
@@ -434,7 +440,8 @@ export abstract class Breakpoint {
       bp =>
         (script.url &&
           isSetByUrl(bp.args) &&
-          new RegExp(bp.args.urlRegex ?? '').test(script.url) &&
+          (new RegExp(bp.args.urlRegex ?? '').test(script.url) ||
+            (script.urlRegexp && bp.args.urlRegex === script.urlRegexp)) &&
```
connor4312, line 444: Since we never set by URL, this function signature and logic can be simplified

src/targets/browser/browserPathResolver.ts
```diff
@@ -2,44 +2,99 @@
...
   baseUrl?: string;
   pathMapping: PathMapping;
   clientID: string | undefined;
+  isBlazor: boolean;
+  remoteFilePrefix: string | undefined;
 }
 
 @injectable()
 export class BrowserSourcePathResolver extends SourcePathResolverBase<IOptions> {
+  private readonly blazorInCodespacesRegexp: RegExp;
```
connor4312, line 38: Please make a new `BlazorSourcePathResolver` that extends the `BrowserSourcePathResolver` and then inject it here

https://github.com/microsoft/vscode-js-debug/blob/9b24a1420f7ce6b4a587e2386de3646ab4834101/src/targets/sourcePathResolverFactory.ts#L26

To avoid adding a ton of blazor special-casing into the generic browser path resolver

digeff: Done

src/adapter/breakpoints/breakpointBase.ts
digeff: Deleted

src/adapter/threads.ts
digeff: Removed

connor4312 (approved):
Let me know when you're ready for this to be merged

digeff:
@connor4312 I just added the unit tests, so it's ready to merge. Thanks! :)

PR #810, opened 2020-10-21
Enable loading source-maps from storage

digeff:
Now if we fail to load the source-map from the url, we'll try to translate the url to an absolute path and load it from there.


connor4312:
If storage is the right place to load sourcemaps from, can the scripts just be built with file URIs for their sourceRoots?

digeff:
I agree that would be the ideal scenario for us.

v1 supported this scenario, so we want to support this for backwards compatibility.
I can add telemetry to this code-path, and if no-one is using this we can remove it later in a few versions. What do you think?

connor4312:
js-debug has been the default on since June and I haven't received any issue reports around this, so I think it's unlikely this is used.

digeff:
You are correct, this is an issue on the tooling/building side.

In Visual Studio we try to make things work without the user needing to configure things when possible; my concern is that even thought this is a tooling/building side issue, some users might be relying on this behavior, and they might have a bad customer experience when we update the debugger, because their breakpoints don't work any more on a project that used to work on the previous VS version.

That's the reason I want to enable this scenario temporarily at least, until we can confirm that users in VS are not using this behavior either.

src/adapter/sources.ts
```diff
@@ -902,13 +902,25 @@ export class SourceContainer {
...
+    } catch (urlError) {
+      try {
+        const sourceMapAbsolutePath = await this.sourcePathResolver.urlToAbsolutePath({
+          url: source.sourceMap.metadata.sourceMapUrl,
+        });
+        if (sourceMapAbsolutePath) {
+          source.sourceMap.metadata.sourceMapUrl = sourceMapAbsolutePath;
+        }
+        sourceMap.map = await this.sourceMapFactory.load(source.sourceMap.metadata);
+      } catch (ignoredPathError) {}
```
connor4312, line 914:
```suggestion
      } catch {}
```

src/adapter/sources.ts
```diff
@@ -902,13 +902,25 @@ export class SourceContainer {
...
-        output: sourceMapParseFailed(source.url, e.message).error.format + '\n',
-        category: 'stderr',
-      });
+    } catch (urlError) {
+      try {
+        const sourceMapAbsolutePath = await this.sourcePathResolver.urlToAbsolutePath({
+          url: source.sourceMap.metadata.sourceMapUrl,
+        });
+        if (sourceMapAbsolutePath) {
+          source.sourceMap.metadata.sourceMapUrl = sourceMapAbsolutePath;
```
connor4312, line 911: Should this be a file URI for the `sourceMapUrl`?

src/adapter/sources.ts
```diff
@@ -902,13 +902,25 @@ export class SourceContainer {
 
     try {
       sourceMap.map = await this.sourceMapFactory.load(source.sourceMap.metadata);
-    } catch (e) {
-      this._dap.output({
-        output: sourceMapParseFailed(source.url, e.message).error.format + '\n',
-        category: 'stderr',
-      });
+    } catch (urlError) {
+      try {
```
connor4312, line 906: Can you add a comment to this block to say what's going on? Also wonder whether this is a code path we could enable only for VS.

src/adapter/sources.ts
digeff: Added call to utils.absolutePathToFileUrl to make it a file URI

src/adapter/sources.ts
digeff: Done

PR #817, opened 2020-10-26
Make Blazor source-path-resolver tests work in Mac and Linux

digeff:
I made the Local scenario test work.

Given that we don't support remote scenarios in Mac and Linux at the moment, I'm just skipping that test.


connor4312:
Thanks!

PR #820, opened 2020-10-28
Supports vscode-webview-resource: source maps

eamodio:
Adds source-map support for `vscode-webview-resource://` urls from vscode webviews

### DIFF
--- a/src/common/urlUtils.ts
+++ b/src/common/urlUtils.ts
@@ -9,10 +9,10 @@ import Cdp from '../cdp/api';
 import { AnyChromiumConfiguration } from '../configuration';
 import { BrowserTargetType } from '../targets/browser/browserTargets';
 import { MapUsingProjection } from './datastructure/mapUsingProjection';
+import { IFsUtils } from './fsUtils';
 import { memoize } from './objUtils';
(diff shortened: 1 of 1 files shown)


connor4312:
From slack discussion -- webview resource uris take the form

```
vscode-webview-resource://ac8c06dc-010a-4ea5-bbda-43d6f25dd21c/file///c%3A/Users/Eric/code/eamodio/vscode-gitlens/dist/webviews/rebase.js.map
```

That uri has `file///` instead of `file:///`. I think this might work by going through `got` which can normalize it, but this is more incidental than intentional -- I'd prefer we fix up the protocol so this doesn't break in the future.

This is the code in VS Code which handles this: https://github.com/microsoft/vscode/blob/ee0aff2460957b825fb4ce2edc3a6d21c90c2c04/src/vs/platform/webview/common/resourceLoader.ts#L104

you can use the standard URL instead of vscode URIs in this case. I would keep the cheap `candidate.startsWith('vscode-webview-resource://')` check (rather than checking `new URL(url).protocol`, since passing data URIs in `new URL()` has caused performance issues.

Issue #823, opened 2020-10-29
Add support for debugging Node internals when they are externally loaded


mmomtchev:
The JS part of the Node internals is usually included in the executable file at linking.
In order to avoid rebuilding Node at each small change of an internal JS, the Node build system allows building an executable without internals and then loading them at each execution:
`./configure --node-builtin-modules-path $(pwd)`
In this case, the internals will be loaded as an (almost) normal JS script and, when executing it, the Node debugger will return the filename as a special `node:` URL.

This small patch adds special handling for those URLs so that they can be mapped to the actual files. This allows for convenient setting of breakpoints. It is a very small change that will impact only a few people, but it will still mean a lot to them :smiley:

connor4312:
I didn't know that was possible, thanks for the PR!

It will be a couple days before I review+merge it since I want to try this locally to have a better understanding of behavior (we're busy with endgame this week at vs code).

mmomtchev:
This is the launch.json I use:
```js
{
	"name": "Launch current file",
	"runtimeExecutable": "${workspaceFolder}/out/Release/node",
	"args": [
		"--expose-internals",
		"--nolazy",
		"${workspaceFolder}/${relativeFile}",
	],
	"request": "launch",
	"type": "pwa-node"
}
```

And you will need to build node with
```
./configure --node-builtin-modules-path $(pwd)
```

PR #824, opened 2020-10-29
add support for debugging external node internals

mmomtchev:
It is a crude change, so I am open to any suggestions


mmomtchev:
#823

connor4312:
This change is very specific because it requires the node.js repo to be opened as the workspace folder (depends on internals living in the "lib" subdirectory). But debugging internals is a very uncommon case so I don't want to add another option just for this.

PR #826, opened 2020-10-29
Add new parameter for inspectUri needed to move BlazorProxy out of the project system

digeff:
We need to add a new parameter for the inspectUri to move the BlazorProxy launch out of the project system


connor4312:
Can you also add this parameter to the documentation in `strings.ts`?

connor4312 (approved):

PR #897, opened 2020-12-16
Update documentation for cwd

jasonwilliams:
Update documentation after changes made in https://github.com/microsoft/vscode-js-debug/commit/158eabf72a9d21d050a23969cf07cbe989f6cd80


OPTIONS.md
```diff
@@ -9,8 +9,8 @@
...
 <h5>Default value:</h4><pre><code>undefined</pre></code><h4>customPropertiesGenerator</h4><p>Customize the properties shown for an object in the debugger (local variables, etc...). Samples:<br>    1. { ...this, extraProperty: &#39;12345&#39; } // Add an extraProperty 12345 to all objects<br>    2. this.customProperties ? this.customProperties() : this // Use customProperties method if available, if not use the properties in this (the default properties)<br>    3. function () { return this.customProperties ? this.customProperties() : this } // Use customDescription method if available, if not return the default properties<br><br>    Deprecated: This is a temporary implementation of this feature until we have time to implement it in the way described here: <a href="https://github.com/microsoft/vscode/issues/102181">https://github.com/microsoft/vscode/issues/102181</a></p>
-<h5>Default value:</h4><pre><code>undefined</pre></code><h4>cwd</h4><p>Absolute path to the working directory of the program being debugged.</p>
```
connor4312, line 12: This file is autogenerated from `strings.ts` -- can you update the doc there?

jasonwilliams:
Sure np

Also while I was looking at the documentation i noticed `outFiles` uses `${workspaceFolder}/**/**`.
Is it possible to do something similar to `cwd` and do `localRoot/**/**` if set?

I ask this because those working in a monorepo probably won't know to set `outFiles` to be scoped down and by default (according to the docs) it will search every folder within the workspace.

connor4312:
There's a possibility for regression there -- I could set my roots to point to `${workspaceFolder}/src`, which would work correctly with the source paths, but the sourcemaps can be outside of that.

OPTIONS.md
jasonwilliams: Can you point me to this file? I don't see it

jasonwilliams: The `strings.ts` file doesn't have `node.attach.cwd` so i don't know where that bit is coming from?
i only see `node.launch.cwd` and `browser.launch.cwd`

connor4312: strings.ts is consumed in generate-contributions.ts, which is what the docs are actually created from. node.launch.cwd is shared in there between launch and attach -- the string naming is sort of a mess, in part because it contains strings copied from the previous debug adapters which were not consistent.

jasonwilliams:
updated

connor4312:
Thanks!

PR #924, opened 2021-02-17
fix: don't autoattach on yarn commands

yonifra:
I'm basing my PR on this fix by @connor4312 https://github.com/microsoft/vscode/issues/106991


connor4312:
Where are the yarn commands installed in this case?

PR #962, opened 2021-04-05
Remote 'New' from terminal label

Tyriar:
We want to avoid using "New"/"Create" so we control how they're displayed. When tabs come in we'll be dropping the prefix all together.

### DIFF
--- a/src/ui/configuration/nodeDebugConfigurationProvider.ts
+++ b/src/ui/configuration/nodeDebugConfigurationProvider.ts
@@ -14,7 +14,7 @@ import {
   AnyTerminalConfiguration,
   breakpointLanguages,
   ResolvingNodeConfiguration,
-  ResolvingTerminalConfiguration
+  ResolvingTerminalConfiguration,
 } from '../../configuration';
 import { findScripts } from '../debugNpmScript';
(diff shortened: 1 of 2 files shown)


src/ui/configuration/nodeDebugConfigurationProvider.ts
```diff
@@ -14,7 +14,7 @@ import {
   AnyTerminalConfiguration,
   breakpointLanguages,
   ResolvingNodeConfiguration,
-  ResolvingTerminalConfiguration,
+  ResolvingTerminalConfiguration
```
connor4312, line 17: removing this caused the build to fail (prettier formatting is enforced in ci)

connor4312 (approved):

src/ui/configuration/nodeDebugConfigurationProvider.ts
Tyriar: Wonder why that happened 😕

connor4312: I have `formatOnSave` on with the `defaultFormatter` set to prettier. I think if you don't have the prettier extension installed, it'll use the TS language features which formats different

PR #964, opened 2021-04-06
CDP Proxy: Allows Other Extensions to Reuse CDP Connection

### LINKED ISSUE #893: Expose (some) CDP commands
I'd love to play around with the [`Page.captureScreenshot` request](https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-captureScreenshot) for an experimental VS Code extension.
It would be awesome if the VS Code Debug Adapter would expose some request that would allow to call these CDB requests through `debugSession.customRequest`.

## Advantages

1) This would allow for some cool (experimental) VS Code extensions. The captureScreenshot request could be used to show a screenshot of variables that point to an HTML element. I guess there are also some other cool methods exposed by chrome that could further push the debugging experience.

## Risks

1) User experience could be degraded if such extensions bring the debug adapter in an inconsistent state by calling some CDP requests they are not supposed to call.

I guess to mitigate 1), the exposed CDP requests could be allow-listed (e.g. only when using the non-insiders build to enable exploring ideas). `Page.captureScreenshot` should be pretty risk free.

What do you think?

swissmanu:
Extends `js-debug` to expose its internal CDP connection to other extensions.

This pull request resolves #893.

## Overview

- A new command `requestCDPProxy` returns websocket connection details to connect to the CDP proxy.
  - If `requestCDPProxy` is called the first time, the proxy is created
  - Otherwise connection details for the previously created proxy are returned
- The debug adapter holds an instance of the CDP proxy and ensures forwarding of any communication between proxy clients and the CDP connection.
- The proxy and its clients use a basic JSON communication protocol described here: https://github.com/swissmanu/vscode-js-debug-cdp-proxy-api

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13553,6 +13553,11 @@
         "execa": "^4.0.0"
       }
     },
+    "vscode-js-debug-cdp-proxy-api": {
+      "version": "0.0.3",
+      "resolved": "https://registry.npmjs.org/vscode-js-debug-cdp-proxy-api/-/vscode-js-debug-cdp-proxy-api-0.0.3.tgz",
+      "integrity": "sha512-Ns6bNXBRfItOSyUnt+y9eZSl3C0TN9YNBV8np1k0p+PQ15H4DtGSq9yT3Wr+NR3S4RdFL16GjKdy/uqw12FQCQ=="
+    },
     "vscode-nls": {
       "version": "4.1.2",
       "resolved": "https://registry.npmjs.org/vscode-nls/-/vscode-nls-4.1.2.tgz",
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
     "source-map-support": "^0.5.19",
     "split2": "^3.1.1",
     "vscode-js-debug-browsers": "^1.0.4",
+    "vscode-js-debug-cdp-proxy-api": "0.0.3",
     "vscode-nls": "^4.1.2",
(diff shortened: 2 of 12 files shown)


swissmanu:
Would it make sense to expose the underlying error here https://github.com/microsoft/vscode-js-debug/blob/main/src/cdp/connection.ts#L271, instead of masking it with `undefined`? This way, users of the CDP proxy could get a concise error message as well.

connor4312:
Yea, the error should be exposed. I'll play around with it. I would _prefer_ to throw/reject with errors, but we depend on omitting the error in so many places throughout js-debug that changing this would be quite risky.

connor4312:
@swissmanu can you give me permission to push to your fork please? 🙂

connor4312:
In the meantime I've pushed my changes onto a branch in https://github.com/microsoft/vscode-js-debug/commit/0c6bd3f6e9c443083cd9c3728d76ec849b87a362

swissmanu:
> @swissmanu can you give me permission to push to your fork please? 🙂

I ticked the "Allow edits by maintainers" checkbox on this pull request only. Permissions to the fork should be work by now. Sorry for the delay.

> I moved the protocol to CDP with a js-debug extension.

I like like your removal of the additonal protocol layer. Makes things straight forward.

connor4312:
🚀 Thank you for your work on this!

PR #976, opened 2021-04-22
adopt virtual workspaces

sandy081:
This extension will be disabled in virtual workspace (remote hub)


package.json
```diff
@@ -170,6 +170,9 @@
   },
   "main": "./src/extension.js",
   "enableProposedApi": true,
+  "capabilities": {
+    "virtualWorkspaces": false
+  }
```
connor4312, line 175:
```suggestion
  },
```

PR #979, opened 2021-04-23
String support chakracore

millerds:
Some types of Office Add-ins use ChakraCore to run their java script which seem to have a problem with vscode evaluating strings from variables in the code.  This stems for the fact that the vscode-js-debug plugin is based on V8 which uses a "value" property in the object returned in their protocol where ChakraCore uses "description".  This change allows for either property to be used with "value" taking precedent over "description".  The allows us to use vscode for this type of office add-in development.

Note the change also includes a test case for this using ChakraCore assemblies.


src/test/node/node-runtime.test.ts
```diff
@@ -110,6 +110,33 @@ describe('node runtime', () => {
     handle.assertLog({ substring: true });
   });
 
+  itIntegrates('chakracore string value', async ({ r }) => {
+    createFileTree(testFixturesDir, {
```
connor4312, line 114: You'll want to bail out here if `process.platform !== 'win32'`

millerds: Thanks.

PR #991, opened 2021-05-06
990 - Support using launch configuration name as terminal title

### LINKED ISSUE #990: RFC - Support using launch task name as terminal title
When launching multiple Node debug instances, multiple integrated terminal sessions are spawned, but they all use the same default Node Debug Console title, and they are not always spawned in the same order if you use the compound tasks.  This makes it difficult to determine which of the terminals is the one you are looking for.

image

Use the launch task name as the terminal title.  This would provide a much better experience for developers and the name field is already required for all launch tasks.
image

danielgary:
Closes #990

### DIFF
--- a/src/targets/node/terminalProgramLauncher.ts
+++ b/src/targets/node/terminalProgramLauncher.ts
@@ -3,7 +3,6 @@
  *--------------------------------------------------------*/
 
 import { inject, injectable } from 'inversify';
-import * as nls from 'vscode-nls';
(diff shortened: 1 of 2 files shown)


connor4312:
This needs to be updated too: https://github.com/microsoft/vscode-js-debug/blob/b6242ab685b2f1473026efbe533da522c580d354/src/test/node/node-runtime.test.ts#L210

Issue #871, opened 2020-11-30
Allow multiple paths in pathMapping


mrgleba:
**Is your feature request related to a problem? Please describe.**
Right now in ```pathMapping``` one con provide only one path.
In our case sometimes one path serves files from multiple locations (we're doing a piece by piece transition from old code to new)
and there is no way to link both directories to the debugger

**Describe the feature you'd like**
Ideally ```pathMapping``` (and ```webRoot```) should allow either a string or string array

PR #1002, opened 2021-05-23
feat: support out of band source map lookup

jchip:
Request for your consideration.  

Supports having a JSON file that specifies the source map URLs for source files.

This would enable tools to generate an out of band source map lookup file rather than relying on the `//# sourceMappingURL` tag in the code.

There are still some more work to do.  If this is something acceptable, I'd like to finish it up with your guidance.

Thank you

### DIFF
--- a/src/adapter/breakpointPredictor.ts
+++ b/src/adapter/breakpointPredictor.ts
@@ -63,6 +63,12 @@ export interface IBreakpointsPredictor {
    * Returns predicted breakpoint locations for the provided source.
    */
   predictedResolvedLocations(location: IWorkspaceLocation): IWorkspaceLocation[];
+
+  /**
+   * Returns source map url for a compiled path
+   * @param compiledPath
+   */
+  getSourceMapUrlForCompiled(compiledPath: string): string;
 }
 
 /**
@@ -151,6 +157,13 @@ export class BreakpointPredictorDelegate implements IBreakpointsPredictor, IDisp
   /**
    * @inheritdoc
    */
+  getSourceMapUrlForCompiled(compiledPath: string) {
+    return this.implementation.getSourceMapUrlForCompiled(compiledPath);
+  }
+
+  /**
+   * @inheritdoc
+   */
   dispose() {
     this.parent?.invalidateNextChild();
   }
@@ -161,6 +174,7 @@ export class BreakpointsPredictor implements IBreakpointsPredictor {
   private readonly predictedLocations = new Map<string, IWorkspaceLocation[]>();
   private readonly longParseEmitter = new EventEmitter<void>();
   private sourcePathToCompiled?: Promise<MetadataMap>;
+  private compiledSourceMapMeta: Record<string, Readonly<ISourceMapMetadata>>;
   private cache?: CorrelatedCache<number, { sourceUrl: string; resolvedPath: string }[]>;
 
   /**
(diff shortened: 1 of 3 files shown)


connor4312:
I'm not sure I understand the need for this. In this PR, it works with files on disk. It seems like it would be simpler to adjust the program that generates the files create them in the right location, or have some post-generation step that fixes the URLs in the files.

I could kind of see this more if it was dealing with potentially uncontrolled files when debugging in browsers, but that's not something we've had requested yet.

jchip:
Hi, the primary use is for a monorepo tool my team developed.  We create hard links in `node_modules` back to ordinary `.js` files in local packages, and use source map to coerce the debugger to load the original `.js` file, but because of the hard link, the original `.js` files get modified with the source map tag also.  The hard link is used so once user modify their original `.js` files, the linked copy gets the change automatically.

Here is the doc about this https://www.electrode.io/fynpo/docs/getting-started/debugging/

connor4312:
The usual way to do this is to use symbolic links which Node will dereference to their source locations automatically (this behavior can be controlled via the `--preserve-symlinks`).

jchip:
We've explored the preserve symlinks option extensively and it doesn't solve the typical issues associated with monorepo using symlinks, like the ones yarn listed.

connor4312:
I think the feature request could be better served by an enhanced pathMapping for Node.js: https://github.com/microsoft/vscode-js-debug/issues/871

In that case you could, in the launch.json, do something like (if the editor was opened in a subpackage):

```
"pathMapping": {
  "${workspaceFolder}/node_modules/foo/*": "${workspaceFolder}/../foo"
}
```

jchip:
thanks.  is `pathMappings` available?  I tried in in the nightly build but the option is not recognized.

connor4312:
It's not exposed for Node.js right now. We just 'fake it' using the cwd when it's passed into common resolution functions, e.g.

https://github.com/microsoft/vscode-js-debug/blob/14323d3edd2c9e8097494e440bf33e3c16a5f2df/src/targets/node/nodeSourcePathResolver.ts#L170

jchip:
Yes, this would be a better solution.  Do you have a timeline on actually implementing it?  I might take a stab at it this weekend.

Would you consider having a similar out of band config file like `.path-mappings.json` under a directory to individually configure this for node.js?  This would enable our mono-repo solution to "just work" for our users with vscode and they don't have to configure `launch.json`.

connor4312:
>  Do you have a timeline on actually implementing it? I might take a stab at it this weekend.

I don't have a timeline for this; this is about the end of our iteration so it is unlikely to happen until next month at the easlier. PR's welcome!

> Would you consider having a similar out of band config file like .path-mappings.json under a directory to individually configure this for node.js? This would enable our mono-repo solution to "just work" for our users with vscode and they don't have to configure launch.json.

I would prefer not to do this. Some other ideas that might work for you, let me know:

- If .path-mappings.json is autogenerated, you could have tooling auto-modify the launch.json correspondingly. ([This](https://www.npmjs.com/package/jsonc-parser) is the comment and fault-tolerant json library vscode uses)
- Writing a small vscode extension that handled launching for you. Could be useful to provide other utilities for users of your monorepo, too. (As someone who built a vaguely similar monorepo tool before working on VS Code, I wish I had done this 🙂) [Guide here](https://code.visualstudio.com/api/get-started/your-first-extension). You could invoke the `vscode.debug.startDebugging` API to programmatically launch a debug session.

jchip:
Thank you.  I will look into it this weekend.  I may just do it for node.js.  I imagine it may be quite a bit more complex when having to deal with all use cases like the browsers.

Definitely 👍  on creating an extension to make working with mono-repo more convenient.  Thanks for the ideas.

By the way, I enjoyed reading your post about your interview with the vscode team.  That was a cool experience.

Issue #1005, opened 2021-05-24
Feature Request: Allow specifying the address when binding to a port


lhchavez:
When opening a Node.js debugging session in my local machine, I'd prefer to be able to have the TCP port that the debug server will bind to in the loopback interface instead of in all of them, mostly out of paranoia when I'm working on a shared WiFi.

Also (and maybe less interesting to other folks), at replit.com, we detect TCP ports bound to `0.0.0.0` and treat them magically as [web servers](https://docs.replit.com/repls/http-servers), so this should also prevent opening a debug session causing an unintended web server from being "deployed".

So I would like for it to be possible to invoke e.g. `node debugServerMain.js localhost:12345` and have that session only bind to localhost.

JacksonKearl:
Is this functionality accessible through vscode UI at all or an is this a CLI-only feature? 

Verified on CLI, though I might recommend having the log show the full address: 
![image](https://user-images.githubusercontent.com/8586769/120756262-e4ba1800-c4c3-11eb-81b6-e908936c734b.png)

connor4312:
Yea this is CLI-only. Thanks for verifying!

PR #1006, opened 2021-05-24
feat: Allow callers of debugServerMain.ts to specify the host

lhchavez:
This change allows callers of debugServerMain.ts to be able to provide
what address the DAP listener will bind to. This prevents random
eavesdroppers from listening into the DAP.

Fixes: #1005


connor4312 (approved):

connor4312:
Thanks!

Issue #1009, opened 2021-05-25
Stopped events are not delivered when launching with `pwa-node`


lhchavez:
**Describe the bug**
When launching a `pwa-node` app through the terminal, the DAP does not deliver any Stopped events.

**To Reproduce**
1. With the following DAP messages: (full repro here: https://replit.com/@luisreplit/NodeDAP#driver.py)
```json
{
  "command": "initialize",
  "arguments": {
    "clientID": "replit",
    "clientName": "replit.com",
    "pathFormat": "path",
    "linesStartAt1": True,
    "columnsStartAt1": True,
    "supportsVariableType": True,
    "supportsVariablePaging": True,
    "supportsRunInTerminalRequest": True,
    "locale": "en-us",
    "supportsProgressReporting": True,
    "supportsInvalidatedEvent": True
  }
}
{
  "command": "setBreakpoints",
  "arguments": {
    "source": {
      "name": "main.js",
      "path": "/home/runner/NodeDAP/main.js"
    },
    "breakpoints": [{
      "line": 4
    }],
    "sourceModified": False
  }
}
{
  "command": "launch",
  "arguments": {
    "type": "pwa-node",
    "request": "launch",
    "program": "/home/runner/NodeDAP/main.js",
    "args": [],
    "stopOnEntry": False,
    "pauseForSourceMap": False,
    "sourceMaps": True,
    "smartStep": True,
    "cwd": "/home/runner/NodeDAP",
    "environment": [],
    "console": "externalTerminal",
    "trace": {
      "level": "verbose",
      "logFile": "/home/runner/NodeDAP/adapter.log"
    }
  },
}
{
  "command": "configurationDone"
}
```
3. There are no Stopped events when the breakpoint is reached, but there are `Debugger.paused` events!

**Log File**
https://gist.githubusercontent.com/lhchavez/e6d3f50c213d53c752e285131240774d/raw/26211f7e66b0943ebec1805d1b382a9aa79d0e1d/nodedap.txt

**VS Code Version:** 1.56.2

**Additional context**
I've been able to reproduce this in VSCode, but not reliably. Not sure what the difference is.

connor4312:
@lhchavez were you able to verify that this is fixed for you?

lhchavez:
yes, thank you so much for merging :D

PR #1010, opened 2021-05-25
Ensure that the debugAdapter's _onInitialize() is called

lhchavez:
When launching a `pwa-node` process, the debugAdapter's _onInitialize()
method is never called, since the dap registers a listener for the
`'initialize'` message _after_ it has arrived. This causes the Thread's
`dap` to never be resolved, so any thread events (e.g. `'paused'`) are
never sent to the client! This in turn causes VSCode to never realize
that the process is stopped, so the debuggee just waits there,
patiently, forever.

This change stores the `Dap.InitializeParams` on the Binder and then
when the Thread is created, the params are forwarded so that the
DebugAdapter can _onInitialize() correctly and the thread-related events
can flow.

Fixes: #1009

### DIFF
--- a/src/adapter/debugAdapter.ts
+++ b/src/adapter/debugAdapter.ts
@@ -303,7 +303,11 @@ export class DebugAdapter implements IDisposable {
     return errors.createSilentError(localize('error.threadNotFound', 'Thread not found'));
   }
 
-  createThread(cdp: Cdp.Api, delegate: IThreadDelegate): Thread {
+  createThread(
+    cdp: Cdp.Api,
+    delegate: IThreadDelegate,
+    initializeParams?: Dap.InitializeParams,
+  ): Thread {
     this._thread = new Thread(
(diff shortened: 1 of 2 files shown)


connor4312 (approved):

connor4312:
Thanks!

connor4312:
@lhchavez I had to revert this since it caused other issues. And I should have looked more closely; what's happening is not really correct. It seems this is reusing the root session's initialized params to trigger the child, but really the child should get its own initialized event. This is what happens in VS Code, and it seems like `_onInitialized` is hit in all cases.

If you think something is still being done wrong in js-debug, please explain a little more and share a repro if possible. I saw your link in the issue to https://replit.com/@luisreplit/NodeDAP#driver.py but I'm not really sure how to use that. Thanks!

lhchavez:
awesome, i'll try to have a more self-contained reproduction case.

Issue #1028, opened 2021-06-11
Error when trying to set a breakpoint in index.html


rinselmann:
**Describe the bug**
When I try to set a breakpoint in a project that contains only an index.html file served from the filesystem, I get the following error: `Error processing setBreakpoints: SyntaxError: Invalid regular expression`

**To Reproduce**
1. Create  new vscode project containing an index.html file with a script block
2. Create a launch configuration to open the index.html file using `path`
3. Start a debugging session
4. Try to set a breakpoint in index.html
5. See Error

**VS Code Version:** 1.56.2

**Additional context**
 I believe this is due to urlToRegex incorrectly handling paths that are partially pre-escaped, which happens if the path ends with index.html. I have a fix that I'm planning on submitting a pull request for.

PR #1029, opened 2021-06-11
fix: Error when trying to set a breakpoint in index.html

rinselmann:
This is a fix for the urlToRegex issue referenced in #1028 along with 2 new test cases.

I'm making the assumption that any part of `aPath` that has already been escaped should also not have the url decode/encode step done, which seems to match the previous intended behavior. I am also making the assumption that there is no expectation at that point that the portion of the path that needs to be escaped will be a valid file url, especially if escapeReStart is not zero. This seems like a safe assumption given that the escape param only seems to be used from one place that works under those assumptions, but the behavior isn't super intuitive. I didn't want to change too much though, especially since I'm new to this codebase and don't know the history of that method.


rinselmann:
It looks like some tests are failing, so I've close my pull request and will make another when the tests pass

rinselmann:
Tests are passing now. Two of the tests that were failing incorrectly due to bugs in the tests that were hidden by the urlToReq bug, so I fixed those. One of the tests was failing because I accidentally broken the case insensitive portion of the conversion, so that should be fixed now, too.

connor4312:
Good find!

PR #1032, opened 2021-06-16
feat: disable pause if all breakpoints have been disabled

rinselmann:
The current behavior is that the debugger is configured to pause while loading source maps if at least one breakpoint has been set during a debugging session. This has the side effect of switching to the sources tab on page reload even if all breakpoints have since been disabled. This commit removes the automatic pausing of the debugger while loading source maps if there are no longer any active breakpoints. Pausing while source maps are loading should resume if any breakpoints are reenabled or new breakpoints are added. This allows developers keep the benefit of more reliable breakpoints early in program without the drawback of extra pauses when they are not needed.

### DIFF
--- a/src/adapter/breakpoints.ts
+++ b/src/adapter/breakpoints.ts
@@ -110,7 +110,7 @@ export class BreakpointManager {
     return this._byPath;
   }
 
-  private _sourceMapHandlerWasUpdated = false;
+  private _sourceMapHandlerInstalled = false;
 
   /**
    * User-defined breakpoints by `sourceReference`.
@@ -425,7 +425,7 @@ export class BreakpointManager {
     }
 
     if (this._byPath.size > 0 || this._byRef.size > 0) {
-      this._updateSourceMapHandler(this._thread);
+      this._installSourceMapHandler(this._thread);
     }
   }
 
@@ -465,8 +465,8 @@ export class BreakpointManager {
     this._predictorDisabledForTest = disabled;
   }
 
-  private async _updateSourceMapHandler(thread: Thread) {
-    this._sourceMapHandlerWasUpdated = true;
+  private async _installSourceMapHandler(thread: Thread) {
+    this._sourceMapHandlerInstalled = true;
     const perScriptSm =
       (this.launchConfig as IChromiumBaseConfiguration).perScriptSourcemaps === 'yes';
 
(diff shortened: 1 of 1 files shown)


rinselmann:
I submitted a pull request the other day to fix a small issue in the url to regex conversion, but this was really the issue I was trying to track down the source of. Thanks!

src/adapter/breakpoints.ts
```diff
@@ -494,7 +494,7 @@ export class BreakpointManager {
     params: Dap.SetBreakpointsParams,
     ids: number[],
   ): Promise<Dap.SetBreakpointsResult> {
-    if (!this._sourceMapHandlerWasUpdated && this._thread) {
+    if (!this._sourceMapHandlerWasUpdated && this._thread && params.breakpoints) {
```
connor4312, line 497:
```suggestion
    if (!this._sourceMapHandlerWasUpdated && this._thread && params.breakpoints?.length) {
```

src/adapter/breakpoints.ts
```diff
@@ -585,6 +585,11 @@ export class BreakpointManager {
 
     this._totalBreakpointsCount += result.new.length;
 
+    if (this._thread && this._totalBreakpointsCount === 0 && this._sourceMapHandlerWasUpdated) {
+      this._thread.setScriptSourceMapHandler(false);
+      this._sourceMapHandlerWasUpdated = false;
```
connor4312, line 590: Let's rename `_sourceMapHandlerWasUpdated` to `_sourceMapHandlerInstalled` to better reflect what it's doing.

rinselmann: should _updateSourceMapHandler also be renamed to _installSourceMapHandler? It may also make sense to have an _uninstallSourceMapHandler, but I had been trying to make my original change as minimal as possible. I didn't really like that I was manually messing with a flag that seems to be owned by the update method, but it was awkward to try to make the update method "figure out" what do do in most cases, but then guarantee that it will uninstall the source map handler in the case I added.

connor4312: I'm good with _installSourceMapHandler/_uninstallSourceMapHandler

rinselmann: I updated the pull request with those changes

connor4312 (approved):
Thanks for the PR!

PR #1044, opened 2021-06-27
Node 16.x breakpoint warning resolved in 16.4

Ashesh3:
Since the said issue is resolved and shipped in 16.4, Suggesting users to downgrade is a bad idea, instead they should be recommended to upgrade to the latest release to resolve this issue.

Reference: https://github.com/nodejs/node/pull/38273


connor4312 (approved):

connor4312:
Excellent, thanks!

PR #1146, opened 2021-11-30
updating webpack version

Shenniey:
Updating webpack version according to [this thread](https://github.com/webpack/webpack/issues/14532), as Visual Studio is currently building using Node v17.x.


connor4312:
Updated in main

PR #1163, opened 2022-01-01
fix: DebugSessionTunnels. Misplaced remote/local port usage

### LINKED ISSUE #1162: Remote development(debug). Incorrect port gets forwarded to UI. Applicable for pwa-chrome and similar launch modes
**Describe the bug**
When using remote workspace with `pwa-chrome` launch mode incorrect dynamic port gets forwarded to UI (remote port instead of local port).
Thus js-debug-companion uses aforementioned port which results in broken debug.
Problem is only reproducible when ports (local/remote) are different.

It seems like this problem has been described before in #982. It has been closed but problem seems to persist.

**To Reproduce**
1. on your local machine run following script (it occupies set port range hence local vscode is forced to use ports with which it runs into the problem).
```javascript
const net = require("net");
function *range(from, to) {
    for (let i = from; i < to; i++) yield i;
}

function runServer(port) {
   const server = new net.Server();
   server.listen(port, "127.0.0.1");
   return server;
}

// adjust port range to your environment. (use ports you will see in vscode port forwarding view from step 2)
const servers = [...range(51000, 59999)].map((port) => runServer(port));

// servers.map((server) => server.close());
```
2. in vscode command prompt run 'Debug: Open Link' and enter some link (for example: https://code.visualstudio.com/)
3. observe debugger does not work

**Log File**

**VS Code Version:** 1.63.2

**Additional context**
applicable for remote development 
my env:
- remote on ubuntu 18.04 (hyper-v VM)
- local on windows 10

zkx5xkt:
- fix #1162 
- changed remote/local address usages in `DebugSessionTunnels` class to conform to vscode api standard as used in `vscode.workspace.openTunnel` api method

### DIFF
--- a/src/ui/companionBrowserLaunch.ts
+++ b/src/ui/companionBrowserLaunch.ts
@@ -68,14 +68,14 @@ const launchCompanionBrowser = async (
       tunnelRemoteServerIfNecessary(args),
       sessionTunnels
         .request(session.id, {
-          localPort: args.serverPort,
+          remotePort: args.serverPort,
           label: 'Browser Debug Tunnel',
         })
         .catch(() => undefined),
     ]);
 
     await vscode.commands.executeCommand('js-debug-companion.launchAndAttach', {
-      proxyUri: tunnel ? `127.0.0.1:${tunnel.remoteAddress.port}` : `127.0.0.1:${args.serverPort}`,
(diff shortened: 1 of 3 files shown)


connor4312:
Thanks!

PR #1186, opened 2022-01-22
Update bootloader so that `$jsDebugIsRegistered` not enumerable on global

gioragutt:
While working on a `node` PR, I encountered the following error when running tests:

image

This is thrown by a cleanup phase in some of `node`'s testing setup:

```ts
function leakedGlobals() {
  const leaked = [];

  for (const val in global) {
    if (!knownGlobals.includes(global[val])) {
      leaked.push(val);
    }
  }

  return leaked;
}
```

I can, technically, adjust `node`'s checks to be able to add `"$jsDebugIsRegistered"` to the allowlist, but I think it's better to change it here because:
1. It helps keep editor-specific changes out of `node`.
2. If the token string were to change in this library, `node` would have to be changed again.

Overall, the purpose of this PR is to make it a tiny bit easier to work on `node` using `vscode`.

### DIFF
--- a/src/targets/node/bootloader.ts
+++ b/src/targets/node/bootloader.ts
@@ -42,7 +42,12 @@ const jsDebugRegisteredToken = '$jsDebugIsRegistered';
       args: process.argv,
     });
 
-    Object.assign(global, { [jsDebugRegisteredToken]: true });
+    // not enumerable to not get picked up by node's testing globals leakage checks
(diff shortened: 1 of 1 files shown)


benjamingr:
This is something that is helpful whenever running tests that check no globals leak by enumerating the global, I've ran into this outside of node core development as well :)

Issue #1190, opened 2022-01-31
Debug Blazor using Edge on Ubuntu


thaystg:
**Describe the bug**
Launch Edge on ubuntu and redirect to the page set in url parameter - using nightly build of vscode-js-debug.

**To Reproduce**
1. Create a new folder called `MyBlazorApp` somewhere
2. Navigate inside the folder and Open in Terminal
3. Type `dotnet new blazorwasm`
4. Type `code .; exit`
5. Allow VS Code to auto-generate `launch.json` and `tasks.json`
6. Inside `launch.json`, add `"browser": "edge"` to the first configuration in the list and set the url to the url defined in your `launchSettings.json`
6. Go to the file `Counter.razor` inside of the `Pages` folder and put a breakpoint where it says `currentCount++;`
7. Press `F5` to debug
8. It launches the browser but doesn't redirect to the url set in `launch.json`

Related to https://github.com/dotnet/aspnetcore/issues/38413

connor4312:
It appears that, in connecting to the Blazor proxy to create a websocket connection, it uses a 302 redirect to a different, `http://` url. As websocket connections do not work over the `http:` protocol, the connection then fails. @thaystg this looks like something to be fixed on the Blazor side of things

![image](https://user-images.githubusercontent.com/2230985/151888732-ee21f11b-009b-4e4a-81f5-15b34ef61e96.png)

This is where the error is caught, you can set a breakpoint here and debug up the stack to see more info 

https://github.com/microsoft/vscode-js-debug/blob/728b43331d3120edd4069c98395110895bafaf4e/src/cdp/webSocketTransport.ts#L44

thaystg:
I don't think it's on BrowserDebugProxy side, once it works correctly on Chrome.

connor4312:
It doesn't work on Chrome nor Edge for me

thaystg:
At least for me when I launch the BlazorApp using browser = edge on chrome, it shows a about:blank page on edge and never redirects to the correct url.

thaystg:
Is your launch.json like this:

```
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Launch and Debug Standalone Blazor WebAssembly App",
            "type": "blazorwasm",
            "request": "launch",
            "cwd": "${workspaceFolder}",
            "browser": "chrome",
            "url": "http://localhost:5133"
        }
    ]
}
```

connor4312:
[...]

PR #1195, opened 2022-02-07
Rewrite http redirects to be backward compatible with blazor redirects

lewing:
when a redirect occurs the WebSocket updates the url property so we read that in the error handler and replace http(s): with ws(s): and try one more time to connect

fixes [https://github.com/microsoft/vscode-js-debug/issues/1190](https://github.com/microsoft/vscode-js-debug/issues/1190) here to support previous releases of the Blazor debugger

### DIFF
--- a/src/cdp/webSocketTransport.ts
+++ b/src/cdp/webSocketTransport.ts
@@ -30,18 +30,31 @@ export class WebSocketTransport implements ITransport {
 
     while (true) {
       try {
-        const ws = new WebSocket(url, [], {
+        const options = {
           headers: { host: 'localhost' },
           perMessageDeflate: false,
           maxPayload: 256 * 1024 * 1024, // 256Mb
           rejectUnauthorized: !(isSecure && targetAddressIsLoopback),
           followRedirects: true,
-        });
+        };
 
+        const ws = new WebSocket(url, [], options);
         return await timeoutPromise(
           new Promise<WebSocketTransport>((resolve, reject) => {
             ws.addEventListener('open', () => resolve(new WebSocketTransport(ws)));
-            ws.addEventListener('error', errorEvent => reject(errorEvent.error)); // Parameter is an ErrorEvent. See https://github.com/websockets/ws/blob/master/doc/ws.md#websocketonerror
+            ws.addEventListener('error', errorEvent => {
+              // Check for invalid http redirects for compatibility with old cdp proxies
+              const redirectedUrl = ws.url.replace(/^http:/, 'ws:').replace(/^https:/, 'wss:');
+
+              if (redirectedUrl === url) reject(errorEvent); // Parameter is an ErrorEvent. See https://github.com/websockets/ws/blob/master/doc/ws.md#websocketonerror
+
+              // Try to reconnect with the redirected URL
+              const redirectedSocket = new WebSocket(redirectedUrl, [], options);
+              redirectedSocket.addEventListener('open', () =>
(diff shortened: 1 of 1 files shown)


src/cdp/webSocketTransport.ts
```diff
@@ -30,18 +30,31 @@ export class WebSocketTransport implements ITransport {
...
           new Promise<WebSocketTransport>((resolve, reject) => {
             ws.addEventListener('open', () => resolve(new WebSocketTransport(ws)));
-            ws.addEventListener('error', errorEvent => reject(errorEvent.error)); // Parameter is an ErrorEvent. See https://github.com/websockets/ws/blob/master/doc/ws.md#websocketonerror
+            ws.addEventListener('error', errorEvent => {
+              // Check for invalid http redirects for compatibility with old cdp proxies
+              const redirectedUrl = ws.url.replace(/^http:/, 'ws:').replace(/^https:/, 'wss:');
+
+              if (redirectedUrl === url) reject(errorEvent); // Parameter is an ErrorEvent. See https://github.com/websockets/ws/blob/master/doc/ws.md#websocketonerror
+
+              // Try to reconnect with the redirected URL
```
mkArtakMSFT, line 51: Should all this section happen in the `else`  case?

src/cdp/webSocketTransport.ts
```diff
@@ -30,18 +30,31 @@ export class WebSocketTransport implements ITransport {
...
+              const redirectedSocket = new WebSocket(redirectedUrl, [], options);
+              redirectedSocket.addEventListener('open', () =>
+                resolve(new WebSocketTransport(redirectedSocket)),
+              );
+              redirectedSocket.addEventListener('error', errorEvent => reject(errorEvent.error));
```
mkArtakMSFT, line 56: Should all this section happen in the `else`  case?

src/cdp/webSocketTransport.ts
```diff
@@ -30,18 +30,34 @@ export class WebSocketTransport implements ITransport {
...
         return await timeoutPromise(
           new Promise<WebSocketTransport>((resolve, reject) => {
             ws.addEventListener('open', () => resolve(new WebSocketTransport(ws)));
-            ws.addEventListener('error', errorEvent => reject(errorEvent.error)); // Parameter is an ErrorEvent. See https://github.com/websockets/ws/blob/master/doc/ws.md#websocketonerror
+            ws.addEventListener('error', errorEvent => {
+              // Check for invalid http redirects for compatibility with old cdp proxies
+              const redirectedUrl = ws.url.replace(/^http:/, 'ws:').replace(/^https:/, 'wss:');
+
+              if (redirectedUrl === url) {
+                reject(errorEvent); // Parameter is an ErrorEvent. See https://github.com/websockets/ws/blob/master/doc/ws.md#websocketonerror
```
connor4312, line 50: Previously this was rejected with `errorEvent.error`, was the change intentional?

src/cdp/webSocketTransport.ts
```diff
@@ -30,18 +30,34 @@ export class WebSocketTransport implements ITransport {
...
+                return;
+              }
+
+              // Try to reconnect with the redirected URL
+              const redirectedSocket = new WebSocket(redirectedUrl, [], options);
```
connor4312, line 55: I suggest instead making a recursive call to the `create` method here so that timeout/cancellation is handled

connor4312 (changes requested):

connor4312 (approved):

PR #1206, opened 2022-02-15
Pass path mapping config to Edge DevTools opener command

bwalderman:
Fixes https://github.com/microsoft/vscode-edge-devtools/issues/787

Microsoft Edge DevTools extension users expect the `webRoot` and `pathMapping` properties in `launch.json` to apply to file links opened through the Edge DevTools. Today, the Edge DevTools extension only considers the `webRoot` and `pathMapping` properties from its own extension settings. This change plumbs the relevant launc config properties through to the `EdgeDevToolOpener` command so that they are available to the DevTools.

Related: https://github.com/microsoft/vscode-edge-devtools/pull/885

### DIFF
--- a/src/ui/edgeDevToolOpener.ts
+++ b/src/ui/edgeDevToolOpener.ts
@@ -24,6 +24,14 @@ const qualifies = (session: vscode.DebugSession) => {
 const toolExtensionId = 'ms-edgedevtools.vscode-edge-devtools';
 const commandId = 'vscode-edge-devtools.attachToCurrentDebugTarget';
 
+function findRootSession(session: vscode.DebugSession): vscode.DebugSession {
+  let root = session;
+  while (root.parentSession) {
+    root = root.parentSession;
+  }
+  return root;
+}
+
 @injectable()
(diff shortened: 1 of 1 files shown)


src/ui/edgeDevToolOpener.ts
```diff
@@ -48,7 +48,7 @@ export class EdgeDevToolOpener implements IExtensionContribution {
         }
 
         try {
-          return await vscode.commands.executeCommand(commandId, session.id);
+          return await vscode.commands.executeCommand(commandId, session.id, session.configuration);
```
connor4312, line 51: I would much rather pull the root session's configuration by iterating up `session.parentSession` than plumbing select properties down through individual sessions just for this command

bwalderman: Thanks! I've made the change. I wasn't sure reaching up to the parentSession would be acceptable in this context.

connor4312 (approved):

connor4312:
lgtm, please run `npm run fmt`, or install the Prettier extension and save the file, to format it so that CI passes.

PR #1216, opened 2022-03-01
Adding replay messages for console in Runtime domain

mliao95:
We are enabling the console in the [Edge DevTools for VS Code extension](https://github.com/microsoft/vscode-edge-devtools).  For this, we need to track two additional events for replay.

- `Runtime.executionContextCreated`
    - We need this to send the execution context to the console in order to evaluate script on the page.
- `Runtime.consoleAPICalled`
    - We need to replay console events back to our console when we connect such that the DevTools will retain the console history when attaching to an existing target.

Adding these CDP methods to replay will be helpful for anyone else using the JS-Debug proxy to stand their own console.

### DIFF
--- a/src/adapter/cdpProxy.ts
+++ b/src/adapter/cdpProxy.ts
@@ -167,6 +167,8 @@ export class CdpProxyProvider implements ICdpProxyProvider {
   ) {
     this.replay.capture(cdp, 'CSS', 'styleSheetAdded');
     this.replay.capture(cdp, 'Debugger', 'paused', true);
+    this.replay.capture(cdp, 'Runtime', 'executionContextCreated');
+    this.replay.capture(cdp, 'Runtime', 'consoleAPICalled');
     cdp.Debugger.on('resumed', () => {
       this.replay.clearEvent('Debugger', 'paused');
     });
--- a/src/adapter/cdpProxy.ts
+++ b/src/adapter/cdpProxy.ts
@@ -66,7 +66,13 @@ class DomainReplays {
   /**
    * Adds a message to be replayed.
    */
-  public addReplay(domain: keyof Cdp.Api, event: string, params: unknown, clearPrevious = false) {
+  public addReplay(
+    domain: keyof Cdp.Api,
+    event: string,
+    params: unknown,
+    clearPrevious = false,
+    capacity?: number,
+  ) {
     if (clearPrevious) {
       this.clearEvent(domain, event);
     }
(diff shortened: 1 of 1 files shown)


connor4312:
Should there be some limit on the replays here? Some applications can get very verbose with their console logs and we wouldn't want memory to grow infinitely over long debug sessions.

mliao95:
> Should there be some limit on the replays here? Some applications can get very verbose with their console logs and we wouldn't want memory to grow infinitely over long debug sessions.

In general, that would be a good idea. However, we would have to make some structural changes on how we store replays for the `Runtime` domain. In terms of console messages, I think it makes the most sense to go for a FIFO approach where we only store the most recent X console messages.  However, we want to persist any `executionContextCreated` calls. The ideal way to do this would be to separate the `Runtime` events in two separate arrays which would require touching the capture and reads from the replay class.

Any thoughts on this? Is there another solution here that I'm not seeing?

connor4312:
`capture` could keep a rolling count of the number of events it's seen, and take an optional `capacity` argument. If the capacity is exceeded, it could look through and splice out the first occurrence of the event from the `replays` list.

We could eventually be more efficient if the replays was a doubly linked linked list, but that would do the job.

mliao95:
> `capture` could keep a rolling count of the number of events it's seen, and take an optional `capacity` argument. If the capacity is exceeded, it could look through and splice out the first occurrence of the event from the `replays` list.
> 
> We could eventually be more efficient if the replays was a doubly linked linked list, but that would do the job.

I got a change working, but `gulp lint` is preventing me from committing.  It looks like this triggers whenever I try adding a new argument to the `capture` method. Here is the not so helpful error message:

```
[18:16:01] Using gulpfile D:\forks\vscode-js-debug\gulpfile.js
[18:16:01] Starting 'lint'...
[18:16:01] Starting 'lint:prettier'...
[18:16:01] Starting 'lint:eslint'...
adapter\cdpProxy.ts
× 1 files were not formatted
[18:16:06] 'lint:prettier' errored after 5.09 s
[18:16:06] Error: Prettier exited with code 1
    at formatError (C:\Users\milia\AppData\Roaming\npm\node_modules\gulp-cli\lib\versioned\^4.0.0\format-error.js:21:10)    at Gulp.<anonymous> (C:\Users\milia\AppData\Roaming\npm\node_modules\gulp-cli\lib\versioned\^4.0.0\log\events.js:33:15)
    at Gulp.emit (events.js:387:35)
    at Gulp.emit (domain.js:470:12)
    at Object.error (D:\forks\vscode-js-debug\node_modules\undertaker\lib\helpers\createExtensions.js:61:10)
    at handler (D:\forks\vscode-js-debug\node_modules\now-and-later\lib\map.js:50:14)
    at f (D:\forks\vscode-js-debug\node_modules\once\once.js:25:25)
    at f (D:\forks\vscode-js-debug\node_modules\once\once.js:25:25)
    at tryCatch (D:\forks\vscode-js-debug\node_modules\async-done\index.js:24:15)
    at done (D:\forks\vscode-js-debug\node_modules\async-done\index.js:40:12)
    at ChildProcess.<anonymous> (D:\forks\vscode-js-debug\gulpfile.js:451:36)
    at ChildProcess.emit (events.js:375:28)
    at ChildProcess.emit (domain.js:532:15)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:277:12)
    at Process.callbackTrampoline (internal/async_hooks.js:131:17)
[18:16:06] 'lint' errored after 5.1 s
```

Have you seen this before and do you know what I am supposed to do here? The repro case is just to change the capture arguments to include a new property (e.g.):
`capture(cdp: Cdp.Api, domain: keyof Cdp.Api, event: string, clearPrevious = false, capacity = 100)`

and then try running `gulp lint`.

connor4312:
You can run `npm run fmt` to apply formatting and linting fixes!

mliao95:
Thanks for the help! I set the capacity to 50 for the time being, but open to changing that as you see fit.  I was testing out the logic of the modified FIFO in [this JSFiddle](https://jsfiddle.net/v80e1wzk/1/). You can start editing the params on line 34 and 36.

connor4312:
Linked list-ified it in https://github.com/microsoft/vscode-js-debug/commit/8ba0b6eebb819a6c19018bcbbeae1a5d831a8933

mliao95:
Thanks for improving the capacity system. 

I took a glance and my only question is there anything differentiating the `Runtime.consoleAPICalled` and `Runtime.executionContextCreated` events? Does the behavior where we evict all `Runtime.consoleAPICalled` events first before evicting a `Runtime.executionContextCreated` event persist after this change?

PR #1241, opened 2022-04-20
feat: apply pathMapping when loading sourcemaps

### LINKED ISSUE #1240: Apply pathMapping when loading sourcemaps
**Describe the bug**
`pathMapping` and `outFiles` provide the ability to use local files for debugging and not transferring them using network protocols. However, `pathMapping` is not applied to the path of a source map file. `sourceMapPathsOverrides` does exist, but is used for files that are referenced by source maps. This results in the following issues.

1. Source maps need to be transferred using network protocols. This sucks especially, if they are large, e.g. due to inline sources.
2. Source maps need to be available in the same way the compiled source is available. In many cases they need to be served by the dev-server. This also prevents usage of source maps with apps that use file-Scheme and run on remote devices, because the remote fs cannot be mapped to the local one.

**To Reproduce**
1. Checkout <https://github.com/timvahlbrock/vscode-js-debug-path-mapping> and open it in VS Code.
2. Run `npm ci` to install dependencies
4. Run `node serve` to start a simple HTTP-Server, which serves the index.html, the app.js but not the source map
5. Add a breakpoint to the app.ts
6. Run `Launch` launch configuration, to compile and launch chrome.
7. See error "**Could not read source map**" in the Debug Console
8. Run command `Debug: Diagnose Breakpoint Problems` and select "What scripts and sourcemaps are loaded"
9. Select the app.js
10. See that path mapping for app.js was successful and it was verified on disk.

**Log File**
```
Verbose logs are written to:
[c:\Users\%user%\AppData\Roaming\Code\logs\20220419T094602\exthost13\ms-vscode.js-debug\vscode-debugadapter-eb2b356c.json.gz]()
Could not read source map for [http://localhost:8125/scripts/app.js](): Unexpected 404 response from [http://localhost:8125/scripts/app.js.map](): 
This is a test
localhost꞉8125/scripts/app.js:1
The content of this file is not really important
```

Full logfile will be send via e-mail.

**VS Code Version:** 1.66.2

**Additional context**
Fixing this should be fairly straightforward. Within sources.ts in [addSource](https://github.com/microsoft/vscode-js-debug/blob/1d734dea33cd4de9c85f65a862d3daff26ea0f81/src/adapter/sources.ts#L846), the source map url needs to be converted like the url of the source file. If that proposal is accepted, I would like to provide a pull-request.

timvahlbrock:
Fixes #1240

### DIFF
--- a/src/adapter/sources.ts
+++ b/src/adapter/sources.ts
@@ -852,6 +852,7 @@ export class SourceContainer {
     contentHash?: string,
   ): Promise<Source> {
     const absolutePath = await this.sourcePathResolver.urlToAbsolutePath({ url });
+
     this.logger.verbose(LogTag.RuntimeSourceCreate, 'Creating source from url', {
       inputUrl: url,
       absolutePath,
--- a/src/common/sourceMaps/sourceMapFactory.ts
+++ b/src/common/sourceMaps/sourceMapFactory.ts
@@ -10,6 +10,7 @@ import { IRootDapApi } from '../../dap/connection';
 import { sourceMapParseFailed } from '../../dap/errors';
 import { MapUsingProjection } from '../datastructure/mapUsingProjection';
 import { IDisposable } from '../disposable';
+import { ILogger, LogTag } from '../logging';
 import { ISourcePathResolver } from '../sourcePathResolver';
 import { fileUrlToAbsolutePath } from '../urlUtils';
 import { ISourceMapMetadata, SourceMap } from './sourceMap';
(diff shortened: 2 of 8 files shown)


timvahlbrock:
Introducing this might break launch configurations, where the source maps are available in the non-mapped path, but not in the mapped path, but I think that this should be a pretty rare case.

src/adapter/sources.ts
```diff
@@ -852,23 +852,31 @@ export class SourceContainer {
...
     const absolutePath = await this.sourcePathResolver.urlToAbsolutePath({ url });
+    let absoluteSourceMapPath = undefined;
+    if (
+      sourceMapUrl &&
+      this.sourcePathResolver.shouldResolveSourceMap({
+        sourceMapUrl,
+        compiledPath: absolutePath || url,
+      })
+    ) {
+      absoluteSourceMapPath = await this.sourcePathResolver.urlToAbsolutePath({
```
connor4312, line 863: The source map might not always exist at an absolute path on disk. For example, when using the webpack-dev-server. That would cause this to be undefined, or an invalid path.

Did the initial code pointer to making the change within the `SourceMapFactory` not work out? If we make the change in there, then we could also easily add fallback logic to avoid breaking existing configurations.

timvahlbrock: Used this way to keep this as much as possible like the pathMapping of the original file source. But I can give it another shot with the code points you provided tomorrow.

timvahlbrock: I pushed another commit. `SourceMapFactory` now tries to load from the path mapped path first, but if that's unsuccessful, loading from the original path is attempted. Also logs a trace message if this happens.

PR #1243, opened 2022-04-20
feat: make hover evaluation timeout configurable

### LINKED ISSUE #1242: Make timeout for evaluation on hover configurable
Currently, value evaluation for hovering [is timed out after 500ms](https://github.com/microsoft/vscode-js-debug/blob/1d734dea33cd4de9c85f65a862d3daff26ea0f81/src/adapter/threads.ts#L438). If CDP communication plus evaluation within the browser takes longer than 500ms, no hover value is shown.

The situation can be resolved in one of the following ways.

- Increase the timeout
- Remove the timeout all together
- Make the timeout configurable within launch or workspace options

Furthermore, VS Code hovering delays the creation of the debug hover by one second to make the UI less hectic (see [#2309](https://github.com/Microsoft/vscode/issues/2309)). Evaluation could be started as soon as hovering is detected, so that the time required for evaluation falls within the UI delay. But this would likely require changes in VS Code itself.

timvahlbrock:
Fixes #1242

### DIFF
--- a/src/adapter/threads.ts
+++ b/src/adapter/threads.ts
@@ -435,7 +435,7 @@ export class Thread implements IVariableStoreLocationProvider {
             includeCommandLineAPI: true,
             objectGroup: 'console',
             generatePreview: true,
-            timeout: args.context === 'hover' ? 500 : undefined,
+            timeout: args.context === 'hover' ? this.getHoverEvalTimeout() : undefined,
           };
 
     if (args.context === 'repl') {
@@ -495,6 +495,17 @@ export class Thread implements IVariableStoreLocationProvider {
       namedVariables: variable.namedVariables,
       indexedVariables: variable.indexedVariables,
     };
+  }
+
+  private getHoverEvalTimeout() {
+    const configuredTimeout = this.launchConfig.timeouts?.hoverEvaluation;
+    if (configuredTimeout === undefined) {
+      return 10 * 1000;
+    }
+    if (configuredTimeout <= 0) {
+      return undefined;
+    }
+    return configuredTimeout;
   }
 
   async _evaluateRepl(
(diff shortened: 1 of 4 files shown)


src/configuration.ts
```diff
@@ -146,7 +146,7 @@ export interface IBaseConfiguration extends IMandatedConfiguration {
   /**
    * Timeouts for several operations (currently only source-maps)
    */
-  timeouts: Partial<SourceMapTimeouts>;
+  timeouts: Partial<SourceMapTimeouts> & HoverEvaluationTimeout;
```
connor4312, line 149: I think we can instead rename `SourceMapTimeouts` and add `hoverEvaluation` to it

timvahlbrock: Thought about this already, but I wasn't sure, whether I should move the definition of `SourceMapTimeouts` then.

timvahlbrock: The `SourceMapTimeouts` type is used in the `Sources` class. If the type is modified & renamed, anywhere in `Sources` where `SourceMapTimeouts` is currently used, the `hoverEvaluation` timeout needs to have been set, although `Sources` although `Sources` is completely unrelated to it and therefor never uses it.

timvahlbrock: I pushed a commit where `SourceMapTimeouts` still is a separate interface, but the types are structured a little bit better.

src/adapter/threads.ts
```diff
@@ -497,6 +497,17 @@ export class Thread implements IVariableStoreLocationProvider {
     };
   }
 
+  private getHoverEvalTimeout() {
+    const configuredTimeout = this.launchConfig.timeouts?.hoverEvaluation;
+    if (configuredTimeout === undefined) {
+      return 10 * 1000;
```
connor4312, line 503: Btw, TS (and recent JS) allows separators so you can write this as `10_000`

connor4312 (approved):
Lgtm, thanks!

PR #1258, opened 2022-05-06
fix: prefer to parse source map directly before failling back to path mapping

connor4312:
Fixes https://github.com/microsoft/vscode/issues/148864. See my comment on the bottom of an explaination of what changed.


roblourens (approved):

timvahlbrock:
Just wanting to comment, that the reason for the original implementation was, that loading source maps from file has better performance than loading over network (which is mostly relevant for large source maps from remote locations only). To take advantage of this with the changes introduced in this PR, the source map just shouldn't be served, which is doable in most cases, I guess.

PR #1261, opened 2022-05-06
Adding support for indexed source maps.

### LINKED ISSUE #1260: Support source map files with sections ('indexed maps')

Using a indexed source map like the kind described in https://github.com/mozilla/source-map/blob/7a263408de1b60e60b6220ca3e36d9b3e4aa047d/lib/source-map-consumer.js#L673-L722 or those produced by the closure-compiler when producing bundled output the JS debugger fails to parse the sourcemap.

https://sourcemaps.info/spec.html#h.535es3xeprgt is the spec for indexed bundles.

When you try to use one of these, this assert will be tripped up: https://github.com/microsoft/vscode-js-debug/blob/main/src/common/sourceMaps/sourceMapFactory.ts#L81 due to the not checking for sections data.

These types of source maps should be supported.

ashgti:
Indexed source maps, described in https://sourcemaps.info/spec.html#h.535es3xeprgt this should allow users to debug bundled output, like that produced by the closure-compiler when producing bundled output.

Fixes #1260

### DIFF
--- a/src/adapter/sources.ts
+++ b/src/adapter/sources.ts
@@ -1067,7 +1067,7 @@ export class SourceContainer {
       const fileUrl = absolutePath && utils.absolutePathToFileUrl(absolutePath);
       const content = this.sourceMapFactory.guardSourceMapFn(
         map,
-        () => map.sourceContentFor(url),
+        () => map.sourceContentFor(url, true),
         () => null,
       );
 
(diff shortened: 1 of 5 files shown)


connor4312:
Thanks for the PR! Can you add some test cases for these as well?

ashgti:
Sure, let me add a unit tests

ashgti:
Added new unit tests to check source map handling.

ashgti:
Let me know if there are any additional tests I should add

connor4312 (approved):
I think this looks good. Thanks for the tests. The one missing thing is support for nested `sections.[].url`s, but I can add that in after this is merged.

alexdima (approved):

PR #1263, opened 2022-05-09
Stop REPL stack traces at the REPL eval frame

### LINKED ISSUE #1259: Stack trace from throw in debug console should hide everything below "eval"

When I do something in the debug console that throws, I get a huge massive error message, which I have to scroll all the way past to figure out what I did wrong (99% of the time, I've hit enter instead of tab when trying to auto-complete; that's a different problem).

The stack trace that is shown for errors that are thrown should drop anything past the eval, since that's where the executed code is actually rooted from. Nothing below eval in the trace string is really controllable; that's just where you are in the debug session.

Anything `eval` and "above" is what should be kept, so you can see errors thrown from functions called from the console, like:

image

jakebailey:
Fixes #1259.

This makes errors thrown in the REPL look like this:

image

Cutting off all of the frames below the eval used to evaluate what was typed in the REPL, which can be arbitrarily large and is already implied by the debugger UI.

I don't have tests for this yet; I'm having trouble running them locally without a whole load timing out or failing because they expect `__proto__` and I get `[[Prototype]]` or something.


connor4312:
> I don't have tests for this yet; I'm having trouble running them locally without a whole load timing out or failing because they expect __proto__ and I get [[Prototype]] or something.

You probably have a mismatched version of Node.js locally -- V8 renamed that property a while ago. Usually I run `npm run test:golden -- -g "name of test"` to run a specific test

jakebailey:
As in, this repo needs node 14? I had been using 16 because the repo uses npm lockfile v2 (otherwise I will break my node 14 install I use for TS, which is still on npm6), but I can just go back to 14 and leave my lock unclean while working.

jakebailey:
The test failures are at least the ones I was expecting, so that's good. I'll fix things up.

jakebailey:
There's still a failure in the console log tests because it looks like the eval source mapping overrides the one in the evaluated code. I'm not quite sure what to do about that.

Additionally, the new source file extension I invented is shown in traces that contain frames from the eval more than once (since the REPL can call back into itself and those frames are shown). I don't think that's ideal, but I'm not sure what to do about that one either. Making the extension be the `.repl.cdp` would hide them, but that's not very desirable when what I want is to hide things before the initial frame, not modify the trace in a significant way.

jakebailey:
I am also questioning whether or not the initial eval frame should be printed or not; given it can show up thanks to self-referential code, it's probably more consistent to show it, but I still have to solve the extension name issue.

src/adapter/threads.ts
```diff
@@ -139,6 +140,9 @@ const sourcesEqual = (a: Dap.Source, b: Dap.Source) =>
   a.sourceReference === b.sourceReference &&
   urlUtils.comparePathsWithoutCasing(a.path || '', b.path || '');
 
+const getReplSourceSuffix = () =>
+  `\n//# sourceURL=eval-${randomBytes(4).toString('hex')}${SourceConstants.ReplExtension}\n`;
```
jakebailey, line 144: I copied the code from the other suffix version, but this won't work because the golden files contain the string, which is random. Hopefully I can find a way to only have them as a marker and keep the original `VM<xx>` text instead.

src/adapter/stackTrace.ts
```diff
@@ -137,14 +137,23 @@ export class StackTrace {
...
   async format(): Promise<string> {
-    const stackFrames = await this.loadFrames(50);
-    const promises = stackFrames.map(frame => frame.format());
+    return await this.formatWithMapper(frame => frame.format());
+  }
+
+  private async formatWithMapper(mapper: (frame: StackFrame) => Promise<string>): Promise<string> {
+    let stackFrames = await this.loadFrames(50);
+    // REPL may call back into itself; slice at the highest REPL eval in the call chain.
+    for (let i = stackFrames.length - 1; i >= 0; i--) {
```
connor4312, line 150: You can also use `findLastIndex`

src/adapter/threads.ts
connor4312: https://github.com/microsoft/vscode-js-debug/blob/42cfe59d08076168ae0099b78891aacdf792f770/src/test/goldenText.ts#L185-L189

src/adapter/stackTrace.ts
```diff
@@ -231,6 +252,7 @@ export class StackFrame {
     name: string,
     rawLocation: RawLocation,
     private readonly isAsync = false,
+    readonly isReplEval = false,
```
connor4312, line 255:
```suggestion
    public readonly isReplEval = false,
```

connor4312 (commented):
Looking good to me overall, some comments

src/adapter/stackTrace.ts
jakebailey: Unfortunately not, for some reason `findIndex` can take a predicate, but `findLastIndex` cannot. (???)

src/adapter/threads.ts
jakebailey: So, you'd prefer if I were to just stirp the number like `VM<xx>`?

jakebailey:
Making this a draft for now. I'm not exactly happy with the stack traces and how they show the faked filename.

Perhaps a better method would be to instead invoke evals as an IIFE with a well-known name?

Ruslan34395 (approved):

connor4312:
> Preferably, I'd like the path to look somewhat like node's REPL

You can edit that here 🙂 

https://github.com/microsoft/vscode-js-debug/blob/98b20c31eabf765274e49c421d436820e4d1692a/src/adapter/sources.ts#L332-L340

> I already don't like the fact that this method of identifying the eval overrides any user provided source map for the REPL, but I suppose nobody's putting their own source mapping into REPL statements.

It actually doesn't -- it would override the _url_, but not the map. It's already quite unlikely for a sourcemap with external references to work, but a fully inline sourcemap would still work.

jakebailey:
Aha, thanks. I'll change that to switch it to `repl` or something.

jakebailey:
I've changed it so that you get `repl:` in the trace. To do this, I made `formatAsNative` not check `url`; maybe that's wrong but the difference feels superficial. You can see the difference in the golden files.

However, I get this error when I run the tests (this has probably been true the whole time):

```
rejected promise not handled within 1 second: TypeError: Cannot read properties of undefined (reading 'endsWith')
stack trace: TypeError: Cannot read properties of undefined (reading 'endsWith')
        at Function.fromDebugger (e:\work\vscode-js-debug\out\src\adapter\stackTrace.js:187:127)
        at Function.fromDebugger (e:\work\vscode-js-debug\out\src\adapter\stackTrace.js:83:44)
        at Thread._createPausedDetails (e:\work\vscode-js-debug\out\src\adapter\threads.js:775:52)
        at Thread._onPaused (e:\work\vscode-js-debug\out\src\adapter\threads.js:544:59)
        at runMicrotasks (<anonymous>)
        at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

The protocol says that `url` in this case must exist and be a string; something upstream is breaking that invariant. Maybe some test that is casting?

connor4312:
When you run the tests, on a failure, it should print the location of a `.json` file. Could you share that? That'll have more information.

jakebailey:
Sorry for the delay; was OOF. Here's the JSON file (with extension txt so GitHub lets me upload it):

[node-runtime-chakracore-string-value.txt](https://github.com/microsoft/vscode-js-debug/files/8765704/node-runtime-chakracore-string-value.txt)

src/test/console/console-format-applies-skipfiles-to-logged-stacks.txt
```diff
@@ -1,2 +1,2 @@
 logged hello world
- at dont-ignore-me.js:1:1
+ at repl:1:1
```
jakebailey, line 2: This test probably needs to be changed to not use the repl, because I think the intent is to test that this file is not dropped from the stack.

jakebailey: That, or this is an example of the source mapping not working anymore.

src/test/evaluate/evaluate-default.txt
```diff
@@ -73,14 +73,14 @@ result: 3
 
 Evaluating#1: setTimeout(() => { throw new Error('bar')}, 0)
 stderr> Uncaught Error Error: bar
-    at <anonymous> (http://localhost:8001/eval1.js:1:26)
+    at <anonymous> (localhost꞉8001/eval1.js:1:26)
```
jakebailey, line 76: These changes are due to me dropping `uiLocation?.source.url` use from `formatAsNative` so the pretty name shows up instead.

jakebailey:
@connor431 I have merged main and simplified a few thing; I don't know that the above error happens anymore locally on my machine, but it doesn't look like Pipelines is running on my PR anymore so I can't tell if I'm still breaking something on Windows. I'm thinking the undefined bug got fixed sometime this month.

This stack trace thing is annoying me quite a bit so I'm hoping to see this one go in.

connor4312 (approved):
lgtm

PR #1265, opened 2022-05-10
fix: Align hoverEvaluation config suggestion with actual default

timvahlbrock:
In 877e0ae1957172d905b892d898e52b5322853899 the hoverEvaluation timeout default was changed back to 500ms for undocumented reasons. However, the suggested default when adding the timeout to a `launch.json` wasn't changed back.


connor4312:
Thanks!

I swapped it back because I think 500ms is a reasonable default. At 10_000ms we have a greater liklihood of blocking the runtime, if something is expensive to compute, which is a bad experience for users. As we haven't had complaints about the current timeout before, I think it's a sane default. Let me know if you disagree.

Issue #1209, opened 2022-02-22
Support for debugging webviews in UWPs


connor4312:
Today js-debug can debug webview2's embedded in C++, but @DavidShoe requested we also add support for webviews in UWPs.

This is a little more involved, and he created a [sample](https://github.com/DavidShoe/NamedPipeTest) repo that demonstrates this. Primarily, acquiring the pipe involves some low level win32 calls. Therefore I've wrapped this in a native Node module in https://github.com/microsoft/vscode-win32-app-container-tokens, and will integrate it from there into js-debug.

connor4312:
Marking verified from @DavidShoe's report

PR #1282, opened 2022-05-20
feat: support UWP webview debugging

connor4312:
Fixes https://github.com/microsoft/vscode-js-debug/issues/1209

cc @DavidShoe


src/build/generate-contributions.ts
```diff
@@ -1022,9 +1022,10 @@ const edgeAttachConfig: IDebugger<IEdgeAttachConfiguration> = {
   configurationAttributes: {
     ...chromiumAttachConfigurationAttributes,
     useWebView: {
-      type: 'boolean',
-      description: refString('edge.useWebView.description'),
-      default: false,
+      type: 'object',
```
roblourens, line 1025: Is back-compat an issue?

roblourens (approved):

src/build/generate-contributions.ts
connor4312: Oh, good catch, this should be oneOf

PR #1296, opened 2022-06-08
feat: use custom toString() methods to generate object descriptions

### LINKED ISSUE #1284: Feature Request: Show toString of Objects In Hover
It would be really helpful if the hover title would show the `.toString()` result of the object:

image

In this case, `.toString()` of the range would be `[3,3)`.

v8 allows you to call a function (such as `.toString()`) but throw before any side-effect is caused.

connor4312:
Closes #1284


src/adapter/templates/getStringyProps.ts
```diff
@@ -0,0 +1,30 @@
...
+ * to their string representations.
+ */
+export const getStringyProps = remoteFunction(function (this: unknown, maxLength: number) {
+  const out: Record<string, string> = {};
+  if (typeof this !== 'object' || !this) {
+    return out;
+  }
+
+  for (const [key, value] of Object.entries(this)) {
+    if (typeof value === 'object' && value && !String(this.toString).includes('[native code]')) {
```
connor4312, line 18: This detection is a bit ugly. Not sure if there's a more clever way to do this. Simply `this.toString === Object.prototype.toString` isn't enough because that also incorrectly stringifies boxed types, errors, and arrays.

hediet (approved):

hediet (approved):

PR #1300, opened 2022-06-09
fix: custom toSTring format not applying to nested props

connor4312:
Last minute fix and not thorough enough tests 🤦


hediet (approved):

hediet:
Still failing...

connor4312:
That test is a flakey one I need to fix...

PR #1302, opened 2022-06-10
feat: keep a ring buffer of logs, and allow saving them 

connor4312:
**Debug: Save Diagnostic JS Debug Logs** will show a save dialog, which will then save the last 512 logs from the selected debug session. This should let people report transient errors more easily, since they don't need to restart with `trace: true` to get some information.

I also cleaned some debt by removing extra `trace` options that are never used.


src/cdp/api.d.ts
```diff
@@ -1532,7 +1532,7 @@ export namespace Cdp {
```
connor4312, line 1532: Automated update here as part of `npm run generateapis`

hediet (approved):

PR #1307, opened 2022-06-14
feat: implement step in targets

connor4312:
For https://github.com/microsoft/vscode/issues/123879

It uses possible breakpoint locations to extra and display possible step ranges. It tries to parse the source line to get the appropriate label and range to return to VS Code, which works pretty well in my testing.

image

Step into works similarly to what core does in "run to line" -- it sets a breakpoint on the desired location, resumes the debugger (skipping all pauses until the breakpoint is hit), then removes the breakpoint and steps in.


src/common/sourceUtils.ts
```diff
@@ -358,3 +358,52 @@ export function getSyntaxErrorIn(code: string): Error | void {
...
+
+  return steppable || other;
+}
+
+const notParensRe = /^[^()]+/g;
+
+/**
+ * Gets a "step into" target at the given offset in the line of code.
+ */
+export function getStepTargetInfo(line: string, offset: number) {
```
roblourens, line 387: Is this still going to return breakpoint locations that are not step into targets? I'm not sure

roblourens (approved):

PR #1311, opened 2022-06-22
Support "supportsDebuggerProperties" from DAP

thaystg:
Implement support to Debugger Properties, then we can pass this information to BrowserDebugProxy and give the Wasm programmer a better debugging experience.


src/dap/api.d.ts
```diff
@@ -1099,6 +1099,11 @@ export namespace Dap {
     setSourceMapSteppingRequest(
       params: SetSourceMapSteppingParams,
     ): Promise<SetSourceMapSteppingResult>;
+
+    /**
+     * Configures whether source map stepping is enabled.
```
radical, line 1104: `source map stepping` - comments needs to be updated

src/cdp/api.d.ts
```diff
@@ -9639,6 +9640,31 @@ export namespace Cdp {
...
+   /**
+   * Types of the 'DotnetDebugger' domain.
+   */
+  export namespace DotnetDebugger {
+    /**
+     * Parameters of the 'DotnetDebugger.setDebuggerProperty' method.
+     */
+    export interface SetDebuggerPropertyParams {
+      /**
+       * Whether to enable to disable debugger property.
```
radical, line 9662:
```suggestion
       * Whether to enable or disable debugger property.
```

src/adapter/debugAdapter.ts
```diff
@@ -117,6 +117,11 @@ export class DebugAdapter implements IDisposable {
     this.dap.on('setExcludedCallers', params => this._onSetExcludedCallers(params));
     this.dap.on('saveDiagnosticLogs', ({ toFile }) => this._saveDiagnosticLogs(toFile));
     this.dap.on('setSourceMapStepping', params => this._setSourceMapStepping(params));
+    this.dap.on('setDebuggerProperty', params => this._setDebuggerProperty(params));
+  }
+
+  private _setDebuggerProperty(params: Dap.SetDebuggerPropertyParams) {
+    this._thread?.cdp().DotnetDebugger.setDebuggerProperty(params);
```
connor4312, line 124: Since the DAP request is generic, should this too have a more generic domain?

thaystg: Fixed.

connor4312 (approved):

PR #1326, opened 2022-07-08
fix: completions not returning stack variables

connor4312:
We were just bailing out too soon. VS Code's behavior of reusing previous completions helped hide this bug.

Fixes https://github.com/microsoft/vscode/issues/153651


tanhakabir (approved):

connor4312:
Looks like tests in Insiders broke, I will change the pipeline to use stable for now: https://github.com/microsoft/vscode/issues/154549

PR #1339, opened 2022-07-22
Double check if native tostring is [object *]

### LINKED ISSUE #1338: Avoid using [object Object] for previews
From https://github.com/microsoft/vscode/issues/155718

> Issue Type: Bug
> 
> Development environment is an angular workspace
> 
> Since VS Code Update 1.69.x debug information are not properly shown. All objects are displayed as [object Object]. (see screenshots to compare Versions)
> 
> Steps to reproduce: Debug any Angular component After breakpoints hits, hover a variable or put it in watch
> 
> // Edit: it still occurs in 1.69.2
> 
> Version 1.68.1 grafik Version 1.69.1 grafik
> 
> VS Code version: Code 1.68.1 ([30d9c6c](https://github.com/microsoft/vscode/commit/30d9c6cd9483b2cc586687151bcbcd635f373630), 2022-06-14T12:48:58.283Z) OS version: Windows_NT x64 10.0.17763 Restricted Mode: No
> 
> System Info
> Extensions (31)

Balastrong:
This PR fixes #1338 

As mentioned in the vscode repo (see https://github.com/microsoft/vscode/issues/155718#issuecomment-1190550755) it might make sense to doublecheck and avoid using the native toString if the output is `[object *]` and there is a better alternative.

### DIFF
--- a/src/adapter/templates/getStringyProps.ts
+++ b/src/adapter/templates/getStringyProps.ts
@@ -15,7 +15,12 @@ export const getStringyProps = remoteFunction(function (this: unknown, maxLength
   }
 
   for (const [key, value] of Object.entries(this)) {
-    if (typeof value === 'object' && value && !String(value.toString).includes('[native code]')) {
+    if (
+      typeof value === 'object' &&
+      value &&
+      !String(value.toString).includes('[native code]') &&
+      !String(this).includes('[object ')
+    ) {
       out[key] = String(value).slice(0, maxLength);
     }
   }
@@ -24,7 +29,12 @@ export const getStringyProps = remoteFunction(function (this: unknown, maxLength
 });
 
 export const getToStringIfCustom = remoteFunction(function (this: unknown, maxLength: number) {
-  if (typeof this === 'object' && this && !String(this.toString).includes('[native code]')) {
+  if (
+    typeof this === 'object' &&
+    this &&
+    !String(this.toString).includes('[native code]') &&
+    !String(this).includes('[object ')
+  ) {
     return String(this).slice(0, maxLength);
   }
 });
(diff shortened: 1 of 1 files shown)


src/adapter/templates/getStringyProps.ts
```diff
@@ -24,7 +29,12 @@ export const getStringyProps = remoteFunction(function (this: unknown, maxLength
 });
 
 export const getToStringIfCustom = remoteFunction(function (this: unknown, maxLength: number) {
-  if (typeof this === 'object' && this && !String(this.toString).includes('[native code]')) {
+  if (
+    typeof this === 'object' &&
+    this &&
+    !String(this.toString).includes('[native code]') &&
+    !String(this).includes('[object ')
```
connor4312, line 36: In some cases, calling `String(this)` may be expensive. We should call it when the first three conditions pass and nest an `if` to return.

```
const str = String(this)
if (!str.startsWith('[object ')) {
  return str.slice(0, maxLength)
}
```

src/adapter/templates/getStringyProps.ts
```diff
@@ -15,7 +15,12 @@ export const getStringyProps = remoteFunction(function (this: unknown, maxLength
   }
 
   for (const [key, value] of Object.entries(this)) {
-    if (typeof value === 'object' && value && !String(value.toString).includes('[native code]')) {
+    if (
+      typeof value === 'object' &&
+      value &&
+      !String(value.toString).includes('[native code]') &&
+      !String(this).includes('[object ')
```
connor4312, line 22:
```suggestion
      !String(value).includes('[object ')
```

(also, please apply the suggestion below to this one)

connor4312:
Thanks for the PR! Couple small comments

Balastrong:
Thank you very much for the comments, they totally make sense :)

I've updated the PR!

connor4312 (approved):

meganrogge (approved):

PR #1341, opened 2022-07-22
Adopt argsCanBeInterpretedByShell

### LINKED ISSUE #1335: Adopt argsCanBeInterpretedByShell for RunInTerminalRequest
Implement launch config treatment for the argsCanBeInterpretedByShell flag. Continuing discussion from https://github.com/microsoft/vscode/issues/149910

roblourens:
Fix #1335

### DIFF
--- a/src/common/arrayUtils.ts
+++ b/src/common/arrayUtils.ts
@@ -1,0 +1,11 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+export function asArray<T>(thing: T | ReadonlyArray<T>): T[] {
+  if (Array.isArray(thing)) {
+    return thing;
+  } else {
+    return [thing as T];
+  }
+}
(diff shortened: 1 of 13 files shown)


src/common/arrayUtils.ts
```diff
@@ -0,0 +1,11 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+export function asArray<T>(thing: T | ReadonlyArray<T>): T[] {
```
roblourens, line 5: Ideally this would return a `ReadonlyArray<T>` when given a `ReadonlyArray<T>`, and return a `T[]` when given a `T[]`. That consumed too much time, I suppose I'm missing a TS trick.

src/common/arrayUtils.ts
```diff
@@ -0,0 +1,11 @@
...
+  if (Array.isArray(thing)) {
+    return thing;
+  } else {
+    return [thing as T];
```
roblourens, line 9: And even when adding a custom isArray function that does a type guard with the proper generic parameter, I couldn't get rid of this cast...

connor4312 (dismissed):

connor4312 (approved):

PR #1347, opened 2022-07-29
Avoid trimMiddle when double click in a variable from VisualStudio

thaystg:
First of all I'm sorry to open a PR and don't open a issue before, we can discuss the solution here and if this is not the right approach I can close this PR.
I have already discussed offline with people from VS team and they told me that they cannot change their side because this could break other clients that implements DAP.

Trying to fix double click in a variable that has more than 100 characteres when debugging from VisualStudio.

As vscode-js-debug extension receives a message from dap like this: 
`{"tag":"dap.receive","timestamp":1659118931254,"metadata":{"connectionId":1,"message":{"type":"request","command":"evaluate","arguments":{"expression":"str","frameId":185,"timeout":1500},"seq":13}},"level":0}`

And there is no context, the default behavior for vscode-js-debug extension is to trim the variable in the middle, then the user cannot see the correct content of the variable.

My suggestion is: if vscode-js-debug receives no context information from dap in evaluate command, considers it as a watch and trim only if the len is higher than 1000.

If you don't think this is the right approach, we can also check if `clientID === 'visualstudio'` to do this change.

Fix https://github.com/dotnet/runtime/issues/71552

### DIFF
--- a/src/adapter/objectPreview/contexts.ts
+++ b/src/adapter/objectPreview/contexts.ts
@@ -42,7 +42,7 @@ const hover: IPreviewContext = {
 };
 const copy: IPreviewContext = { budget: Infinity, quoted: false };
 const watch: IPreviewContext = { budget: 1000, quoted: true, postProcess: escape };
-const fallback: IPreviewContext = { budget: Infinity, quoted: true };
+const fallback: IPreviewContext = { budget: 100_000, quoted: true };
 
 export const getContextForType = (type: PreviewContextType | string | undefined) => {
   switch (type) {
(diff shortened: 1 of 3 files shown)


connor4312:
I think this is a good change, but the default should be changed here instead:

https://github.com/microsoft/vscode-js-debug/blob/5084ec844e06ca6dd8c09841472c184410cc2b7d/src/adapter/objectPreview/contexts.ts#L45

connor4312:
I would prefer something high like `100_000`.

Also, there's a unit test that will need to be updated: https://dev.azure.com/vscode/VS%20Code%20debug%20adapters/_build/results?buildId=69996&view=logs&jobId=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=f33089cd-d3fe-578c-2c5e-4ee560e10a0d

connor4312 (approved):
Thanks!

PR #1354, opened 2022-08-03
Fix when there are two sources with same prefix

thaystg:
Again, I'm sorry to create the PR and don't discuss it in an issue before, we can discuss here and I can close it if this is not the right approach.

Like: other.js and other.js2, as we were not adding the $ it was setting the breakpoint in both files if we want to set on other.js.

In the case of debugging a Blazor app, it was only setting in the first found file, which could be the wrong one.

Fixes https://github.com/dotnet/runtime/issues/56029


connor4312:
The change looks OK, however tests are currently failing

thaystg:
> The change looks OK, however tests are currently failing

Fixed.

connor4312:
It looks like there are still a few failures: one with Blazor debugging, and some others where this regex is now excluding sources loaded with a query string from matching.

connor4312 (approved):

PR #1360, opened 2022-08-04
fix: race in entrypoint breakpoints

connor4312:
A race was causing a ~10% failure in a test, and could have caused some
failures for users in edge cases (where rarely stepping into the
first function or class declaration in a file would not work.)

Also a janitorial update of Node and pipeline OS images.


connor4312:
There's still the "webview breakpoints" flake which caused the last failures. I will diagnose that when I'm back on windows.

roblourens (approved):

Issue #1328, opened 2022-07-10
Add argument for specifying host for `debugServerMain.ts`, `vsDebugServer.ts`, `flatSessionLauncher.ts`


gfszr:
**Is your feature request related to a problem? Please describe.**
All debug adapter frontends do not offer an argument to specify the host to listen on. This causes several confusions, for instance: `vsDebugServer` listens on the root session on external IP, but child sessions listen to `127.0.0.1` hardcoded: https://github.com/microsoft/vscode-js-debug/blob/98af79b65794d81771fb62443dac240b0b03ef37/src/common/findOpenPort.ts#L99-L105

**Describe the feature you'd like**
Add an optional host argument to all debug adapter executables, which specifies a hostname all created TCP servers would listen on.

weinand:
Verified by code inspection

PR #1362, opened 2022-08-06
Allow specifying host:port on vsDebugServer.ts

gfszr:
This PR addresses parts of https://github.com/microsoft/vscode-js-debug/issues/1328 by adding a new argument, `host`, that comes after specifying a `port`. When `host` is given, it is being propagated internally to where servers are being created. For backwards compatability reasons:
1. Only a port can still be given.
2. Default listen IP stays the same as previously, both on the initial server and on child session servers.

### DIFF
--- a/src/serverSessionManager.ts
+++ b/src/serverSessionManager.ts
@@ -26,11 +26,15 @@ export class ServerSessionManager<T extends IDebugSessionLike> {
   private readonly portLeaseTracker: IPortLeaseTracker;
   private disposables: IDisposable[] = [];
   private servers = new Map<string, net.Server>();
+  private host = '127.0.0.1';
 
-  constructor(globalContainer: Container, sessionLauncher: ISessionLauncher<T>) {
+  constructor(globalContainer: Container, sessionLauncher: ISessionLauncher<T>, host?: string) {
     this.sessionManager = new SessionManager(globalContainer, sessionLauncher);
     this.portLeaseTracker = globalContainer.get(IPortLeaseTracker);
     this.disposables.push(this.sessionManager);
+    if (host) {
+      this.host = host;
+    }
   }
 
   /**
@@ -106,6 +110,7 @@ export class ServerSessionManager<T extends IDebugSessionLike> {
         deferredConnection.resolve(session.connection);
       },
       port,
+      this.host,
     );
 
     this.servers.set(debugSession.id, debugServer);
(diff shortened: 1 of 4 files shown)


src/vsDebugServer.ts
```diff
@@ -126,15 +126,34 @@ class VsDebugServer implements ISessionLauncher<VSDebugSession> {
   }
 }
 
-const debugServerPort = process.argv.length >= 3 ? +process.argv[2] : undefined;
+let debugServerPort: number | undefined = undefined;
+let debugServerHost: string | undefined = undefined;
+if (process.argv.length >= 3) {
+  // Interpret the argument as either a port number, or 'address:port'.
+  const address = process.argv[2];
+  const colonIndex = address.lastIndexOf(':');
```
connor4312, line 134: I suggest that we don't do this, since this won't work if the user wants to listen on an ipv6 interface. (We could do better parsing, but I think having separate arguments is fine, especially since this CLI is mostly consumed by machines and not humans)

src/serverSessionManager.ts
```diff
@@ -26,11 +26,15 @@ export class ServerSessionManager<T extends IDebugSessionLike> {
   private readonly portLeaseTracker: IPortLeaseTracker;
   private disposables: IDisposable[] = [];
   private servers = new Map<string, net.Server>();
+  private host = '127.0.0.1';
 
-  constructor(globalContainer: Container, sessionLauncher: ISessionLauncher<T>) {
+  constructor(globalContainer: Container, sessionLauncher: ISessionLauncher<T>, host?: string) {
```
connor4312, line 31: You can make the constructor signature be

```suggestion
  constructor(globalContainer: Container, sessionLauncher: ISessionLauncher<T>, private readonly host = '127.0.0.1') {
```

And then remove the property above and conditional below

connor4312 (commented):
Apologies for the delay in review. Two small comments!

gfszr:
> Apologies for the delay in review. Two small comments!
It's totally fine, don't worry about it 😄

src/vsDebugServer.ts
gfszr: I've seen a similar support of `:` parsing in another adapter entry point (`debugServerMain.ts` I think?), so I reused the same code. However, as it is rather easy to change now and will prevent future breaking of the command line, changing it now seems like the right thing to do 👍

connor4312 (approved):

hediet (approved):

PR #1365, opened 2022-08-11
Remove resolvers proposal and use tunnels instead

alexr00:
I've moved the tunnels api proposal into a separate proposal, which JS Debug, since it isn't a resolver, should use instead. I made sure it still builds, but I haven't tested past that.


package.json
```diff
@@ -173,7 +173,8 @@
     "portsAttributes",
     "findTextInFiles",
     "workspaceTrust",
-    "resolvers"
+    "resolvers",
```
alexr00, line 176: Once all extensions that need to be switched over to `"tunnels"` are switched, I will update the checkes in VS Code and we can remove `"resolvers"`.

connor4312 (approved):

PR #1376, opened 2022-08-27
Fix hex typo

turuslan:
Memory inspector swapped "E" and "D" nibbles.  
It happened because there was a typo in hex decoding.  
Problem was only with reading memory, written memory was correct.


connor4312 (approved):

joaomoreno (approved):
Oh wow!

PR #1385, opened 2022-09-03
Pass cwd to getVersionText

andresiraola:
[Volta](https://volta.sh/)'s shim relies on the project `package.json` to find the correct node executable version to be used. Because `getVersionText` is not setting `cwd` when calling `node --version`, the shim responds with an error.

This PR fixes that by passing `cwd` all the way to `getVersion`. As far as I've seen, all the other calls `node` get the `cwd` set correctly. This was the only exception.

### DIFF
--- a/src/targets/node/nodeBinaryProvider.ts
+++ b/src/targets/node/nodeBinaryProvider.ts
@@ -192,6 +192,7 @@ export interface INodeBinaryProvider {
     env: EnvironmentVars,
     executable?: string,
     explicitVersion?: number,
+    cwd?: string,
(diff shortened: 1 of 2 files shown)


connor4312:
Not sure why tests failed catastrophically for this PR, as they're fine on `main` and locally. Re-running...

connor4312 (approved):

lramos15 (approved):

connor4312:
And now they're fine. Maybe we had a bad Insiders build 🤷‍♂️

Issue #1383, opened 2022-09-02
Unable to preview named import for ts with sourcemap


liuxingbaoyu:
**Describe the bug**
When debugging ts files with source maps, named imported variables are not previewed properly.

**To Reproduce**
1. `git clone https://github.com/liuxingbaoyu/jsdebug-sourcemap-bug`
2. `yarn / npm i`
3. set breakpoint at `src\util\scope.ts` L18
4. start debugging `Launch Program`
5. preview `BIND_SCOPE_LEXICAL`
6. see error

**VS Code Version: 1.71.0 (user setup)

**Additional context**
![image](https://user-images.githubusercontent.com/30521560/188241421-379db8c2-bc6e-4c79-b7db-c219df17f5dc.png)
![image](https://user-images.githubusercontent.com/30521560/188241473-2b6ca158-e76d-4395-a665-292040b1845a.png)

connor4312:
This is a result of how TS emits its `names`, not something the debugger controls.

liuxingbaoyu:
Thanks for the quick reply, I'll look into the source map.

liuxingbaoyu:
I've re-investigated and this seems to be something the debugger can fix.

As a simple test, I modified https://github.com/microsoft/vscode-js-debug/blob/main/src/common/sourceMaps/renameProvider.ts#L21 to `const identifierRe = /[$a -z_][$0-9A-Z_$]*.?[$0-9A-Z_$]*/iy;`, it works fine.

![image](https://user-images.githubusercontent.com/30521560/194187512-c48a0713-ae69-48c3-8073-d07a450832c4.png)

connor4312:
The identifierRe is pretty strict because the sourcemap only gives us where a name _starts_, not the length of it. But I think including `.` in that should be fine.

PR #1419, opened 2022-10-07
fix: Rename identifier for source map

liuxingbaoyu:
I found that `computeColumnSpans` can be used to get a more accurate identifier.
I only did a simple test locally, but this should theoretically be fine.

Fixed: #1383

### DIFF
--- a/src/common/sourceMaps/renameProvider.ts
+++ b/src/common/sourceMaps/renameProvider.ts
@@ -114,7 +114,7 @@ export class RenameProvider implements IRenameProvider {
       const start = toOffset.convert(startPos);
 
       const lastGeneratedColumn = (mapping as MappingItem & { lastGeneratedColumn: number })
-        .lastGeneratedColumn as number;
+        .lastGeneratedColumn;
 
       const endPos = new Base01Position(mapping.generatedLine, lastGeneratedColumn + 1);
       const end = toOffset.convert(endPos);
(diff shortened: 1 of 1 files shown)


liuxingbaoyu:
I don't understand what's going on with CI, is this related to PR?

connor4312:
Thanks for the PR. I will get to this in a little bit--I'm looking at switching to a new, _mostly_ compatible sourcemap library which at minimum will cause some merge conflicts here, but I will get it merged.

connor4312:
Thanks again, I extended this in https://github.com/microsoft/vscode-js-debug/pull/1481 to avoid computing the whole set of column spans and just using the next mapping as we iterate (I`eachMapping` ensures order)

liuxingbaoyu:
Sorry, there seems to be an issue with this PR. (may block publishing)
It seems that some generator sourcemaps may have inaccurate identifier ranges.
For example `(aaaa)`, some generators may generate `aaaa)` as an item.
I fear this PR will make this worse.
Maybe we can introduce a detection to allow fallback, we use regex to detect all text in the identifier range, if it contains ` `,`)`,`,`, then let us fall back to using regex logic. (maybe allow `.`)
Also I realize there's a possible problem with the new logic treating member expressions as allowing previews, which has the slightest chance of side effects (getters), but that's probably fine since we're using the location in the source code to look for, which means that only compiler-generated member expressions are previewed.

connor4312:
hm, thanks. For now I think I will just go back to the original approach, and modify the regex to allow member access with `.`

Issue #1179, opened 2022-01-16
Node debugging extremely slow on startup and after first breakpoint


imnitishng:
**Description**
ExpressJS application becomes unresponsive for a long time whenever VSCode debugger is attached to the application.
The application works as expected and starts up instantly when run normally in a terminal (`node dist/server.js`) or using node default debugger (`node inspect dist/server.js`). Just a note that the Express JS application in context here isn't super resource intensive, just some basic CRUD operations and debugging experience on VSCode on other low specs Linux machines is seamless and has no such issue.
The issue only occurs whenever VSCode's debugger is attached to the application, be it in any form (launch configuration, javascript debug terminal or node process attach). 

Here is the full debugging scenario for more context on the issue
1. If the application is launched using the launch configuration the application won't start instantly (i.e. if I goto `localhost:3000` the browser will stuck loading for 2-3 mins), it eventually starts and the first breakpoint is hit but once I step forward in the debugger nothing happens and I have to wait 1-2 mins to reach the next line, after that the debugger works normally. So the main issue is the unresponsive behaviour on attaching debugger and after step over the first breakpoint.
2. Now if I start the application normally in the terminal, it starts working the browser almost instantly, now once I attach VSCode  debugger to the node process and hit any of the application endpoint it hangs and I'll have to wait for the debugger to hit the first breakpoint, and then again wait after stepping forward from the first breakpoint. After this things work normally without any waiting times.

**To Reproduce**
1. Run an express JS application with the following run configurations
```json
{
    "configurations": [
// for scenario 2 listed above
        {
            "name": "Attach by Process ID",
            "processId": "${command:PickProcess}",
            "request": "attach",
            "skipFiles": [
                "<node_internals>/**",
                "${workspaceFolder}/node_modules/**/*.js",
            ],
            "sourceMaps": true,
            "trace": true,
            "type": "pwa-node"
        },
// for scenario 1 listed above
        {
            "type": "pwa-node",
            "request": "launch",
            "name": "Debug with watch task",
            "skipFiles": [
                "<node_internals>/**",
                "${workspaceFolder}/node_modules/**/*.js"
            ],
            "program": "${workspaceFolder}/dist/server.js",
            "outFiles": [
                "${workspaceFolder}/dist/**/*.js"
            ],
            "sourceMaps": true,
            "trace": true,
            "preLaunchTask": "tsc: watch - tsconfig.json"
        }
    ]
}
``` 

**Log File**
[...]

PR #1422, opened 2022-10-07
fix: improve skipfile performance

connor4312:
Previously, `skipFiles` applied a new regex and eventual call to `Debugger.setBlackboxPatterns` for every source. In #1179, this resulted in very bad performance, since a regex was sent for each file loaded in the user's `node_modules` (several thousand). We needed to do this since `setBlackboxPatterns` doesn't have a "negation", so we had to compute on the debugger side which files satisfied `skipFiles` and send those to the runtime.

This commit makes some things smarter, and some things less smart. It restricts `skipFiles` to very simple globs, allowing only wildcards and negations. But, in exchange, it's able to make a finite number of regexes based solely on those patterns. In most cases, this means that we never need to send further `setBlackboxPatterns` updates to the runtime after loading.

For example,

```js
{ skipFiles: ['**/node_modules/**', '!**/node_modules/my-library/**'] }
```

...compiles to...

```
/^(?!.*\/node_modules\/my-library\/.*$).*\/node_modules\/.*$/
```

We do still use the old strategy if a URL isn't skipped but maps to a path that does skip. We could eventually make this smarter if the SourcePathResolver's `absolutePathToUrlRegex` was mixed in somehow... but those can't be combined or negated as simply as glob patterns.

This does run the risk of breaking users who use advanced glob patterns in their `skipFiles`, but I haven't really encountered any of those and I believe the tradeoff is worth it.

Fixes https://github.com/microsoft/vscode-js-debug/issues/1179

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,9 @@ This changelog records changes to stable releases since 1.50.2. "TBA" changes he
 
 ## Nightly (only)
 
-- fix: copying a date object resulting in an empty object ([vscode#162747](https://github.com/microsoft/vscode/issues/162747))
 - feat: add automatic support for nested sourcemaps ([#1390](https://github.com/microsoft/vscode-js-debug/issues/1390))
+- fix: copying a date object resulting in an empty object ([vscode#162747](https://github.com/microsoft/vscode/issues/162747))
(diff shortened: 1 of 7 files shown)


roblourens (dismissed):
Looks great, but maybe there actually should be a comment on the launch config property about what syntax is supported. I could imagine people using `{}` or other fancy syntax. Maybe a note in the release notes

roblourens (approved):

Issue #1424, opened 2022-10-10
SourceRoot not Considered as Absolute Path 


FrankEssenberger:
**Describe the bug**

Here is a small [sample repo](https://github.com/FrankEssenberger/source-map-tests).

We try to use the sourceRoot to add the sources in a remote location so that we do not need to ship them. However, the mapping:

```json
{"version":3,"file":"string.js","sourceRoot":"http:/raw.githubusercontent.com/SAP/cloud-sdk-js/main/packages/util/src/","sources":["string.ts"],"names":[],"mappings":";......"}
```

leads to an error because the CWD is added in front of the source route.

**To Reproduce**

- Checkout repo.
- Run install
- Execute `debug-test.mjs` via `launch.json`
- Step into method -> see the Error

```Could not load source '/Users/XXX/source-map-test/node_modules/@sap-cloud-sdk/util/dist/http:/raw.githubusercontent.com/SAP/cloud-sdk-js/main/packages/util/src/string.ts': Unable to retrieve source content.```

**VS Code Version:** 1.72.0

FrankEssenberger:
@connor4312 Thanks a lot for the quick reply. I guess the `/` was taken by the `join()` method  of the `path` module. Looks like this is also used internally because even when the source root contains the double slash in the map file in the resulting path printed in the error message the double slash is gone. 

Anyhow if it is not supported it will not work. Since you said `do not currently support` is there a hope that you do it in the future? If so do you have a rough timeline?

connor4312:
I don't have a timeline at the moment, but this may be easy to add. Code pointer:

https://github.com/microsoft/vscode-js-debug/blob/a5995596511282127cf31932449d89c22588ee24/src/adapter/sources.ts#L1136-L1140

Currently `fileUrl` is only a file URL, but if we could do an extra check to see if the source URL is an absolute URL and if so pass that as a ~fileUrl~ contentUrl instead

FrankEssenberger:
I guess there is already a [fetchHtt](https://github.com/microsoft/vscode-js-debug/blob/a5995596511282127cf31932449d89c22588ee24/src/adapter/resourceProvider/basicResourceProvider.ts#L52) but it is not used because the [absolutePath](https://github.com/microsoft/vscode-js-debug/blob/a5995596511282127cf31932449d89c22588ee24/src/adapter/resourceProvider/basicResourceProvider.ts#L38) is fooled since the URL is not starting with a `/`. 

[...]

PR #1427, opened 2022-10-14
Allow for URL as sourceRoot

FrankEssenberger:
Relates to #1424.

@connor4312 I created a small PR. I was a bit lost in the code base. The whole SourceContainer has so much state that I could not get a unit test running. I then just added a test and checked that the `fetchHttp` is called and the sources are really loaded from the local host.

The behaviour in the `getComputedSourceRoot` is now a bit inconsistent for the `file://` the protocol  is removed and later added, which should of course not happen if it is a `http://` source. A proper solution would perhaps to have an object not only a string for the URL. Then the object could hold the information I am `file`, `data`, `http`, `ftp` or whatever.

I did not want to change much so I added just minimal checks.

### DIFF
--- a/src/adapter/scriptSkipper/implementation.ts
+++ b/src/adapter/scriptSkipper/implementation.ts
@@ -260,6 +260,7 @@ export class ScriptSkipper {
     if (
       !skipped &&
       source.absolutePath &&
+      urlUtils.isAbsolute(source.absolutePath) &&
       this._testSkipAuthored(urlUtils.absolutePathToFileUrl(source.absolutePath))
     ) {
       this.setIsUrlBlackboxSkipped(url, true);
--- a/src/adapter/sources.ts
+++ b/src/adapter/sources.ts
@@ -1082,9 +1082,10 @@ export class SourceContainer {
     const todo: Promise<unknown>[] = [];
     for (const url of map.sources) {
       const absolutePath = await this.sourcePathResolver.urlToAbsolutePath({ url, map });
-      const resolvedUrl = absolutePath
-        ? utils.absolutePathToFileUrl(absolutePath)
-        : map.computedSourceUrl(url);
+      const resolvedUrl =
+        absolutePath && utils.isAbsolute(absolutePath)
+          ? utils.absolutePathToFileUrl(absolutePath)
+          : map.computedSourceUrl(url);
 
       const existing = this._sourceMapSourcesByUrl.get(resolvedUrl);
       if (existing) {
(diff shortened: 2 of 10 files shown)


FrankEssenberger:
> @microsoft-github-policy-service agree [company="{your company}"]

@microsoft-github-policy-service agree [company="SAP"]

FrankEssenberger:
@microsoft-github-policy-service agree company="SAP"

FrankEssenberger:
@connor4312 I added a [exception to find](https://github.com/microsoft/vscode-js-debug/pull/1427/commits/5f4195d47757312c25f558239b3d706ae7240cd2) cases where the method `absolutePathToFileUrl` was called with some URLs. I wonder if I should also find and add checks for this?

FrankEssenberger:
Ok, I found out that the VSC experience is much better compared to webstorm my default IDE. I wanted to test the flow E2E and replaced the source root also with `https://raw.githubusercontent.com/SAP/cloud-sdk-js/main/packages/util/src` and I saw a new assertion with the source from our repo. So far so good. 

However, when the config `extension` and use the [sample repo](https://github.com/FrankEssenberger/source-map-tests) it does not work and I get an error with a better URL 

```
Could not load source '/https:/raw.githubusercontent.com/SAP/cloud-sdk-js/main/packages/util/src/string.ts': Unable to retrieve source content.
```

but I have no Idea why it is not working. When I debug in the repo the URL is correct and source code can be fetched. Is there some setup needed?

FrankEssenberger:
I spend s bit of time to find out why it is working in the repo but not with the separate repo. I debugged the [debugAdapter.ts](https://github.com/microsoft/vscode-js-debug/blob/7594d05518df6700df51771895fcad0ddc7f92f9/src/adapter/debugAdapter.ts#L316-L326) because some errors sounded like they could lead to the error I see in the manual test using the `Extension` launch and the separate repo. However, using the same URL as source route the content is just downloaded and I see it:

image

So there must be something different with the two cases but I have no idea how to debug the `Extension` launch. Perhaps you can give me some hint? @connor4312 

Best
Frank

connor4312:
Looking at this this week. I think there should be a slightly wider change here to prefer URIs to absolute paths. Unfortunately DAP is only path-based and doesn't know about URIs, but for example `urlToAbsolutePath` should be renamed if it doesn't always return absolute paths, and instead we should always return URIs (instead of requiring the caller to check what it received).

I've started making changes based on this PR and hope to have this merged on or by Friday.

FrankEssenberger:
@connor4312 cool that you take this over. I was afraid about bigger refactoring since it is not my code base :-). I was also lacking the deeper understanding and some tests seems flaky on my machine or in the pipeline. Anyway would be great if this works so that we can improve the debug experience without increasing the size of the shipped artifact.

FrankEssenberger:
Perhaps we can mention your PR once it is there for reference and close this one here.

uplight-dev:
Is there anyone reviewing this?

PR #1463, opened 2022-11-29
chore: allow disabling gh release to unblock publishing

connor4312:
Some of our tokens changed which broke automated GH releases. I will toggle it off for this iteration and create the release manually in order to get js-debug out.

### DIFF
--- a/.ci/publish.yml
+++ b/.ci/publish.yml
@@ -16,6 +16,10 @@ parameters:
     displayName: 🚀 Publish Extension
     type: boolean
     default: false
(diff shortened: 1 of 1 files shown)


roblourens (approved):
Why do you publish github releases anyway?

connor4312:
> Why do you publish github releases anyway?

For OSS builds, we cannot contact the marketplace for legal reasons, but we still want to be able to pull built-in extensions. Therefore, we publish the vsix to Github releases and download from there if there's not a marketplace URL configured.

PR #1464, opened 2022-12-02
Bump the timeout of setting the auto attach child processes env vars in extensionHost debugging

### LINKED ISSUE #1374: "autoAttachChildProcesses": true doesn't work in Windows with "type": "extensionHost"
**Describe the bug**

**To Reproduce**
1. Create a new VS Code "Extension - TS" project.
2. Modify its "launch.json" to include `"autoAttachChildProcesses": true`.
3. Add launching a long running node child process via `child_process` to the extension.
4. The child process won't be debugged. In fact, any node child process started by the VS Code Extension Development Host won't be debugged.

Such process are debugged correctly when I do the same in macOS 12.5.1, but doesn't work in either Windows 10 or 11.

**Log File**

[vscode-debugadapter-e946278c.json.gz](https://github.com/microsoft/vscode-js-debug/files/9424923/vscode-debugadapter-e946278c.json.gz)

**VS Code Version:** 1.70.2

**Additional context**
We are using this flag to be able to debug a node server that we spawn from our extension, this also applies to LSP, debug adapters that are implemented in node so that you are able to easily debug them.

segevfiner:
This might still race if the extension host starts proccesses fast enough, but that's a separate issue

Fixes #1374

### DIFF
--- a/src/targets/node/extensionHostAttacher.ts
+++ b/src/targets/node/extensionHostAttacher.ts
@@ -149,7 +149,7 @@ export class ExtensionHostAttacher extends NodeAttacherBase<IExtensionHostAttach
       { openerId: targetId },
     );
 
-    for (let retries = 0; retries < 5; retries++) {
+    for (let retries = 0; retries < 200; retries++) {
       const result = await cdp.Runtime.evaluate({
         contextId: 1,
         returnByValue: true,


segevfiner:
cc @connor4312

OmerRosenbaum:
This resolved the issue for me!

connor4312 (approved):

connor4312:
Thanks for the PR and confirmation 🙂 This is too late to get into next week's stable release, but will be on [nightly](https://github.com/microsoft/vscode-js-debug#nightly-extension) this evening.

rzhao271 (approved):

PR #1467, opened 2022-12-06
[wasm] Implementing support to supportsSetSymbolOptions

thaystg:
Adding support to setSymbolsOptions when debugging a webassembly app

Related to: https://github.com/dotnet/runtime/pull/79284


connor4312:
It looks like the `initialize` test needs to be updated. You can do this most easily by just deleting the expected initialize.txt file and re-running the test locally to output the snapshot.

connor4312 (approved):

Issue #1433, opened 2022-10-20
Debug console very slow when debugging new module-ified/bundled TypeScript compiler


jakebailey:
**Describe the bug**

In TypeScript 5.0, we're changing our codebase to be implemented as modules, which are then bundled. When debugging our codebase, the debug console is very, very slow; completions, evaluation, etc, take many many seconds.

**To Reproduce**
1. Clone https://github.com/jakebailey/TypeScript/tree/typeformer-2
2. Copy `launch.template.json` to `launch.json` in `.vscode`
3. Set a breakpoint inside `createTypeChecker` in `checker.ts`
4. Open `2dArrays.ts`
5. Launch the mocha tests
6. In the debug console, try and use the `ts` variable. On my machine, this took ~4 seconds. Evan evaluating a global like `String` takes that amount of time.

**Log File**

[vscode-debugadapter-ba3a87b5.json.gz](https://github.com/microsoft/vscode-js-debug/files/9832626/vscode-debugadapter-ba3a87b5.json.gz)

**VS Code Version:** 1.72.2

**Additional context**

I have tried `"sourceMapRenames": false,` but didn't see any change.

PR #1473, opened 2022-12-08
fix: improve repl/variable/completion performance for TS

connor4312:
Fixes https://github.com/microsoft/vscode-js-debug/issues/1433

There were a couple problems:

- TS uses a `customDescriptionGenerator`. Previously this was being invoked on each individual property. I've merged this with the logic we added for custom `toString()` representations, so that we run it only once for an object and return a map of its properties.
- TS has thousands of variables in each scope. We call `Runtime.getProperties` to get scope variables to use in completions. It seems like getProperties, which does not have any 'limit' parameter, is blocking. We still need this to do sensible completions, but now each stacktrace caches its completions instead of getting new ones on every request.

With these changes, there may still be a little REPL stall on the first evaluation (until the first completions call finishes, which takes 3-5s on my macbook), but after that it should be appropriately snappy.

### DIFF
--- a/src/adapter/templates/getStringyProps.ts
+++ b/src/adapter/templates/getStringyProps.ts
@@ -28,7 +28,7 @@ export const getStringyProps = templateFunction(function (
           continue;
         }
       } catch (e) {
-        out[key] = `${e} (couldn't describe ${key})`;
+        out[key] = `<<indescribable>>${JSON.stringify([String(e), key])}`;
         continue;
       }
     }
@@ -58,7 +58,7 @@ export const getToStringIfCustom = templateFunction(function (
         return String(repr);
       }
     } catch (e) {
-      return `${e} (couldn't describe object)`;
+      return `<<indescribable>>${JSON.stringify([String(e), 'object'])}`;
     }
   }
 
--- a/src/adapter/templates/getStringyProps.ts
+++ b/src/adapter/templates/getStringyProps.ts
@@ -2,23 +2,42 @@
  * Copyright (C) Microsoft Corporation. All rights reserved.
  *--------------------------------------------------------*/
 
-import { remoteFunction } from '.';
+import { templateFunction } from '.';
 
 /**
  * Gets a mapping of property names with a custom `.toString()` method
  * to their string representations.
  */
-export const getStringyProps = remoteFunction(function (this: unknown, maxLength: number) {
+export const getStringyProps = templateFunction(function (
+  this: unknown,
+  maxLength: number,
+  customToString: (defaultRepr: string) => unknown,
+) {
   const out: Record<string, string> = {};
+  const defaultPlaceholder = '<<default preview>>';
   if (typeof this !== 'object' || !this) {
     return out;
   }
 
   for (const [key, value] of Object.entries(this)) {
+    if (customToString) {
+      try {
+        const repr = customToString.call(value, defaultPlaceholder);
+        if (repr !== defaultPlaceholder) {
+          out[key] = String(repr);
+          continue;
+        }
+      } catch (e) {
+        out[key] = `${e} (couldn't describe ${key})`;
+        continue;
+      }
+    }
+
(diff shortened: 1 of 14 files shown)


connor4312:
Updated test fixtures to make tests happy. The issue was previously we showed the custom description in the `[[Prototype]]` of an object. But this doesn't exist (reliably/safely) at runtime, so the new approach doesn't get it. I think that's fine, perhaps even desirable.

src/adapter/variableStore.ts
```diff
@@ -1027,14 +993,28 @@ class VariablesMap {
...
 
   constructor(
     @inject(IRenameProvider) private readonly renameProvider: IRenameProvider,
     @inject(ICdpApi) private readonly cdp: Cdp.Api,
     @inject(IDapApi) private readonly dap: Dap.Api,
     @inject(AnyLaunchConfiguration) private readonly launchConfig: AnyLaunchConfiguration,
     private readonly locationProvider: IVariableStoreLocationProvider,
-  ) {}
+  ) {
+    this.contextSettings = {
```
connor4312, line 1005: Previously we were code-genning every time we made a preview for a variable. Fortunately Acorn is pretty speedy, but we can precompute that instead easily :)

src/adapter/templates/getStringyProps.ts
```diff
@@ -2,31 +2,66 @@
...
   for (const [key, value] of Object.entries(this)) {
+    if (customToString) {
+      try {
+        const repr = customToString.call(value, defaultPlaceholder);
+        if (repr !== defaultPlaceholder) {
+          out[key] = String(repr);
+          continue;
+        }
+      } catch (e) {
+        out[key] = `${e} (couldn't describe ${key})`;
```
Tyriar, line 31: Does this need localization?

src/adapter/templates/getStringyProps.ts
```diff
@@ -2,31 +2,66 @@
...
+) {
+  if (customToString) {
+    try {
+      const defaultPlaceholder = '<<default preview>>';
+      const repr = customToString.call(this, defaultPlaceholder);
+      if (repr !== defaultPlaceholder) {
+        return String(repr);
+      }
+    } catch (e) {
+      return `${e} (couldn't describe object)`;
```
Tyriar, line 61: Does this need localization?

src/adapter/variableStore.ts
```diff
@@ -461,19 +432,13 @@ class VariableContext {
...
   ): Promise<{ result?: Cdp.Runtime.RemoteObject; errorDescription?: string }> {
     try {
       const customValueDescription = await this.cdp.Runtime.callFunctionOn({
         objectId: object.objectId,
-        functionDeclaration: extractFunctionFromCustomDescriptionGenerator(
-          parameterNames,
-          codeToEvaluate,
-          catchAndReturnErrors,
-        ),
+        functionDeclaration: functionDeclaration,
```
Tyriar, line 441:
```suggestion
        functionDeclaration,
```

src/adapter/variableStore.ts
```diff
@@ -742,25 +707,26 @@ class ObjectVariable extends Variable implements IMemoryReadable {
...
+          functionDeclaration: getToStringIfCustom.decl(
+            `${customStringReprMaxLength}`,
+            this.context.customDescriptionGenerator || 'null',
+          ),
           objectId: this.remoteObject.objectId,
           returnByValue: true,
         });
-        if (result.value) {
-          return (this.customStringRepr = result.value);
+        if (result?.result.value) {
```
Tyriar, line 718: `result.result` is a bit ugly, does a better name exist for the variable?

Tyriar (changes requested):

src/adapter/templates/getStringyProps.ts
connor4312: Eh, probably. This is awkward since this code is executed in the runtime.

connor4312:
flake that I need to fix

PR #1512, opened 2023-01-09
fix: followup for vscode#168013

connor4312:
Fixes https://github.com/microsoft/vscode/issues/168013.

We need to provide a hint for extensionless files that the editor can't guess itself.

### DIFF
--- a/src/adapter/sources.ts
+++ b/src/adapter/sources.ts
@@ -220,7 +220,7 @@ export class Source {
   /**
    * Gets a suggested mimetype for the source.
    */
-  get mimeType(): string | undefined {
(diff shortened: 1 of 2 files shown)


src/adapter/sources.ts
```diff
@@ -217,6 +217,17 @@ export class Source {
     return this._scriptIds;
   }
 
+  /**
+   * Gets a suggested mimetype for the source.
+   */
+  get mimeType(): string | undefined {
```
amunger, line 223: Seems like the name of this getter could convey that it is a guess/suggestion, rather than just relying on comments

amunger (approved):

PR #1532, opened 2023-01-24
fix: skipFiles being slow in remotes

connor4312:
Fixes https://github.com/microsoft/vscode/issues/170412

We were not rebasing remote paths correctly in the initial launch, which caused the fallback logic to kick in and manually apply skipFiles to every single script (and recompile globs), which was quite slow.

And add a test that I wrote while barking up the wrong tree.


src/adapter/scriptSkipper/implementation.ts
```diff
@@ -195,7 +200,7 @@ export class ScriptSkipper {
       return true;
     }
 
-    return this._testSkipAuthored(this._normalizeUrl(url));
```
connor4312, line 198: Just an extraneous call I noticed

roblourens (approved):

PR #1548, opened 2023-02-03
fix: make variable renames scope-aware

### LINKED ISSUE #1542: Variable renamed incorrectly, leading to thrown error
**Describe the bug**

When debugging the TS codebase (which is bundled with esbuild), I periodically hit cases where the variable name I'm using in the debug console doesn't work and ends up throwing.

**To Reproduce**
1. Clone https://github.com/microsoft/TypeScript, npm install, etc
2. Set breakpoint on the line after `const links = ...` in `getTypeOfVariableOrParameterOrProperty` in checker.ts.
3. With `.vscode/launch.template.json` copied to `.vscode/launch.json`, open `2dArrays.ts` and run the test via the "mocha tests" launcher.
4. Once the breakpoint is hit, open the debug console and type `links`.
5. Note the stack trace which mentions `links2`.
6. Disable source mapped stepping; note the variable is called `links`, not `links2`, and now the debug console works.

**Log File**

[vscode-debugadapter-07d5029d.json.gz](https://github.com/microsoft/vscode-js-debug/files/10561633/vscode-debugadapter-07d5029d.json.gz)

**VS Code Version:** 1.74.3 with debugger v2023.1.3017

**Additional context**

Also, the error that's thrown bypasses #1259, somehow; the error is huge.

connor4312:
Fixes #1542

Prototype code. It works, but I want to move parsing blocks to a worker, since parsing large files (like Typescript's `run.js`) takes about a second, which will stall the extension host.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This changelog records changes to stable releases since 1.50.2. "TBA" changes he
 - fix: typeerror for users of vsDebugServer.bundle.js ([#1502](https://github.com/microsoft/vscode-js-debug/issues/1502))
 - fix: expansion of non-primitive getters not working ([#1525](https://github.com/microsoft/vscode-js-debug/issues/1525))
 - fix: support rich ANSI output for complex logs ([vscode#172868](https://github.com/microsoft/vscode/issues/172868))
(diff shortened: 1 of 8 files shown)


connor4312:
Closing in favor of https://github.com/microsoft/vscode-js-debug/pull/1827

PR #1576, opened 2023-02-24
fix: repl stacktrace with renames showing too much info

### LINKED ISSUE #1259: Stack trace from throw in debug console should hide everything below "eval"

connor4312:
Fixes #1259

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ This changelog records changes to stable releases since 1.50.2. "TBA" changes he
 
 ## Nightly (only)
 
-Nothing (yet)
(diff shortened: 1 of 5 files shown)


chrmarti (approved):

jonesk7734 (commented):
Working

PR #1580, opened 2023-02-28
fix: unbound breakpoints in sourcemaps on Chrome 112

### LINKED ISSUE #1567: unbound breakpoints only when putting the breakpoints before debugging
**Describe the bug**
Whenever I put the breakpoints at any random line, as long as it's before debugging starts running the breakpoints become unbound. But if I put a breakpoints after I start debugging and my source code is loaded to the loaded scripts, it's working normally.

**To Reproduce**
1. set a breakpoint at any lines before starting the debugger
2. start debugging
3. See error
The web page keeps loading indefinitely
**Log File**
[vscode-debugadapter-6a08325a.json.gz](https://github.com/microsoft/vscode-js-debug/files/10791553/vscode-debugadapter-6a08325a.json.gz)

**VS Code Version:** 1.75.1

**Additional context**

connor4312:
Chrome Dev (112) no longer has any stackframes for instrumentation
breakpoints, which was resulting in a thrown error.

Fixes #1567

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This changelog records changes to stable releases since 1.50.2. "TBA" changes he
 
 - fix: repl stacktrace with renames showing too much info ([#1259](https://github.com/microsoft/vscode-js-debug/issues/1259#issuecomment-1409443564))
 - fix: recursive source map resolution parsing ignored locations ([vscode#169733](https://github.com/microsoft/vscode/issues/169733))
+- fix: unbound breakpoints in sourcemaps on Chrome 112 ([#1567](https://github.com/microsoft/vscode-js-debug/issues/1567))
 
 ## v1.76 (February 2023)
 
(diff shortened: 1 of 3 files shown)


joyceerhl (approved):

joj:
@connor4312 does this mean that any old version of VS and VSCode will fail with newer versions of Chrome? Should we let people in Chromium know about this? This feels like something that has the potential of breaking every IDE that supports CDP, not just ours :)

connor4312:
> does this mean that any old version of VS and VSCode will fail with newer versions of Chrome

Yes. It would break any client who uses instrumentation breakpoints _and_ assumes (without checking) that there is at least one call frame on every pause event. I don't know how widely used this pattern is, but I sent a mail to someone on their team whom I've talked to recently regarding instrumentation pauses.

connor4312:
Fyi, Chrome has rolled this back in 112 https://bugs.chromium.org/p/chromium/issues/detail?id=1421661

Issue #1582, opened 2023-03-02
Breakpoints are not bound for code in iframe when the iframe is reloaded.


hediet:
**Describe the bug**
Breakpoints are not bound for code in iframe when the iframe is reloaded.

**To Reproduce**
1. clone the monaco editor repository, run `npm install` and open the folder in VS Code
2. Run the debug configuration "Monaco Editor Playground"
3. Reload the preview, by clicking on the reload button:

![image](https://user-images.githubusercontent.com/2931520/222451097-680c1d5f-554a-49bf-a723-e53fc357d3b0.png)

4. Put a breakpoint in `resolveCompletionItem` (`typescript/languageFeatures.ts`)

![image](https://user-images.githubusercontent.com/2931520/222451224-2eace6b9-da60-4ed4-90de-ea981a0f13ef.png)

5. Notice that the breakpoint does not get bound.

![image](https://user-images.githubusercontent.com/2931520/222451428-ec136e89-bde8-4ed3-aede-3c4aedc63b1f.png)

6. Reload the preview again
7. Notice the breakpoint is now bound.
8. Set another breakpoint - notice it does not get bound again

![image](https://user-images.githubusercontent.com/2931520/222451570-dd2c90e8-622f-4062-8182-b0e3e457faa8.png)

**Log File**

[vscode-debugadapter-6ab9f902.json.gz](https://github.com/microsoft/vscode-js-debug/files/10872115/vscode-debugadapter-6ab9f902.json.gz)

**VS Code Version:** Latest

hediet:
Thanks a lot! Just tried it - it works now :)

PR #1585, opened 2023-03-03
fix: assorted bad source behaviors when reloading a page

connor4312:
Fixes https://github.com/microsoft/vscode-js-debug/issues/1582

We weren't dealing with removed execution contexts at all (e.g. removing an iframe or worker), only cleared ones (e.g. refreshing the entire page). Now, removing an execution context properly removes its sources, and scripts from multiple contexts are handled.

I think this is also an underlying fix for several other bugs I have open; will investigate tomorrow.

### DIFF
--- a/src/adapter/breakpoints.ts
+++ b/src/adapter/breakpoints.ts
@@ -26,17 +26,17 @@ import { PatternEntryBreakpoint } from './breakpoints/patternEntrypointBreakpoin
 import { UserDefinedBreakpoint } from './breakpoints/userDefinedBreakpoint';
 import { DiagnosticToolSuggester } from './diagnosticToolSuggester';
 import {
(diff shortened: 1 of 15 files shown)


src/adapter/breakpoints.ts
```diff
@@ -610,7 +609,7 @@ export class BreakpointManager {
     await Promise.all(
       result.unbound.map(b => {
         this._byDapId.delete(b.dapId);
-        b.disable();
+        return b.disable();
```
connor4312, line 612: this fixes a test flake that's failed like 1 in 500 runs for the past two years

joyceerhl (approved):

joyceerhl (approved):

hediet:
Noice

PR #1590, opened 2023-03-04
fix: ensure debug console separates output lines

dancrumb:
Debug console output sometimes gets clumped together when capturing stdout/err instead of console calls since data is sent to DAP as soon as it arrives.

This change changes the behavior do that output is sent to DAP on newlines.

It also looks out for ETX characters and, if it finds those, it switches to sending output to DAP on each ETX. This allows for better handling of multiline log entries.

Fixes https://github.com/microsoft/vscode/issues/175763


dancrumb:

PR #1613, opened 2023-03-23
fix: ipv4 -> ipv6 fallback

Franiac:
Fixes https://github.com/microsoft/vscode/issues/167353

Fixes error in https://github.com/microsoft/vscode-js-debug/pull/1475


connor4312:
Have you tested that this fixes the issue? The original code before your change looks correct to me

Franiac:
I am super sorry, I misread the code, that's embarrasing 😞 

Can you tell when VSCode will utilize a fixed version of vscode-js-debug?

PR #1619, opened 2023-03-29
refactor: improve and order handling in the `Binder`

connor4312:
Previously the `Binder` was a little messy. The binder managed the "tree" of debug sessions, but teardown was not handled in a particular order. So, for example, when terminating a Node.js session with a bunch of child processes, the parent process could be terminated and disconnected while its children were nominally still "running" if their async shutdown logic didn't finish yet.

Now that things are nicely ordered, I can fix this one https://github.com/microsoft/vscode/issues/173993

### DIFF
--- a/src/binder.ts
+++ b/src/binder.ts
@@ -16,7 +16,7 @@ import { DisposableList } from './common/disposable';
 import { EventEmitter, IDisposable } from './common/events';
 import { ILogger, LogTag, resolveLoggerOptions } from './common/logging';
 import { MutableLaunchConfig } from './common/mutableLaunchConfig';
-import { mapValues, once } from './common/objUtils';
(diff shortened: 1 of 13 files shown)


src/binder.ts
```diff
@@ -30,173 +31,190 @@ import DapConnection from './dap/connection';
...
   }
 
-  private async _disconnect() {
-    if (!this._launchers) {
-      return;
+  /**
+   * Disconnects all running targets. Resolves when all have disconnected.
+   */
+  private async _disconnectRoot() {
+    if (this._root.state < TargetState.Terminating) {
```
rzhao271, line 210: The "<" seems off to me. I would have expected that terminating is a "stronger state" or ">=" to disconnecting.

rzhao271 (approved):
LGTM. Ignore my comment earlier.

PR #1622, opened 2023-03-29
Re-add code to download loc files in gulpfile.js for VS build

EricCornelson:


gulpfile.js
```diff
@@ -17,6 +17,10 @@ const util = require('util');
 const deepmerge = require('deepmerge');
 const esbuild = require('esbuild');
 const esbuildPlugins = require('./src/build/esbuildPlugins');
+const signale = require('signale');
+const got = require('got').default;
+const unzipper = require('unzipper');
```
connor4312, line 22: It looks like this is missing in the package.jons.

Also, I would prefer using jszip over unzipper; we recently moved to jszip due to issues and CG alerts on the seemingly-abandoned unzipper package.

Usage of jszip in vscode-test: https://github.com/microsoft/vscode-test/blob/83f719c7364c316ae1736e75ef1a505ad88461f1/lib/download.ts#L278-L294

gulpfile.js
```diff
@@ -27,6 +31,9 @@ const buildDir = 'dist';
 const buildSrcDir = `${buildDir}/src`;
 const nodeTargetsDir = `targets/node`;
 
+const distDir = 'dist';
+const distSrcDir = `${distDir}/src`;
```
connor4312, line 35: These are defined as buildDir and buildSrcDir already

gulpfile.js
```diff
@@ -278,6 +285,40 @@ gulp.task('package:createVSIX', () =>
...
+        const match = /vscode-language-pack-(.*?)\/.+ms-vscode\.js-debug.*?\.i18n\.json$/.exec(
+          entry.path,
+        );
+        if (!match) {
+          return entry.autodrain();
+        }
+
+        const buffer = new streamBuffers.WritableStreamBuffer();
+        const locale = match[1];
+        entry.pipe(buffer).on('finish', () => {
```
connor4312, line 303: You can use the `pipeline` method of streams, as in the code example I linked above, instead of manual pipes 🙂

connor4312 (dismissed):

connor4312:
Thanks! Code changes look good, but it looks like you `npm install`'d with the devdiv packages registry, instead of public npm, which causes the build to fail. You should be able to fix this by resetting the change to package-lock.json (`git checkout main package-lock.json`), commenting out any registry info in your `~/.npmrc`, and npm-installing again.

connor4312 (approved):

Issue #1633, opened 2023-04-01
Command "chrome.exe --headless --user-data-dir=${USER_DATA_DIR}" does NOT work if USER_DATA_DIR comes as or with any parent as directory junction


NotAndOr:
**Describe the bug**
Command "chrome.exe --headless --user-data-dir=${USER_DATA_DIR}" does NOT work if USER_DATA_DIR comes as or with any parent as [directory junction](https://learn.microsoft.com/en-us/windows/win32/fileio/hard-links-and-junctions), which may fail Microsoft Edge DevTools.

**To Reproduce**
1. [Enable portable mode of VS Code](https://code.visualstudio.com/docs/editor/portable) by create folder "data" as directory junction rather than real folder.
2. Open a folder with html file inside, for example a local copy of folder "demo-to-do" from [Microsoft Edge Demos](https://github.com/MicrosoftEdge/Demos/).
3. Try to [start Edge DevTools by right-clicking an HTML file](https://learn.microsoft.com/en-us/microsoft-edge/visual-studio-code/microsoft-edge-devtools-extension/get-started-right-click-html), for example file index.html in folder "demo-to-do".
![Snipaste_2023-04-01_16-08-43](https://user-images.githubusercontent.com/2784615/229274202-9772331e-4b23-45dc-9e17-bec1bf9a0ce2.png)
4. See what pop up are NOT Edge DevTools tab and Edge DevTools: Browser tab,
![Edge DevTools tab and Edge DevTools: Browser tab](https://learn.microsoft.com/en-us/microsoft-edge/visual-studio-code/microsoft-edge-devtools-extension/get-started-images/devtools-extension-v211.png)
BUT error dialog box.
![Snipaste_2023-04-01_16-09-26](https://user-images.githubusercontent.com/2784615/229274897-f4ce8548-a7b5-4bda-a90c-b9bc32dee722.png)

**Log File**
Not available.

**VS Code Version:**  1.76.2

**Additional context**
1. [Generating launch.json](https://learn.microsoft.com/en-us/microsoft-edge/visual-studio-code/microsoft-edge-devtools-extension/open-devtools-and-embedded-browser#click-the-generate-launchjson-button) and invoking "Launch Edge Headless and attach DevTools" results in same error.
1. Invoking "Launching Edge and attach DevTools" workarounds this error.
1. It is reasonable to conclude that argument --headless do matters.
1. Switching to "pwa-chrome" in launch.json results in same as "pwa-msedge".
[...]

PR #1656, opened 2023-04-12
Launch browser with argument user-data-dir specified without directory junction.

NotAndOr:
It fixes [#1633](#https://github.com/microsoft/vscode-js-debug/issues/1633) on OS Windows, let me know if there is some introduced impact not taken into my account, for example some unanticipated side effect on OS other than Windows.


src/targets/browser/browserLauncher.ts
```diff
@@ -98,17 +98,25 @@ export abstract class BrowserLauncher<T extends AnyChromiumLaunchConfiguration>
...
+          this.storagePath,
+          runtimeArgs?.includes('--headless') ? '.headless-profile' : '.profile',
+        ),
       );
     }
 
-    try {
-      fs.mkdirSync(this.storagePath);
-    } catch (e) {}
+    if (!fs.existsSync(this.storagePath)) {
```
connor4312, line 111: The `if` checks should not be necessary for recursive `mkdirSync`

NotAndOr: Oh, yes, I have tested and confirmed that error EEXIST can be thrown by mkdirSync only when non-recursive, so existsSync is not necessary.
That commit has been amended.

connor4312 (approved):

aeschli (approved):

PR #1701, opened 2023-05-06
fix: solving the problem of escaped backslashes in regular expressions.

JohnhanLiu:
## Fix
backslash which in regexp SHOULD NOT be escaped as a string.

## Reproduce Sample Code
```js
const regArr = [/hello\w+/];
const reg0 = regArr[0];
debugger;
```

## Incorrect:
image

## Correct
image

## After fixed
image

### DIFF
--- a/src/adapter/objectPreview/index.ts
+++ b/src/adapter/objectPreview/index.ts
@@ -61,6 +61,9 @@ export function previewRemoteObject(
     context,
     valueFormat,
   );
+
+  if (object.preview?.subtype === 'regexp') return result;
+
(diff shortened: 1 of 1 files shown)


JohnhanLiu:
@connor4312 PTAL :)

JohnhanLiu:
Hi, @connor4312, sorry to mention again. 

Could you take a look and give me a little advice on this PR. In my opinion, it would be great to solve this problem ASAP.

connor4312 (approved):

Tyriar:
@connor4312 CI failing?

PR #1705, opened 2023-05-11
chore: 'hoist' package to reduce pipeline variance

connor4312:
This adds a new build mode where the built assets will replace those in the repo. This can be used in pipelines to avoid having the "workingDir" that doesn't exist until later.

I also looked at modifying the checked-in package.json, but this was difficult because:

(a) we modify the _name_ for nightly builds, which includes local
    development, and having this change based on the build is noisy.
(b) there's just a lot of package.json (about 13k lines) which
    uncompressed is sizable

### DIFF
--- a/.ci/publish-nightly.yml
+++ b/.ci/publish-nightly.yml
@@ -17,7 +17,6 @@ parameters:
 extends:
   template: azure-pipelines/extension/pre-release.yml@templates
   parameters:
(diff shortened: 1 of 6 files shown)


connor4312:
the nightly 'failure' is just a test build where publishing as disabled

joyceerhl (approved):

PR #1716, opened 2023-06-07
fix: Blazor sources sometimes being missing

connor4312:
Ref https://github.com/dotnet/runtime/issues/86754

A true fix is being applied in .net, but this will take time to trickle down.


src/adapter/threads.ts
```diff
@@ -1412,18 +1412,19 @@ export class Thread implements IVariableStoreLocationProvider {
     // Hack: Node 16 seems to not report its 0th context where it loads some
     // initial scripts. Pretend these are actually loaded in the 2nd (main) context.
     // https://github.com/nodejs/node/issues/47438
-    if (
```
rzhao271, line 1415: Is this block and/or the comment still needed?

Issue #1719, opened 2023-06-08
Breakpoints for worker bind after the worker has already started


rchiodo:
**Describe the bug**
If your worker map file is too large, breakpoints won't bind until after the worker is already running

**To Reproduce**
The only steps I have to reproduce is to use our real product. I'm running a worker thread that contains our entire language server and the breakpoints for the worker thread bind on a time delay after the worker has actually started.

Meaning I can't debug startup of the worker thread.

The breakpoints do bind later, and I can debug messages sent to the server (as I can stick a breakpoint in the client).

**Log File**

[vscode-debugadapter-d1089d59.json.gz](https://github.com/microsoft/vscode-js-debug/files/11692734/vscode-debugadapter-d1089d59.json.gz)

**VS Code Version:** 1.79.0-insiders

**Additional context**

roblourens:
Hey @rchiodo, is this working for you now?

rchiodo:
Yep, working good. I used it everyday now.

PR #1723, opened 2023-06-09
fix: sourcemap predictor not filtering nested session on windows

connor4312:
Fixes https://github.com/microsoft/vscode-js-debug/issues/1719


connor4312:
❌ is a flake

lramos15 (approved):

Issue #1752, opened 2023-07-09
Process picker broken on nixOS


thenbe:
**Describe the bug**
Similar to https://github.com/microsoft/vscode-js-debug/issues/1747. Caused by absolute path to `ps`.

```
Error listing processes: spawn /bin/ps ENOENT
```

![2023-07-08-19-07-35](https://github.com/microsoft/vscode-js-debug/assets/33713262/5becbe40-50b9-4b08-b33e-12bc29b4c9cf)

Issue #1747, opened 2023-07-07
Child process not terminated


thenbe:
**Describe the bug**
The top process is killed, but it's immediate *child* is not killed - it's just re-parented.

## `htop` output

Before stopping:

![2023-07-07-15-32-51](https://github.com/microsoft/vscode-js-debug/assets/33713262/6f2da84e-261a-4821-bcfe-8612877435c9)

```bash
│  │  └─ /nix/store/l43bdmxxlkrzwic3q24ik366qyqg3s1g-nodejs-18.16.1/bin/node /nix/store/m82k0dfkwp7m8zyshvrg2pdca67szlgw-pnpm-8.6.5/bin/.pnpm-wrapped vavite-loader vite dev
│  │     └─ node ./node_modules/.bin/../@vavite/node-loader/cli.js vite dev
│  │        └─ node ./node_modules/.bin/../vite/bin/vite.js dev
```

---

After stopping:

![2023-07-07-15-36-06](https://github.com/microsoft/vscode-js-debug/assets/33713262/305e6081-10dc-4aa7-a698-8f724959566b)

```bash
│  │     └─ node ./node_modules/.bin/../@vavite/node-loader/cli.js vite dev
│  │        └─ node ./node_modules/.bin/../vite/bin/vite.js dev
```

---

**To Reproduce**
1. Clone [repro](https://github.com/thenbe/repro-dirty-stop)
1. `pnpm install`
1. In vscode, click the green play button to run the [`launch.json`](https://github.com/thenbe/repro-dirty-stop/blob/master/.vscode/launch.json) config.
1. Click on the stop button.
1. Click the green play button again.
1. Observe error in console `Error: Port 3000 is already in use`

**Log File**

[vscode-debugadapter-a10b2212.json.gz](https://github.com/microsoft/vscode-js-debug/files/11982409/vscode-debugadapter-a10b2212.json.gz)

**VS Code Version:** 1.79.2
**Extension:** ms-vscode.js-debug-nightly v2023.6.2117

**Additional context**

If setting `"killBehavior": "polite"`, the top *two* processes are killed. But the third (`node ./node_modules/.bin/../vite/bin/vite.js dev` seen in output above) is left running.

```
❯ nix-shell -p nix-info --run "nix-info -m"                                                                        04:09:18 
 - system: `"x86_64-linux"`
 - host os: `Linux 6.1.36, NixOS, 23.11 (Tapir), 23.11pre501260.78419edadf0`
[...]

PR #1753, opened 2023-07-09
fix: process picker broken on some Linux distros

thenbe:
Fixes https://github.com/microsoft/vscode-js-debug/issues/1752
Also see https://github.com/microsoft/vscode-js-debug/issues/1747

### DIFF
--- a/src/ui/processTree/darwinProcessTree.ts
+++ b/src/ui/processTree/darwinProcessTree.ts
@@ -43,7 +43,7 @@ export class DarwinProcessTree extends BaseProcessTree {
    * @inheritdoc
    */
   protected createProcess() {
-    return this.spawn('/bin/ps', [
+    return this.spawn('ps', [
       '-xo',
       // The "aaaa" is needed otherwise the command name can get truncated.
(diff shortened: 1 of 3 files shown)


thenbe:

src/ui/processTree/darwinProcessTree.ts
```diff
@@ -43,7 +43,7 @@ export class DarwinProcessTree extends BaseProcessTree {
    * @inheritdoc
    */
   protected createProcess() {
-    return this.spawn('/bin/ps', [
+    return this.spawn('ps', [
```
thenbe, line 46: I'm aware this is the darwin file, but I figured we might as well be consistent.

connor4312 (approved):

thenbe:
Yup I see them. I'll have to take a closer look at them on a mac (on my nix machine, those tests and a bunch more are failing even before this commit).

connor4312:
Thanks!

Tyriar (approved):

lramos15 (approved):

PR #1758, opened 2023-07-13
fix member expression completions

zardoy:
e.g. `arr.len|`, seems it was broken for years

fixes https://github.com/microsoft/vscode/issues/168021

### DIFF
--- a/src/adapter/completions.ts
+++ b/src/adapter/completions.ts
@@ -143,18 +143,25 @@ export class Completions {
     const source = parseProgram(options.expression);
     const offset = new PositionToOffset(options.expression).convert(options.position);
     let candidate: () => Promise<ICompletionWithSort[]> = () => Promise.resolve([]);
+    let parent: Node | undefined = undefined;
 
     traverse(source, {
       enter: node => {
         const asAcorn = node as AcornNode;
         if (asAcorn.start < offset && offset <= asAcorn.end) {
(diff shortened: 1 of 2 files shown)


zardoy:
though there is still no completions for `["|"]` (would be useful for number props)

src/adapter/completions.ts
```diff
@@ -143,18 +143,28 @@ export class Completions {
     const source = parseProgram(options.expression);
     const offset = new PositionToOffset(options.expression).convert(options.position);
     let candidate: () => Promise<ICompletionWithSort[]> = () => Promise.resolve([]);
+    let parent: Node | undefined = undefined;
 
     traverse(source, {
       enter: node => {
```
connor4312, line 149: estraverse actually gives the parent as the second argument to this function call

connor4312 (commented):
Took me a lil to grok but it looks okay. I think the main thing is just using estraverse's parent since the current implementation could incorrectly pass the `parent` as the previous sibling node.

zardoy:
>  I think the main thing is just using estraverse's parent since the current implementation could incorrectly pass the parent as the previous sibling node.

But it is inside of the `asAcorn.start < offset && offset <= asAcorn.end` check. Anyway, it was easy to change. Btw thank you for the quick reply!

connor4312 (approved):

hediet (approved):

PR #1759, opened 2023-07-16
Remove extra bracket from command line copy utility

hbendev:
Excellent project, I love that it just always works out of the box! Thanks! This is just a minor fix where one of the utility outputs were off by a character.

The output was:
```bash
Output has been truncated to the first 100000 characters. Run `copy(object))` to copy the full output.
```

Which just results in a syntax error:
```bash
Uncaught SyntaxError SyntaxError: Unexpected token ')'
```


connor4312 (approved):
Thanks!

chrmarti (approved):

PR #1780, opened 2023-08-07
Speed up sourcemap lookups by checking existance of .map files

sunilsurana:
**Problem**
Debugging experience is slow in large monorepos, since all the files are read to check location of sourcemap file

**PR fix**
Check if .map file exists for compiled file we don't need to read file to find location of sourcemap location. 

**Possible issue**
If there is a js file and it has .map file which is not a sourcemap file with same js file name, it can lead to issue.
This however should be rare edge case.

### DIFF
--- a/src/common/sourceMaps/sourceMapRepository.ts
+++ b/src/common/sourceMaps/sourceMapRepository.ts
@@ -4,11 +4,10 @@
 
 import { xxHash32 } from 'js-xxhash';
 import { FileGlobList } from '../fileGlobList';
-import { readfile, stat } from '../fsUtils';
+import { IFsUtils, readfile, stat } from '../fsUtils';
 import { parseSourceMappingUrl } from '../sourceUtils';
 import { absolutePathToFileUrl, completeUrl, fileUrlToAbsolutePath, isDataUri } from '../urlUtils';
 import { ISourceMapMetadata } from './sourceMap';
-
 /**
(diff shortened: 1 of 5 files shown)


src/common/sourceMaps/sourceMapRepository.ts
```diff
@@ -66,14 +66,20 @@ export interface ISearchStrategy {
...
 export const createMetadataForFile = async (
+  fsUtils: IFsUtils,
   compiledPath: string,
   fileContents?: string,
 ): Promise<Required<ISourceMapMetadata> | undefined> => {
-  if (typeof fileContents === 'undefined') {
-    fileContents = await readfile(compiledPath);
+  let sourceMapUrl;
+  const possibleSourceMapURL = `${compiledPath}.map`;
+  if (await fsUtils.exists(possibleSourceMapURL)) {
```
connor4312, line 75: Rather than making a separate `exists` call here, you can pass the siblings elements in the directory to the `fileProcessor` function.

It's called by `readSomething` which is called for the root entry (no siblings) and in `handleDirectoryEntry`. In the latter case, it's called reading either from the cache tree (where `Object.keys(cache.children)` are siblings) or from the filesystem (where `files` are siblings)

sunilsurana: Updated. I'm not fully sure thou if you meant the same changes

alexr00 (approved):

connor4312 (approved):

Issue #1789, opened 2023-08-24
WASM debugging support


connor4312:
js-debug should be able to debug WebAssembly.

There's two parts of this. The base devtools experience allows users to step into and around WebAssembly, displayed as WAT. Supporting this can be done pretty easily.

The next step is giving source-level support for debugging. The Chromium team recently made their DWARF debugger [open source](https://source.chromium.org/chromium/chromium/src/+/main:third_party/devtools-frontend/src/extensions/cxx_debugging/). While this very sadly doesn't work out of the box for Rust [yet](https://github.com/rustwasm/wasm-bindgen/issues/2389), it should eventually (we can hope), and it does support any other language that outputs DWARF symbols.

While the extension is built for integration in Chrome devtools, it looks like it's architected in a way that pretty much all the work happens in the `DevToolsPluginWorker`. I think that we could package this up into an node module that the extension could consume and run in a Node worker_thread.

The downside is the the DWARF SymbolBackend includes(?) chunks from LLVM and even compressed is pretty hefty at 11MB, so we may not want to have it included in vscode by default like js-debug is. My thought is that we can have an extension that exports a known API for js-debug to consume, alongside an npm package that ships only the _types_ in the package (or maybe I just copy the types directly into js-debug). We can try to both 
`require('vscode').extensions.getExtension('ms-vscode.js-debug-dwarf')` as well as `require('@vscode/dwarf-debugging')` to enable pure DAP consumers to get this too. And of course we can prompt the user to install this extension when they first step into WASM code.

1. [x] Enable basic debugging with WAT
2. [x] Create a module bundling the devtools c/++ worker debugger
3. [x] Integrate the module into js-debug
4. [x] Externalize the module into an extension with suggestions

connor4312:
This is largely done now in PRs #1803 through #1808. Tomorrow I will do some final cleanup and try to get some tests going, and then it will ship in the nightly.

fleabitten232:
Does this currently support the "path substitution" feature from the Chrome plugin? I compile in a container, which means that the source file paths are incorrect for the machine I want to debug on. 

Appreciate all the work that's been done on this and I'm eager to try it out!

connor4312:
The path goes through the usual path mapping logic in js-debug, so for your case you'd want to set the localRoot/remoteRoot in your launch.json as described in https://code.visualstudio.com/docs/nodejs/nodejs-debugging#_remote-debugging

PR #1791, opened 2023-08-25
feat: support basic webassembly debugging

### LINKED ISSUE #1715: [Bug/Question] Breakpoint auto-relocate to other file in V1.78.0
Hey @connor4312 , I'm the maintainer of microsoft/react-native-tools repo. Currently I noticed a behavior from `V1.78.0 built-in`(nightly version is from `v2023.4.2709`), when I debug react-native project using `react-native-tools`(depend on js-debug) extension, put a breakpoint in App.tsx then it auto changes to require.js file. 

I'm not sure it's the issue or design by some updates, could you please help to look into this and give me the guidance? Thank!

https://github.com/microsoft/vscode-js-debug/assets/19662219/e91c1a9d-d2db-4dd2-b9d3-af95d54dacbb

If you need other info, please let me know.

connor4312:
Supports viewing, stepping through, and setting breakpoints in
webassembly (decompiled as WAT) in the editor. Includes a
basic tmLanguage for WAT.

Unfortunately we eagerly have to decompile WASM in order to get line
mappings to show e.g. in breakpoints. Location mapping is currently
mostly synchronous and I didn't want to make everything async
for webassembly. However, we don't keep the WAT source in memory,
instead request it again if it's needed. I opted to do this to reduce
memory usage for user applications that just happen to contain WASM
where they aren't always interested in debugging it.

For #1789
Fixes #1715 on the way


src/adapter/breakpoints.ts
```diff
@@ -510,19 +519,7 @@ export class BreakpointManager {
 
     const wasEntryBpSet = await this._sourceMapHandlerInstalled?.entryBpSet;
     params.source.path = urlUtils.platformPathToPreferredCase(params.source.path);
-
-    // If we see we want to set breakpoints in file by source reference ID but
-    // it doesn't exist, they were probably from a previous section. The
-    // references for scripts just auto-increment per session and are entirely
-    // ephemeral. Remove the reference so that we fall back to a path if possible.
```
connor4312, line 517: This is no longer the case, and caused problems. Since a few years ago, we set the source referenced based on a hash of the URL and other information to try to make them stable between reloads, so we should leverage that information.

roblourens (approved):
Wat... 😁 

I just wonder whether the wat grammar should be in this extension or in the builtin javascript extension, or its own builtin extension

deepak1556 (approved):

connor4312:
I think having it be built-in is sensible. I'll move it over on Monday.

PR #1804, opened 2023-09-15
feat: initial DWARF debugger integration

connor4312:
> This is part 1 in a stacked PR. All other PRs will get merged into here, and then into main, after they're review individually (to avoid a massive 3k line PR review)

Enables stepping around and setting breakpoints in native code.
Variables represenation is still a todo, and this needs some polish.

It refactors the `Source.sourceMap` into a more generic location
provider, which may be backed by the new IWasmSymbols interface. This
then talks to the DWARF debugger. The SourceContainer's `_sourceMaps`
are also now just folded into the `Source.sourceMap`; that information
was duplicate and the only benefit was deduplication if two sources had
the same sourcemap, but this is really rare.

This also made all location mapping asynchronous, so there were a
few refactors around there, too.

It also refactors how I implemented WAT (text format decompilation) last
iteration. That previously "pretended" the source of the WASM was WAT,
but this caused issues because the location transformation we did on the
scripts is done before location mapping. So instead, the WAT is an extra
fake 'file' in the symbols and we resolve any unknown locations into the
WAT file.

One that that _doesn't_ work yet is entrypoint breakpoints for wasm,
which means that breakpoints set before the debug session starts may be
missed. I have a thread out to the Chromium folks to see if there's
a solution to this.

For #1789


connor4312:
Looks like compilation will be unhappy until I publish the @vscode/dwarf-debugging module, which I will do soon, but not yet. Still, I would appreciate a review--I'm planning to make a PR stack and then collapse them once everything's good.

roblourens (dismissed):
Nice!

mjbvz (dismissed):

deepak1556 (approved):

PR #1812, opened 2023-09-21
chore: remove redundant type definitions for del

icholy:
`del@6.0.0` ships its own types https://github.com/sindresorhus/del/blob/v6.0.0/index.d.ts

Log:
```
npm WARN deprecated @types/del@4.0.0: This is a stub types definition. del provides its own type definitions, so you do not need this installed.
```


connor4312 (approved):

joyceerhl (approved):
CI failing @connor4312

PR #1846, opened 2023-10-12
fix: also use script ID bps if there is a different embeddername

connor4312:
For https://github.com/microsoft/vscode/issues/195037\#issuecomment-1759343645


connor4312:
There's a few flakes in there

rzhao271 (approved):

PR #1853, opened 2023-10-17
feat: Tidying Up of Instrumentation Breakpoints

OnesAndZer0s:
"Browser Breakpoints", renamed to "Event Listener Breakpoints", are now organized in a tree-view structure, utilizing the new Treeview-Checkbox API
image


OnesAndZer0s:

.vscode/settings.json
```diff
@@ -17,16 +17,19 @@
...
+    "**/Thumbs.db": true,
+    "**/obj": true,
     "**/.dynamic-testWorkspace": true,
     "**/.vscode-test": true,
-    "**/demos": true,
+    "**/demos": true
   },
   "[typescript]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
+    "editor.defaultFormatter": "vscode.typescript-language-features"
```
connor4312, line 27: This repo is formatted using prettier, please don't change this.

.vscode/settings.json
```diff
@@ -17,16 +17,19 @@
...
   },
   "editor.codeActionsOnSave": {
     "source.organizeImports": true
   },
   "editor.formatOnSave": true,
   "typescript.tsserver.experimental.enableProjectDiagnostics": true,
+  "hide-files.files": [],
```
connor4312, line 34: Accidental?

src/ui/customBreakpointsUI.ts
```diff
@@ -50,39 +74,86 @@ class BreakpointsDataProvider implements vscode.TreeDataProvider<Breakpoint> {
...
+
+    // disable every breakpoint in breakpoints
+    for (const breakpoint of breakpoints)
+      breakpoint.checkboxState = vscode.TreeItemCheckboxState.Unchecked;
+
     for (const session of this._debugSessionTracker.getConcreteSessions())
-      session.customRequest('disableCustomBreakpoints', { ids: breakpointIds });
+      session.customRequest('disableCustomBreakpoints', { ids: breakpoints.map(e => e.id) });
+
+    this.fixCheckState();
```
connor4312, line 136: I don't think you need to do this here, you already handle unchecking parents on L123

src/ui/customBreakpointsUI.ts
```diff
@@ -93,34 +164,92 @@ export function registerCustomBreakpointsUI(
...
-      const quickPick = vscode.window.createQuickPick();
-      const items = provider.breakpoints.filter(b => !b.enabled);
-      quickPick.items = items;
-      quickPick.onDidAccept(() => {
-        provider.addBreakpoints(quickPick.selectedItems as Breakpoint[]);
-        quickPick.dispose();
-      });
-      quickPick.show();
+      const items = provider.breakpointCategories
+        .sort(BreakpointCategory.compare)
```
connor4312, line 199: This mutates the `breakpointCategories` on the provider, which we should avoid. (If you move to a map you get this automatically since this becomes `[...provider.breakpointCategories.values()]`)

You could also use the `QuckPickItemKind.Separator` to retain information about the groups 🙂

src/ui/customBreakpointsUI.ts
```diff
@@ -93,34 +164,92 @@ export function registerCustomBreakpointsUI(
...
+          );
+        });
     }),
   );
 
   context.subscriptions.push(
     vscode.commands.registerCommand(Commands.RemoveAllCustomBreakpoints, () => {
-      provider.removeBreakpoints(provider.breakpoints.filter(b => b.enabled).map(b => b.id));
+      provider.removeBreakpoints(
+        provider.breakpointCategories
```
connor4312, line 224: Wouldn't this only remove categories that have all of their children checked, and not individual children?

src/ui/customBreakpointsUI.ts
```diff
@@ -93,34 +164,92 @@ export function registerCustomBreakpointsUI(
...
-      Commands.RemoveCustomBreakpoint,
-      (treeItem: vscode.TreeItem) => {
-        provider.removeBreakpoints([treeItem.id as string]);
-      },
-    ),
+    vscode.commands.registerCommand(Commands.RemoveCustomBreakpoints, () => {
+      const items = provider.breakpointCategories
+        .sort(BreakpointCategory.compare)
+        .map(bC => bC.children.map(b => ({ id: b.id, label: `${bC.label}: ${b.label}` })))
+        .flat();
```
connor4312, line 237: Can probably have a helper function to make this list, reused in `AddCustomBreakpoints`, with some `filter` function here.

connor4312 (commented):
Thanks for the PR! Some comments 🙂

src/ui/customBreakpointsUI.ts
OnesAndZer0s: True. I put it there just in case

src/ui/customBreakpointsUI.ts
OnesAndZer0s: included in [this commit](https://github.com/microsoft/vscode-js-debug/pull/1853/commits/f66568d03d1fe72d63d4141ffa3f3a8b034206bd)

connor4312:
Can we put XHR breakpoints into a separate PR to keep this one focused on the existing changes?

OnesAndZer0s:
> Can we put XHR breakpoints into a separate PR to keep this one focused on the existing changes?

#1856

src/ui/customBreakpointsUI.ts
```diff
@@ -93,34 +124,94 @@ export function registerCustomBreakpointsUI(
...
+      if (!e.items[0][0].label) return;
+      const allBr = provider.breakpoints.get(e.items[0][0].label.toString());
+      if (!allBr) return;
+      if (e.items[0][1]) {
+        provider.addBreakpoints(allBr);
+      } else {
+        provider.removeBreakpoints(allBr);
+      }
+      return;
+    }
```
connor4312, line 143: I'm still not a fan of special-casing the "if there's exactly one change in this event". VS Code doesn't guarantee that group selections or deselections are presented this way. Instead we could check `if (item instanceof Breakpoint)` in add/removeBreakpoints and do the necessary logic on children there

lramos15 (approved):

PR #1856, opened 2023-10-19
feat: Addition of XHR/feat Breakpoints

OnesAndZer0s:
image


connor4312:
I wonder if we could fold this into the existing Event Listener Breakpoints view to avoid add more UI clutter. Spitballing:

- There should be a special "XHR Breakpoints" category to contain these. The category has an inline `+` action to add a new breakpoint
- If no XHR breakpoints were set yet, the category has a child item to "Add new breakpoint for URL", which opens a quickpick to add the BP
- Child items have a checkbox to enable/disable as well as an inline `X` action to remove them

OnesAndZer0s:
> I wonder if we could fold this into the existing Event Listener Breakpoints view to avoid add more UI clutter.

OnesAndZer0s:
> * There should be a special "XHR Breakpoints" category to contain these. The category has an inline `+` action to add a new breakpoint
I agree, that sounds like a good idea. However, I have some qualms:
- There is a decent change that the "XHR/Fetch Breakpoints" category would get confused with the `xhr` event breakpoint category.
- I do not know if it is able to add a condition to make a `Category` with custom buttons within `generate-contributions`. Maybe something with `when`?

Otherwise, yeah that does sounds nice.

connor4312:
Perhaps we call them "URL XHR Breakpoints"?

> Maybe something with when?

Indeed, ctrl+f for "If you want to show an action for specific tree items" on https://code.visualstudio.com/api/extension-guides/tree-view#view-actions

connor4312:
It seems like the package-lock.json was rewritten, probably with a newer version of npm. Can you undo that change? We use Node 18 and its included version of npm in our pipelines.

package.nls.json
```diff
@@ -1,5 +1,9 @@
 {
   "add.eventListener.breakpoint": "Toggle Event Listener Breakpoints",
+  "add.xhr.breakpoint": "Add XHR/fetch Breakpoint",
+  "breakpoint.xhr.contains":"Break when URL contains:",
+  "breakpoint.xhr.any":"Any XHR/fetch",
+  "edit.xhr.breakpoint": "Edit XHR/fetch Breakpoint",
```
connor4312, line 6: Wonder if just calling them "Fetch Breakpoints" would confuse people. I'm not sure anyone who's learned JavaScript in the last 5 years has ever written an XHR request manually 😛

src/ui/customBreakpointsUI.ts
```diff
@@ -3,15 +3,39 @@
...
   CustomBreakpointId,
   customBreakpoints,
   ICustomBreakpoint,
+  IXHRBreakpoint,
 } from '../adapter/customBreakpoints';
 import { Commands, CustomViews } from '../common/contributionUtils';
 import { EventEmitter } from '../common/events';
 import { DebugSessionTracker } from './debugSessionTracker';
 
+const XHRBreakpointsCategory = 'XHR/Fetch URLs';
```
connor4312, line 17: We should localize this string, and inline where it's used (localization is not set up until the extension's activate() is called, so top-level strings like this wouldn't get translated)

src/ui/customBreakpointsUI.ts
```diff
@@ -125,27 +174,67 @@ class BreakpointsDataProvider implements vscode.TreeDataProvider<Breakpoint | Ca
...
       }
     }
 
     const ids = this.allBreakpoints.filter(b => b.checked).map(b => b.id);
+    const xhr = this.xhrBreakpoints.filter(b => b.checkboxState).map(b => b.id);
+    const xhrDisabled = this.xhrBreakpoints.filter(b => !b.checkboxState).map(b => b.id);
     for (const session of this._debugSessionTracker.getConcreteSessions()) {
       session.customRequest('setCustomBreakpoints', { ids });
+      session.customRequest('enableXHRBreakpoints', { ids: xhr });
+      session.customRequest('disableXHRBreakpoints', { ids: xhrDisabled });
```
connor4312, line 198: Perhaps we make `setCustomBreakpoints` take the `ids` as well as a list of `xhrPatterns: string[]` -- avoid having extra calls and work.

src/ui/customBreakpointsUI.ts
```diff
@@ -206,4 +295,45 @@ export function registerCustomBreakpointsUI(
       provider.setEnabled(provider.allBreakpoints.map(bp => [bp, false]));
     }),
   );
+
+  context.subscriptions.push(
+    vscode.commands.registerCommand(Commands.AddXHRBreakpoints, () => {
+      const inputBox = vscode.window.createInputBox();
+      inputBox.title = 'Add XHR Breakpoint';
+      inputBox.placeholder = 'Enter a URL or a pattern to match';
```
connor4312, line 303: I'm not sure it's clear to the user what pattern means here. Is it a regex, a glob match, something else? Maybe we copy devtools' wording:

image

src/ui/customBreakpointsUI.ts
```diff
@@ -206,4 +295,45 @@ export function registerCustomBreakpointsUI(
...
+  context.subscriptions.push(
+    vscode.commands.registerCommand(Commands.EditXHRBreakpoint, (treeItem: vscode.TreeItem) => {
+      const inputBox = vscode.window.createInputBox();
+      inputBox.title = 'Edit XHR Breakpoint';
+      inputBox.placeholder = 'Enter a URL or a pattern to match';
+      inputBox.value = (treeItem as XHRBreakpoint).match;
+      inputBox.onDidAccept(() => {
+        const match = inputBox.value;
+        provider.removeXHRBreakpoints([treeItem as XHRBreakpoint]);
+        provider.addXHRBreakpoints([new XHRBreakpoint({ match }, treeItem.checkboxState == 1)]);
```
connor4312, line 322:
```suggestion
        provider.addXHRBreakpoints([new XHRBreakpoint({ match }, treeItem.checkboxState == vscode.TreeItemCheckboxState.Checked)]);
```

src/ui/customBreakpointsUI.ts
```diff
@@ -84,39 +120,52 @@ class BreakpointsDataProvider implements vscode.TreeDataProvider<Breakpoint | Ca
...
+  getChildren(
+    item?: Breakpoint | Category | XHRBreakpoint,
+  ): vscode.ProviderResult<(Breakpoint | Category | XHRBreakpoint)[]> {
     if (!item) {
       return [...this.categories.values()];
     }
 
     if (item instanceof Category) {
+      if (item.contextValue === 'xhrCategory') {
+        return this.xhrBreakpoints.sort(XHRBreakpoint.compare);
```
connor4312, line 146: We could also have a tree item in the XHR category that would make a new breakpoint

```ts
const title = l10n.t('Add new URL');
const addNew = new vscode.TreeItem(title);
addNew.command = { title, command: Commands.AddXHRBreakpoints };
return [...this.xhrBreakpoints, addNew];
```

src/ui/customBreakpointsUI.ts
```diff
@@ -125,27 +174,67 @@ class BreakpointsDataProvider implements vscode.TreeDataProvider<Breakpoint | Ca
...
+    for (const session of this._debugSessionTracker.getConcreteSessions())
+      session.customRequest('enableXHRBreakpoints', { ids: match });
+    // if all XHR breakpoints are removed, disable the category
+    if (this.xhrBreakpoints.length != 0) {
+      const category = this.categories.get(XHRBreakpointsCategory);
+      if (category) {
+        category.checkboxState ??= vscode.TreeItemCheckboxState.Checked;
+        category.collapsibleState ??= vscode.TreeItemCollapsibleState.Expanded;
+      }
+    }
```
connor4312, line 218: This and the block below can probably use a single reconciliation method like.

```ts
const category = this.categories.get(XHRBreakpointsCategory);
category.checkboxState = this.xhrBreakpoints.length && this.xhrBreakpoints.every(b => b.checkboxState === vscode.TreeItemCheckboxState.Checked)
  ? vscode.TreeItemCheckboxState.Checked : vscode.TreeItemCheckboxState.Unchecked;
```

...the block in `removeXHRBreakpoints` at least has a bug that if the only disabled URL pattern is removed, then the parent won't get re-checked. Doing the above scan is a little more work (as it would be done on every change) but would fix that.

package.nls.json
OnesAndZer0s: I have seen people use it, and since it is still apart of the Javascript standard, and that Chromium has that labeled as "XHR/Fetch", I labeled it as such. I personally feel like it would confuse people, but that is just me.

src/ui/customBreakpointsUI.ts
```diff
@@ -125,25 +180,68 @@ class BreakpointsDataProvider implements vscode.TreeDataProvider<Breakpoint | Ca
...
+    for (const breakpoint of breakpoints) this.xhrBreakpoints.push(breakpoint);
+    for (const session of this._debugSessionTracker.getConcreteSessions())
+      session.customRequest('setCustomBreakpoints', { xhr: match });
+    const category = this.categories.get(XHRBreakpointsCategory);
+    if (category) {
+      category.checkboxState =
+        this.xhrBreakpoints.length &&
+        this.xhrBreakpoints.every(b => b.checkboxState === vscode.TreeItemCheckboxState.Checked)
+          ? vscode.TreeItemCheckboxState.Checked
+          : vscode.TreeItemCheckboxState.Unchecked;
```
connor4312, line 227: We should use this (probably as a helper like `syncXHRCategoryState`) in the below method too

src/ui/customBreakpointsUI.ts
```diff
@@ -125,25 +180,68 @@ class BreakpointsDataProvider implements vscode.TreeDataProvider<Breakpoint | Ca
...
-          breakpoint.parent.checkboxState = state;
-        } else if (enabled && breakpoint.parent.children.every(c => c.checked)) {
-          breakpoint.parent.checkboxState = state;
+      } else if (breakpoint instanceof Breakpoint || breakpoint instanceof XHRBreakpoint) {
+        const parent = this.getParent(breakpoint) as Category;
+        if (!enabled && parent.checked) {
+          parent.checkboxState = state;
+        } else if (
+          enabled &&
+          (this.getChildren(parent) as XHRBreakpoint[]).every(
```
connor4312, line 194: If we add `parent` and `checked` on the XHRBreakpoint then we can avoid this :)

src/adapter/debugAdapter.ts
```diff
@@ -475,15 +478,48 @@ export class DebugAdapter implements IDisposable {
...
+  //   params: Dap.DisableXHRBreakpointsParams,
+  // ): Promise<Dap.DisableXHRBreakpointsResult> {
+  //   const promises: Promise<void>[] = [];
+  //   for (const id of params.ids) {
+  //     this._xhrBreakpoints.delete(id);
+  //     if (this._thread) promises.push(this._thread.updateXHRBreakpoint(id, false));
+  //   }
+  //   await Promise.all(promises);
+  //   return {};
+  // }
```
connor4312, line 521: 🔥

connor4312 (commented):
looking good, a couple last comments

src/ui/customBreakpointsUI.ts
OnesAndZer0s: `XHRBreakpoint` already has `checked`. What would be avoiding?

OnesAndZer0s:
> It seems like the package-lock.json was rewritten, probably with a newer version of npm. Can you undo that change? We use Node 18 and its included version of npm in our pipelines.

Fixed.

Is there any other suggestions?

connor4312 (approved):

DonJayamanne (approved):

PR #1906, opened 2023-12-13
fix: match mjs and cjs in outFiles by default

connor4312:
Fixes https://github.com/microsoft/vscode/issues/200665

Requires use of my picomatch fork due to https://github.com/micromatch/picomatch/issues/125.
We could alternatively use brace expansion until that's fixed, but that's much less
efficient that using a logical or.


connor4312:
Test failures should be fixed with the test-electron update coming in https://github.com/microsoft/vscode-js-debug/pull/1905

TylerLeonhardt (approved):

PR #1952, opened 2024-02-26
fix: race when stopping on node-internals exceptions

connor4312:
Fixes https://github.com/microsoft/vscode/issues/204581


src/adapter/threads.ts
```diff
@@ -951,6 +957,7 @@ export class Thread implements IVariableStoreLocationProvider {
         // and an instrumentation pause in step out should not be possible.
         if (expectedPauseReason.direction === StepDirection.In) {
           // no-op
+          debugger;
```
rzhao271, line 960: Intentional?

rzhao271 (approved):

Issue #1967, opened 2024-03-14
Fix preview rendering of number '0' when 'description' is not set.


Beanyy:
**Describe the bug**
Some Javascript Debuggee CDP implementations (Hermes) return the a RemoteObject of `type=number` , `value=0` and the `description` and `unserializableValue` fields unset. This causes RemoteObjects with `type=number` and  `value=0` to be rendered  as NaN due to the logic of formatting numbers here:
https://github.com/microsoft/vscode-js-debug/blob/35c4aea8e08be1ab829bf4e4d585433f7e2b7726/src/adapter/objectPreview/index.ts#L477-L484

If param.value is falsy (as is for zero), it uses the description, but if that is unset it will display NaN. 
I'm suggesting making an explicit check for positive zero here and assigniing it to value if the description is not set.

**To Reproduce**

**Log File**

**VS Code Version:** Replace me!

**Additional context**

Beanyy:
> Good find, PR welcome if you'd like to contribute

Created a PR here: https://github.com/microsoft/vscode-js-debug/pull/1968

PR #1968, opened 2024-03-14
Fix formatting of number 0 in remote object when description is not set

Beanyy:
Fix for issue reported here: https://github.com/microsoft/vscode-js-debug/issues/1967

Some Javascript Debuggee CDP implementations (Hermes) return the a RemoteObject of `type=number` , `value=0` and the `description` and `unserializableValue` fields unset. This causes RemoteObjects with `type=number` and  `value=0` to be rendered  as NaN due to the logic of formatting numbers here:
https://github.com/microsoft/vscode-js-debug/blob/35c4aea8e08be1ab829bf4e4d585433f7e2b7726/src/adapter/objectPreview/index.ts#L477-L484

### DIFF
--- a/src/adapter/objectPreview/index.ts
+++ b/src/adapter/objectPreview/index.ts
@@ -479,7 +479,7 @@ function formatAsNumber(
       return param.unserializableValue;
     }
 
-    const value = param.value || +param.description;
+    const value = param.value !== undefined ? param.value : +param.description;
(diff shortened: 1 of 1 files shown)


connor4312 (approved):
Don't worry about the test failures, those are some known flakes I need to fix

deepak1556 (approved):

Issue #2001, opened 2024-04-26
The split2 module is very slow when handling large amounts of CDP data


connor4312:
I noticed in a profile that the split2 module took a ridiculous amount of time when requesting a large source from a debugee:

<img width="439" alt="image" src="https://github.com/microsoft/vscode-js-debug/assets/2230985/5b0369ae-c04e-491c-af85-257cda8ff17c">

Let's replace it with the optimized splitter I wrote for VS Code

andreamah:
Are there verification steps for this?

connor4312:
No good steps really, but folks have been using js-debug-nightly for copilot-chat where this was an issue and it seems to have worked, so I'll mark verified.

PR #2002, opened 2024-04-26
fix: improve protocol handling performance in all cases

connor4312:
Please observe this scientific comparison of rename performance in the copilot chat repo:

image

Closes https://github.com/microsoft/vscode-js-debug/issues/2001


src/test/node/node-runtime.test.ts
```diff
@@ -712,4 +716,46 @@ describe('node runtime', () => {
       handle.assertLog({ substring: true });
     });
   });
+
+  describe('etx', () => {
```
connor4312, line 720: realized we didn't actually have test cases for these before

lramos15 (approved):

roblourens:
#meme or #performance

image

PR #2003, opened 2024-04-26
chore: fixup flakey profile tests

connor4312:


connor4312:
Still one flake it looks like, but the profile flakes are fixed

joyceerhl (approved):

Issue #2051, opened 2024-08-06
Support network inspection


connor4312:
With Node.js [adding network inspection capability](https://github.com/nodejs/node/pull/53593), we should have some kind of inspector for it. It can also of course work for browser inspection too.

connor4312:
Nice, I think it's now good enough to have on by default on 23.11+

PR #2053, opened 2024-08-08
feat: add basic network view, support experimental networking for node

connor4312:
This adds a basic network tree view that shows requests and responses
in the debugee. It has a default "go to" action which opens a cURL
representation of the request, and context menu actions to open the
response body either in a text editor or the hex editor. It also has
actions to copy the URL and replay the request.

I initially was hoping to turn this on by default for Node.js since
their networking support in 22.6.0 inspired this change, but their
functionality right now is [very limited](https://github.com/nodejs/node/pull/53593#issuecomment-2276367389) (we basically get the URL and
nothing else.) Therefore I added an option to turn it on, but it's not
on by default there.

I think I might end up toggling this off for the next stable release
for now until we can polish it some more, but it's ship it in nightly
to get some feedback!

Closes https://github.com/microsoft/vscode-js-debug/issues/2051


connor4312:
build failure is a known flake that's on my todo

lramos15 (approved):

vish01:
How do I enable this feature? I installed latest nightly version but don't see "network" tab in there.

connor4312:
It's automatically available if the debug target supports networking

vish01:
I am trying to debug an API call from a local VScode extension through debugger(which technically opens a new VScode window), is that supported?

connor4312:
VS Code extensions run in the extension host which is not on Node >=22.6 where networking support is available.

vish01:
Hmm I see, is that something I can update for my extension and upgrade node to >=22.6 or extension host node version needs to be upgraded by you guys?

connor4312:
The extension host's node is bundled with VS Code and isn't upgradable by users. And as I mentioned in the PR the network support Node added in 22.6 is _very_ preliminary and not incredibly useful at the moment, so it'll be a minute.

This works for browsers however.

gabrieldomgen:
Hi @connor4312 - I've tried that in the nightly build (I know it's not useful atm but wanted to check it out)

I have `"experimentalNetworking" set as "on" on my launch.json - and it does add the correctly flag to node, but I'm just not seeing the network panel nor any of the commands related to it. Running on node 22.9

Any ideas on what it could be?

connor4312:
You'll want to also enable the VS Code setting `debug.javascript.enableNetworkView`

suhli:
> You'll want to also enable the VS Code setting `debug.javascript.enableNetworkView`

still not showing the network views

.vscode/launch.json:
```json
{
      "name": "service(networking)",
      "type": "node",
      "request": "launch",
      "program": "${workspaceFolder}/service/dist/service/src/main.js",
      "timeout": 30000,
      "outputCapture": "std",
      "cwd": "${workspaceFolder}/service",
      "args": [],
      "env": {
        "NODE_ENV":"dev"
      },
      "experimentalNetworking": "on"
 }
```
.vscode/settings.json:
```json
{
  "debug.javascript.enableNetworkView": true
}
```

> node -v    
> v22.15.0

vscode:
```
Version: 1.100.0-insider (user setup)
Commit: abe4aab6a7ddb38f4a83c38b53cb4f28dddf0c97
Date: 2025-05-06T20:36:33.967Z
Electron: 34.5.1
ElectronBuildId: 11369351
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Windows_NT x64 10.0.26100
```

PR #2080, opened 2024-09-13
fix: don't inject NODE_OPTIONS in Deno

lucacasonato:
In Deno 2, Deno has a `process` global. Because of this this extension tries to inject environment variables into the process through the inspector using `process.env`. However when running Deno without I/O permissions (the default), accessing process.env.* will show a permission prompt. This is confusing for users, as their code did not trigger this prompt, but the extensions' injected code did.

Because this code doesn't do anything in Deno anyway (because we do not respect `NODE_OPTIONS`), there is no point in running it in Deno at all. This commit disables this injection code path in Deno through environment detection.

### DIFF
--- a/src/targets/node/nodeAttacherBase.ts
+++ b/src/targets/node/nodeAttacherBase.ts
@@ -21,7 +21,7 @@ const appendVars: readonly string[] = ['NODE_OPTIONS', 'VSCODE_INSPECTOR_OPTIONS
 export abstract class NodeAttacherBase<T extends AnyNodeConfiguration> extends NodeLauncherBase<T> {
   protected async appendEnvironmentVariables(cdp: Cdp.Api, vars: EnvironmentVars) {
     const expression =
-      `typeof process==='undefined'||process.pid===undefined?'process not defined':(()=>{`
+      `typeof process==='undefined'||process.pid===undefined||typeof Deno==='object'?'process not defined':(()=>{`
       + Object.entries(vars.defined())
         .map(([key, value]) => {
           const k = JSON.stringify(key);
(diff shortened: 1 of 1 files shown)


src/targets/node/nodeAttacherBase.ts
```diff
@@ -21,7 +21,7 @@ const appendVars: readonly string[] = ['NODE_OPTIONS', 'VSCODE_INSPECTOR_OPTIONS
 export abstract class NodeAttacherBase<T extends AnyNodeConfiguration> extends NodeLauncherBase<T> {
   protected async appendEnvironmentVariables(cdp: Cdp.Api, vars: EnvironmentVars) {
     const expression =
-      `typeof process==='undefined'||process.pid===undefined?'process not defined':(()=>{`
+      `typeof process==='undefined'||process.pid===undefined||typeof Deno==='object'?'process not defined':(()=>{`
```
connor4312, line 24: We should return something other than `process not defined` here in the Deno case to avoid the retry loop below

lucacasonato: `process not defined` already breaks out of the retry loop, right?

lucacasonato: Sorry, I read that code incorrectly. I'll have it return something else.

connor4312 (approved):
thanks!

aeschli (approved):

src/targets/node/nodeAttacherBase.ts
connor4312: Yea, for some reason in Node (haven't investigated the internals) the process can be undefined for a short time immediately after attachment, so we retry for a little while

PR #2092, opened 2024-10-03
feat: Add named pipes entrypoint for Visual Studio

EricCornelson:
Adding an entrypoint for external users (e.g. Visual Studio) which uses named pipes instead of TCP sockets.


connor4312:
Can we handle this in the existing vsDebugServer? E.g. it already can be run as `node debug-server.js <port-number>`, maybe if the port number is not numeric/is an absolute path we treat it as a socket/pipe.

connor4312:
Closing as part of our housekeeping, if you still need this let me know or reopen it

PR #2096, opened 2024-10-04
fix: only autofill "debug link" input if the hostname resolves

connor4312:
Fixes https://github.com/microsoft/vscode/issues/228950


connor4312:
Red tests fixed in https://github.com/microsoft/vscode-js-debug/pull/2098

rzhao271 (approved):

PR #2097, opened 2024-10-04
fix: show the subsession name before the root prefix

connor4312:
Fixes https://github.com/microsoft/vscode/issues/229100


connor4312:
Red tests fixed in https://github.com/microsoft/vscode-js-debug/pull/2098

rzhao271 (approved):

PR #2152, opened 2024-12-17
fix: use shell integration for more reliable execution for node-temrinal

connor4312:
Refs microsoft/vscode#236397

### DIFF
--- a/src/targets/node/terminalNodeLauncher.ts
+++ b/src/targets/node/terminalNodeLauncher.ts
@@ -26,7 +26,6 @@ import { IProgram } from './program';
 
 const SHELL_INTEGRATION_TIMEOUT = 1000;
 const SHELL_PROCESS_ID_TIMEOUT = 1000;
-const SHELL_INTEGRATION_SETTING = 'terminal.integrated.shellIntegration.enabled';
 
 class VSCodeTerminalProcess implements IProgram {
   public readonly stopped: Promise<IStopMetadata>;
@@ -177,8 +176,7 @@ export class TerminalNodeLauncher extends NodeLauncherBase<ITerminalLaunchConfig
 
   private async writeCommand(terminal: vscode.Terminal, command: string) {
     // Wait for shell integration if we expect it to be available
-    const config = vscode.workspace.getConfiguration();
-    if (!terminal.shellIntegration && config.get(SHELL_INTEGRATION_SETTING) !== false) {
+    if (!terminal.shellIntegration) {
       let listener: vscode.Disposable;
       await Promise.race([
         delay(SHELL_INTEGRATION_TIMEOUT),
--- a/src/targets/node/terminalNodeLauncher.ts
+++ b/src/targets/node/terminalNodeLauncher.ts
@@ -24,6 +24,10 @@ import { hideDebugInfoFromConsole, INodeBinaryProvider, NodeBinary } from './nod
 import { IProcessTelemetry, IRunData, NodeLauncherBase } from './nodeLauncherBase';
 import { IProgram } from './program';
 
+const SHELL_INTEGRATION_TIMEOUT = 1000;
+const SHELL_PROCESS_ID_TIMEOUT = 1000;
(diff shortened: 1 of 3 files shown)


src/targets/node/terminalNodeLauncher.ts
```diff
@@ -174,6 +175,35 @@ export class TerminalNodeLauncher extends NodeLauncherBase<ITerminalLaunchConfig
     });
   }
 
+  private async writeCommand(terminal: vscode.Terminal, command: string) {
+    // Wait for shell integration if we expect it to be available
+    const config = vscode.workspace.getConfiguration();
+    if (!terminal.shellIntegration && config.get(SHELL_INTEGRATION_SETTING) !== false) {
```
Tyriar, line 181: Shell integration may activate or not regardless of this setting, the setting only determines whether we try to automatically inject it. It could fail or it could be manually installed in their start up scripts. So you shouldn't be checking this setting here.

src/targets/node/terminalNodeLauncher.ts
```diff
@@ -174,6 +175,35 @@ export class TerminalNodeLauncher extends NodeLauncherBase<ITerminalLaunchConfig
...
+      let listener: vscode.Disposable;
+      await Promise.race([
+        delay(SHELL_INTEGRATION_TIMEOUT),
```
Tyriar, line 184: For python we're trying to do some caching mechanism to skip this timeout depending on whether the terminal is likely to have shell integration. You could track whether the shell exe and args had shell integration last time and if not revert to sendText immediately (or after some small timeout if needed to avoid this pwsh problem).

src/targets/node/terminalNodeLauncher.ts
```diff
@@ -24,6 +24,10 @@ import { hideDebugInfoFromConsole, INodeBinaryProvider, NodeBinary } from './nod
 import { IProcessTelemetry, IRunData, NodeLauncherBase } from './nodeLauncherBase';
 import { IProgram } from './program';
 
+const SHELL_INTEGRATION_TIMEOUT = 1000;
+const SHELL_PROCESS_ID_TIMEOUT = 1000;
+const SHELL_INTEGRATION_SETTING = 'terminal.integrated.shellIntegration.enabled';
```
Tyriar, line 29: nit: `const enum` is great for this type of thing, as the values are inlined and are then grouped and can have jsdoc in code.

Tyriar (changes requested):

src/targets/node/terminalNodeLauncher.ts
connor4312: My thought was just to avoid the 1s race if I see the user has explicitly disabled it

src/targets/node/terminalNodeLauncher.ts
connor4312: Wonder if the API could signal that to avoid all of us duplicating logic. Some `terminal.isLikelyToHaveShellIntegration`

src/targets/node/terminalNodeLauncher.ts
Tyriar: But then you're degrading the experience for users that have manually installed it, which are arguably those that are more engaged with the feature as they've gone and sought it out.

src/targets/node/terminalNodeLauncher.ts
Tyriar: I've thought about that, but it's such an ugly API and doubt it would make it through the syncs tbh. There are some basic examples of fallbacks in the comments but @anthonykim1/Python was about to try out this idea soon.

connor4312: It doesn't look like the shell exe / args are available on the vscode.Terminal object, do you have suggestions on how to get those?

Tyriar: They're in creationOptions, if the shell path is undefined you can assume it's env.shell

Tyriar (approved):

PR #2159, opened 2025-01-08
Adds originalUrl to GetPreferredUILocationResult

hediet:
Introduce originalUrl in GetPreferredUILocationResult to enhance URL retrieval from stack traces.

### DIFF
--- a/src/adapter/debugAdapter.ts
+++ b/src/adapter/debugAdapter.ts
@@ -153,7 +153,7 @@ export class DebugAdapter implements IDisposable {
           source: params.source,
         };
       } else {
-        throw new Error('Missing source information');
+        throw new ProtocolError(errors.missingSourceInformation());
       }
     }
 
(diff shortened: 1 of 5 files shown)


src/adapter/debugAdapter.ts
```diff
@@ -136,9 +136,25 @@ export class DebugAdapter implements IDisposable {
...
-      return params;
+      if (params.source) {
+        // Return unmodified input source
+        return {
+          column: params.column,
+          line: params.line,
+          source: params.source,
+        };
+      } else {
+        throw new Error('Missing source information');
```
connor4312, line 156: You should make a new error in `dap/errors.ts` and throw that so it gets an error code and is typed nicely as an expected error

connor4312 (approved):

hediet:
Thanks for the quick review!

Skippyy96:
Thank you

Issue #2183, opened 2025-03-16
Path normalization in setBreakpoints


erikson84:
I have recently run into an issue using vscode-js-debug with nvim-dap in Neovim, Powershell and Windows.

I opened [a bug report](https://github.com/mfussenegger/nvim-dap/issues/1466) with nvim-dap, but I was wondering if lack of path normalization in the adapter could be at fault.

Long story short: I was having problems with unbound breakpoints when debugging an Angular application, which was working fine before. I was able to narrow down the problem: when nvim-dap sends the path field for the source object in a setBreakpoint request, it uses the absolute path with forward slashes; VSCode, on the other hand, sends the exact same path, but with backslashes. It works in VSCode, but the forward slash in nvim-dap makes the breakpoint unbound.

I was wondering if this path normalization could be done at the adapter level, insteaf of creating another conditional case for nvim-dap.

connor4312:
In general we expect the `source.path` to be the appropriate path representation for the platform. For Windows paths, that's with backslashes. But I would be okay being tolerant to forward slashes here, if you find where the underlying issue is I would take a PR

mfussenegger:
> In general we expect the source.path to be the appropriate path representation for the platform. For Windows paths, that's with backslashes.

Where's the definition for this appropriate path representation?
E.g. https://en.wikipedia.org/wiki/Path_(computing) defines both `/` or `\` as valid separator on windows.

If some normalization is required, I think this should be pointed out in the specification.

connor4312:
Windows paths use backslashes, see the "Microsoft Windows" section on the page you linked. More on paths on Windows: https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats

erikson84:
I noticed someone filed a [similar issue with vscode-java-debug](https://github.com/microsoft/java-debug/issues/318): I tested and noticed breakpoints are getting unbound in my Java code, too. What is strange is that everything was working fine two weeks ago; maybe some plugins was setting `noshellslash` in my Neovim config.

connor4312:
> Coud something like this break the setBreakpoints method in another context?

That seems like a safe change

HppZ:
is this related?

![Image](https://github.com/user-attachments/assets/5924b9b4-3de4-418f-8c5b-349deff78175)

PR #2184, opened 2025-03-20
fix: ensure backslash in path when setting breakpoints in windows

erikson84:
Fixes https://github.com/microsoft/vscode-js-debug/issues/2183


erikson84:

connor4312 (approved):

PR #2185, opened 2025-03-21
feat: support ModuleConfiguration for `ms-vscode.wasm-dwarf-debugging`

CPunisher:
Issue https://github.com/microsoft/vscode-dwarf-debugging-ext/issues/7 gives a solution of path substitutions in browser. But it doesn't work when debugging webassembly in nodejs. I wonder why we cannot directly add a configuration. Here it is.

Docs and descriptions come from the jsdocs in `@vscode/dwarf-debugging/chrome-cxx/mnt/extension/ModuleConfiguration`

### DIFF
--- a/OPTIONS.md
+++ b/OPTIONS.md
@@ -51,7 +51,8 @@
 <h5>Default value:</h4><pre><code>true</pre></code><h4>timeout</h4><p>Retry for this number of milliseconds to connect to Node.js. Default is 10000 ms.</p>
 <h5>Default value:</h4><pre><code>10000</pre></code><h4>timeouts</h4><p>Timeouts for several debugger operations.</p>
 <h5>Default value:</h4><pre><code>{}</pre></code><h4>trace</h4><p>Configures what diagnostic output is produced.</p>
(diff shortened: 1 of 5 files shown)


CPunisher:

CPunisher:
> `sourceMapPathOverrides` should work for Node.js just as well as the browser. We don't rely on any special path mapping done by the dwarf module (and I wouldn't really want to add an extra way to map paths) and instead pass the path through the normal path mapping logic (`ISourcePathResolver.urlToAbsolutePath`)

I agree that it's good to keep only one way to map paths. I should appologize for my inaccurate statement: 
> But it doesn't work when debugging webassembly in nodejs

Let me clarify my issue. There is actually [a bug in the `
C/C++ DevTools Support (DWARF)` chrome extension](https://issues.chromium.org/issues/401522579). A temporary solution is adding path substitutions in extension options, so I'd like to do the same thing in vscode.

I have tried to add `sourceMapPathOverrides`, but after debugging I find that the dwarf symbol url doesn't handled by `NodeSourcePathResolver.urlToAbsolutePath`. So  I wonder there are some cases that `urlToAbsolutePath` can't do but path substitution can when it comes to dwarf symbol resolution.

PR #2211, opened 2025-05-01
Fix bug where the WasmWorker instance is disposed but never re-spawned on page reloads + writeMemory when WASM memory is SharedArrayBuffer

mikaelwaltersson:
The `WasmWorkerFactory` instance caches a single `IWasmWorker` instance but the `dispose` method on the `IWasmWorker` instance still terminates the underlying worker. The effect is that the worker in the `@vscode/dwarf-debugging` terminates on browser page reloads and never re-spawns again until debugging is restarted.

This PR addresses this issue by still caching a single `IWasmWorker` instance but re-spawning it if the previous call to `dispose` caused the _"ref count"_ to reach zero.

Possible related to https://github.com/microsoft/vscode-dwarf-debugging/issues/6

Also fixes `writeMemory` when WASM memory is `SharedArrayBuffer`.

### DIFF
--- a/src/adapter/dwarf/wasmSymbolProvider.ts
+++ b/src/adapter/dwarf/wasmSymbolProvider.ts
@@ -103,6 +103,10 @@ export class WasmWorkerFactory implements IWasmWorkerFactory {
       getStopId: id => `${cdpId}:${id}`,
       dispose: () => {
         this.cdp.delete(cdpId);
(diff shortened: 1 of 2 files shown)


connor4312 (approved):
thanks!

rzhao271 (approved):

edwiin17:
> thanks!

Issue #2207, opened 2025-04-29
Offline build for NixOS


relief-melone:
I am currently trying to create a native vimPlugin implementation for NixOS for vscode-js-debug. Now there are some specifics which make building node related packages pretty hard on NixOS. one is the nature the package.json being non deterministic which goes against the basic principle of the way NixOS operates. Also a package-lock does not guarantee the packages have not been altered (version deleted and republished).
Adding  to this principle. NixOS does not allow network access during the build process. This is the tricky part for this one as npx gulp vsDebugServer does rely on internet access.

I am more than happy to provide a PR addressing the issue. However I do lack experience with gulp. So it would help to get some assistance here.

I did find that in the gulpfile.js the l10n bundle download is an example that relies on downloading the zip from a github repo. My suggestion would be that you could set env vars during the build process that specify the location of such files. Would this be a viable option or are there any roadblocks I did not think of that would make an offline build impossible?

connor4312:
> Also a package-lock does not guarantee the packages have not been altered (version deleted and republished).

The package-lock does include package checksums which are proof against modification.

> My suggestion would be that you could set env vars during the build process that specify the location of such files. 

Sure, I would take a PR to add this. If such an environment variable is set we would jsut return early from

https://github.com/microsoft/vscode-js-debug/blob/3ed213faba62916c4d73233bd837a57179b40b2a/gulpfile.js#L356

relief-melone:
I am working on this. It is alot of work (and a nice NixOS learning journey). The first concept I might come up with probably would need some improvements. As stated before one of the key differences is that for NixOS to utilize npm packages they need to be "converted" from the package.json to inputs (basically also a package list derived from package.json but one that includes all the content hashes for the packages to ensure that the build process is deterministic)

This requires me to include all the explicit architecture packages as well (so. e.g. in addition to esbuild @esbuild/linux-x64) is also needed.

I don't want the impact on this repo being to big. So just as a heads up. If someone is more experienced with this and has a clever idea on how to minimize the impact that would be great.

relief-melone:
Ok. I was able to buld a vimPlugin with NixOS for this. But included some heavy changes and I don't think it is ready for a PR yet as the impact on this repo should be reduced. The good news is it could be done by using the gulp task for the dapDebugServer which does not need an internet connection.

The progress of this as well as the changes I had to do are also in this discussion board

[...]

PR #2221, opened 2025-05-09
Feature.v1.100.0 vim plugin

relief-melone:


relief-melone:
Mentioned in https://github.com/microsoft/vscode-js-debug/issues/2207

The goal of the PR is not to be merged. The changes I made here were necessary to being able to create a vimPlugin for NixOS. I understand that removing scripts e.g. is not something that can be accepted. However I would like to take the opportunity to discuss what changes you think are acceptable and what changes need a different solution.

In any case it would be great if we were able to build a native vimPlugin for NixOS from this repo. Thank you for the work put in.

connor4312:
Thanks for the PR. However, I don't want to include `.nix` files in this repo. I don't use NixOS and don't have the spare bandwidth to learn and maintain those. Is there a way you can accomplish your goals without requiring .nix files in this repo?

relief-melone:
Hmm. It looks like I forgot to send my previous comment. However I gave it some thought and In was able to incorporate all the changes into the PR that goes to https://github.com/NixOS/nixpkgs.

If all goes well there should be no changes needed to this repository as I just refer to the rev v1.100.0. This is the PR

https://github.com/NixOS/nixpkgs/pull/406702

and this is the link to the discourse regarding the topic over at nixos.org also with information on how to use the plugin within NixOS.

https://discourse.nixos.org/t/vimplugin-for-javascript-debugging-vscode-js-debug-and-nvim-dap-vscode-js/64053

PR #2223, opened 2025-05-12
chore: Fix default url for turbopack

kdy1:
Fix the default URL of turbopack to match the content of https://github.com/vercel/next.js/pull/76075.

x-ref: https://github.com/vercel/next.js/issues/77670#issuecomment-2873760418

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -984,6 +984,7 @@ export function defaultSourceMapPathOverrides(webRoot: string): { [key: string]:
     'webpack://?:*/*': `${webRoot}/*`,
     'webpack:///([a-z]):/(.+)': '$1:/$2',
     'meteor://💻app/*': `${webRoot}/*`,
+    'turbopack://[project]/*': '${workspaceFolder}/*',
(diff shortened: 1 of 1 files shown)


src/configuration.ts
```diff
@@ -984,7 +984,7 @@ export function defaultSourceMapPathOverrides(webRoot: string): { [key: string]:
     'webpack://?:*/*': `${webRoot}/*`,
     'webpack:///([a-z]):/(.+)': '$1:/$2',
     'meteor://💻app/*': `${webRoot}/*`,
-    'turbopack://[project]/*': '${workspaceFolder}/*',
+    'turbopack:///[project]/*': '${workspaceFolder}/*',
```
connor4312, line 987: Should this be optional for back-compat with old versions?

```suggestion
    'turbopack:///?[project]/*': '${workspaceFolder}/*',
```

connor4312, line 987: I think we need another whole entry here for back-compat with old versions

connor4312 (approved):

rzhao271 (approved):

PR #2308, opened 2026-01-12
fix: don't duplicate --allow-all for deno debug configuration

nayeemrmn:
- Follows #1348

Doesn't append `--allow-all` if it's specified by the user


src/ui/configuration/nodeDebugConfigurationResolver.ts
```diff
@@ -143,13 +143,18 @@ export class NodeConfigurationResolver extends BaseConfigurationResolver<AnyNode
           config.attachSimplePort = port;
           config.continueOnAttach ??= true;
 
-          const runtimeArgs = [`--inspect-brk=127.0.0.1:${port}`, '--allow-all'];
+          const runtimeArgs = [`--inspect-brk=127.0.0.1:${port}`];
           if (!config.runtimeArgs) {
             config.runtimeArgs = ['run', ...runtimeArgs];
```
connor4312: Is it intentional that this removes `--allow-all`  from the default runtime args if none are specified?

Issue #1617, opened 2023-03-29
Breakpoint is failing on nodemon + jest + docker / compose setup


yuguan91:
**Describe the bug**
Breakpoints fails to work on nodemon + jest + docker. Sometimes I get a breakpoint at wrong line numbers in a readonly file with the same name, sometimes no breakpoints at all.

My local files are located in /User/me/some/path/app
Inside docker container, the files are located in root `'/app'`

I have a prelaunch task:
```
        {
            "label": "Start App",
            "type": "shell",
            "command": "./scripts/start-app.sh",
            "group": "test",
            "problemMatcher": [],
            "isBackground": true,
            "presentation": {
                "reveal": "always",
                "panel": "shared"
            }
        }
```

where ./scripts/start-app.sh essentially does:
```
docker-compose exec api \
    nodemon --no-lazy --legacy-watch --watch /app \
        --inspect-brk=app:9229 \
        /app/test/node_modules/jest/bin/jest.js --coverage \
            --config=/service/jest.config.json --runInBand
```

And an attach configuration:
```
{
            "type": "node",
            "request": "attach",
            "name": "App",
            "port": 9229,
            "restart": true,
            "preLaunchTask": "Start App",
            "localRoot": "${workspaceFolder}/app/",
            "remoteRoot": "/",
            "outFiles": [
                "${workspaceFolder}/app/**/*.js",
                "!**/node_modules/**"
            ],
            "skipFiles": [
                "<node_internals>/**/*.js",
                "${workspaceFolder}/**/node_modules/**/*.js"
            ],
            "pauseForSourceMap": true,
            "disableOptimisticBPs": true,
            "internalConsoleOptions": "neverOpen",
}
```

Debugger is able to attach to the process, because break point does hit the first line of`resolve()` in path.js. But `debugger` statement doesn't work, nor does any break points.

I've got the vscode-js-debug nightly extension, and debugged it to a point where I think I found the cause:

When using jest, my code is getting compiled to add instrumentation. The source map somehow already has the original local path, so when _onScriptParsed calls rebaseRemoteToLocal() the following happens:
```
  public rebaseRemoteToLocal(remotePath: string) {
    // this.options === {
    //   removeRoot: '/',
    //   localRoot: '/User/me/some/path/app'
    // }
    // remotePath === '/User/me/some/path/app/file.js'

    if (!this.options.remoteRoot || !this.options.localRoot || !this.canMapPath(remotePath)) {
[...]

PR #2345, opened 2026-04-17
fix: resolve relative sourceMappingURL against local path when remoteRoot/localRoot is set

chrisdp:
## Summary

When using `remoteRoot`/`localRoot` to debug a script on a remote device, a relative `sourceMappingURL` that navigates _outside_ the `remoteRoot` subtree fails to load with ENOENT.

**Root cause:** `sourceMapURL` is resolved against the raw remote script URI before `remoteRoot`/`localRoot` path translation is applied. If the resolved path doesn't start with `remoteRoot`, no translation is applied and js-debug tries to open a non-existent device path as a local file.

**Example:** with `remoteRoot: .../remote`, `localRoot: .../local`, and `sourceMappingURL: ../maps/index.js.map` in `remote/index.js`:

```
Remote URI:  file://.../remote/index.js
Resolved:    file://.../maps/index.js.map   ← outside remoteRoot, ENOENT
```

The correct resolution uses the locally-translated script path as the base:

```
Local path:  .../local/index.js
Resolved:    .../local/maps/index.js.map    ✓
```

## Fix

In `SourceContainer.addSource`, after `urlToAbsolutePath` produces the locally-translated `absolutePath` for the script, re-resolve a relative `sourceMapURL` against that local path. This only applies when:

- `absolutePath` is non-empty (translation actually happened)
- `event.url` is a `file://` URL (excludes http/webpack dev-server cases where the map must be fetched from the remote server)
- `sourceMapURL` is not a data URI

Absolute `sourceMapURL`s are unaffected — `completeUrl` ignores the base when the input is already an absolute URL.

This also fixes the secondary effect where `absSourceRoot` was derived from the remote script dirname instead of the local one, causing TypeScript source paths to resolve to nonexistent locations.

## Test

Added an integration test that creates this fixture layout:

```
remote/
  index.js          ← deployed script, sourceMappingURL=../maps/index.js.map
local/
  maps/
    index.js.map    ← source map (only exists locally, never deployed)
  src/
    index.js        ← original source
```

With `remoteRoot: .../remote` and `localRoot: .../local`, the golden file asserts the stack frame references `local/src/index.js` (the mapped original source). Without the fix, the source map isn't found and the frame falls back to the compiled path.

## Related issues

- #1240 — Apply pathMapping when loading sourcemaps
- #1617 — Breakpoint failing with Docker/compose remoteRoot setup
- microsoft/vscode#100973 — remoteRoot and sourceMaps do not work properly with TypeScript and Docker


chrisdp:

chrisdp:
Would love to get feedback so I can make changes or address issues related to this PR. 

Thanks!

PR #2349, opened 2026-04-22
Provide an opt-out for debug entrypoint resolution (compiled file remapping)

firelizzard18:
`NodeLauncher.tryGetCompiledFile` uses the breakpoint predictor to remap a non-`.js` `program` path to its compiled output before launching. In most cases this is helpful, but it has a failure mode: if the source file appears in a source map as a *dependency* of another entry point (e.g. a bundled tool), the predictor resolves `program` to that other entry point instead of the file you specified. There is currently no way to opt out.

**[Reproduction](https://github.com/firelizzard18/vscode-js-debug-issue-2349):**

- `some-lib.mjs` contains `if (esMain(import.meta)) { ... }` for running standalone
- `some-tool.mjs` imports `some-lib.mjs` and has been compiled, producing a source map that references `some-lib.mjs` as a source
- Launch config: `"program": "${workspaceFolder}/some-lib.mjs"`
- Result: the debugger silently launches `some-tool.mjs` instead

Deleting build output doesn't help — the predictor just picks a different unrelated entry point. There is no warning or indication that remapping occurred.

This also affects projects using Node's `--experimental-strip-types` flag to run `.ts` files directly with no compilation step. In that workflow there is no compiled output, yet the predictor will still attempt remapping if any source map in the workspace references the file.

**Fix:**

Adds a `debug.javascript.resolveDebugEntrypoint` workspace setting (default `true`, preserving existing behavior). When set to `false`, `tryGetCompiledFile` returns the `program` path as-is.

```jsonc
// .vscode/settings.json
{
  "debug.javascript.resolveDebugEntrypoint": false
}
```


connor4312 (approved):
thanks!

roblourens (approved):

PR #2369, opened 2026-06-03
fix: use DI for vscode API in nodeLauncher instead of direct import

lucygramley:
## Problem

The `vsDebugServerBundle` (and other non-extension bundles where `isInVsCode=false`) fails to build with:

\\\
X [ERROR] Could not resolve "vscode"

    src/targets/node/nodeLauncher.ts:7:24:
      7 │ import * as vscode from 'vscode';
\\\

This started after aeb21f00 ("Provide an opt-out for debug entrypoint resolution") added `import * as vscode from 'vscode'` to `nodeLauncher.ts`.

## Root Cause

The direct `import * as vscode from 'vscode'` causes two problems for standalone (non-VS Code) bundles:

1. **Build error** – esbuild can't resolve the `vscode` module since it's only in the `external` list when `isInVsCode` is true
2. **Runtime error** – even if marked external, the emitted `require('vscode')` would fail at runtime since the module doesn't exist outside VS Code

## Fix

Follows the established pattern used in `defaultBrowserProvider.ts`, `nodeBinaryProvider.ts`, and other files:

- `import type * as vscodeType from 'vscode'` – type-only import, erased at runtime
- `@optional() @inject(VSCodeApi) private readonly vscode?: typeof vscodeType` – DI injection, `undefined` outside VS Code
- Guard the `readConfig` call with an `if (!this.vscode)` check, falling back to the original behavior (no entrypoint resolution)

### DIFF
--- a/src/targets/node/nodeLauncher.ts
+++ b/src/targets/node/nodeLauncher.ts
@@ -2,9 +2,9 @@
  * Copyright (C) Microsoft Corporation. All rights reserved.
  *--------------------------------------------------------*/
 
-import { inject, injectable, multiInject } from 'inversify';
+import { inject, injectable, multiInject, optional } from 'inversify';
 import { extname, resolve } from 'path';
-import * as vscode from 'vscode';
+import type * as vscodeType from 'vscode';
 import { IBreakpointsPredictor } from '../../adapter/breakpointPredictor';
 import { IPortLeaseTracker } from '../../adapter/portLeaseTracker';
 import Cdp from '../../cdp/api';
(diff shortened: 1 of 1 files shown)


connor4312:
THanks for identifying the issue and the PR. We should actually switch the `import` to `import type` and DI the vscode API as in

https://github.com/microsoft/vscode-js-debug/blob/f460326b31c6643798a1af09228b6a973ec3fba3/src/common/defaultBrowserProvider.ts#L33

Otherwise the code will still error when executed.

lucygramley:
Thanks Connor - I've adjusted the PR. Let me know if it doesn't align with your comments :)

src/targets/node/nodeLauncher.ts
```diff
@@ -313,7 +315,11 @@ export class NodeLauncher extends NodeLauncherBase<INodeLaunchConfiguration> {
       return targetProgram;
     }
 
-    const resolve = readConfig(vscode.workspace, Configuration.ResolveDebugEntrypoint);
+    if (!this.vscode) {
+      return targetProgram;
+    }
+
+    const resolve = readConfig(this.vscode.workspace, Configuration.ResolveDebugEntrypoint);
```
connor4312:
```suggestion
    const resolve = this.vscode ? readConfig(this.vscode.workspace, Configuration.ResolveDebugEntrypoint) : true;
```

to preserve old behavior (non-breaking)

connor4312 (approved):
thanks!

joj (approved):

PR #2393, opened 2026-08-06
fix: identifier regexp supports unicode identifiers

### LINKED ISSUE #2391: Make identifier regular expression configurable
There are several places in the code where a regular expression is used to find identifiers in JavaScript code. As mentioned in the comment, this is a very approximate regular expression because it only considers Latin characters. As a result, if the code contains non-Latin identifiers, some debugger features do not work correctly.

Here is one example:
https://github.com/microsoft/vscode-js-debug/blob/a72801b175c1856138e30679624c97b3b062dc65/src/common/sourceMaps/renameScopeAndSourceMap.ts#L15-L16
https://github.com/microsoft/vscode-js-debug/blob/a72801b175c1856138e30679624c97b3b062dc65/src/common/sourceMaps/renameScopeAndSourceMap.ts#L47

Here, the regular expression is used to construct variable renames based on source maps. In my scenario, the generated JavaScript code contains non-Latin identifiers, so the renaming does not work and the `Variables` view displays the names from the generated code instead. This makes debugging inconvenient because the variable names in the source code do not match the names shown in the `Variables` view.

A possible solution would be to make the identifier regular expression configurable. An `identifierRegex` field could be added to `IBaseConfiguration`. If specified, it would be used instead of the default regular expression. This would allow users to provide a more appropriate regular expression.

itrapashko:
Fixes #2391

The new regular expression supports all JavaScript identifiers. It was taken from the following article: https://v8.dev/features/regexp-match-indices

### DIFF
--- a/src/adapter/completions.ts
+++ b/src/adapter/completions.ts
@@ -282,7 +282,7 @@ export class Completions {
 
     const start = getStart(node.property) - 1;
 
-    // For any properties are aren't valid identifiers, quote them as foo['bar!']
+    // For any properties that aren't valid identifiers, quote them as foo['bar!']
     for (const item of result) {
       if (!validIdentifierRe.test(item.label)) {
         item.text = `[${JSON.stringify(item.label)}]`;
--- a/src/adapter/completions.ts
+++ b/src/adapter/completions.ts
@@ -132,6 +132,8 @@ export interface ICompletions {
   completions(options: ICompletionContext & ICompletionExpression): Promise<Dap.CompletionItem[]>;
 }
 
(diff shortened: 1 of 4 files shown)


itrapashko:
@connor4312 I have added test to `completion.test.ts` as suggested by Copilot, but it did not pass because of completions sorting order. Current implementation uses leading `~` symbols to sort completion items. It works for ASCII letters, but identifiers with greater character codes (e.g. Cyrillic) sort after prototype members, because string comparison is lexicographic. For example, `~~привет` sorts after `~~~constructor` even though it has fewer `~`. Should this issue be addressed in this PR? All other problems are fixed.

connor4312 (approved):
THis is good, thanks!

dmitrivMS (approved):

aeschli (approved):

PR #2398, opened 2026-08-23
feat: narrow outFiles to workspace dependencies in npm/yarn monorepos

### LINKED ISSUE #1730: Using dependency graph to reduce file scanning
Hello, we are experiencing very long debugging times in the time it takes to reach a breakpoint. Even with scoping down the scanned files with the outfile argument

We work in a large yarn v1 monorepo with a sparse dependency graph. We'd like to use the dependency graph to restrict the amount of file scanning js-debug needs to do on startup. With the arrival of npm workspaces, it feels reasonable for js-debug to be able to reason about workspaces in a monorepo

Aparajith24:
In an npm/yarn `workspaces` monorepo, `outFiles` was previously left to fall back to scanning the entire opened folder for sourcemaps. This happened because the widen-only logic in `guessOutFiles` only kicks in when the debugged program lives *outside* the opened workspace folder, but in a monorepo, the folder VS Code has open typically *is* the repo root, so the program is always "inside" it and the widen logic never narrows anything. The result: debugging one package in a large monorepo scanned every other unrelated package's compiled output too, making it slow to hit breakpoints.

This adds a narrower path: when the opened workspace folder itself has a "workspaces" field in its "package.json" (npm/yarn workspaces), we:

1. Find the specific package being debugged (nearest package.json walking up from the entry file).
2. Read the workspace's package list from the workspaces glob patterns (e.g. packages/*).
3. Walk that package's dependencies/devDependencies/peerDependencies to find which of them are other packages in the same workspace*.
4. Set `outFiles` to just the debugged package plus those workspace-local dependencies, instead of the whole repo.

This only applies when outFiles isn't already set by the user (same guard the existing widen-only logic uses), so anyone with a manually configured outFiles is unaffected. Projects without a workspaces field are also completely unaffected, this is additive, not a change to existing behavior.

Closes #1730

### DIFF
--- a/src/test/extension/nodeConfigurationProvider.test.ts
+++ b/src/test/extension/nodeConfigurationProvider.test.ts
@@ -433,6 +433,68 @@ describe('NodeDebugConfigurationProvider', () => {
     });
   });
 
+  describe('workspaces outFiles', () => {
+    it('narrows to the debugged package and its workspace deps', async () => {
+      createFileTree(testFixturesDir, {
+        'package.json': JSON.stringify({ name: 'root', workspaces: ['packages/*'] }),
+        packages: {
+          a: {
+            'index.js': '',
+            'package.json': JSON.stringify({ name: 'pkg-a', dependencies: { 'pkg-b': '^1.0.0' } }),
+          },
+          b: {
+            'index.js': '',
+            'package.json': JSON.stringify({ name: 'pkg-b' }),
+          },
+          c: {
+            'index.js': '',
+            'package.json': JSON.stringify({ name: 'pkg-c' }),
+          },
+        },
+      });
+
+      const result = await provider.resolveDebugConfiguration(folder, {
+        type: DebugType.Node,
+        name: '',
+        request: 'launch',
+        program: 'packages/a/index.js',
+      });
+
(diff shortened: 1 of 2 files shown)


Aparajith24:
> oyeappu (@Aparajith24) please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

PR #2402, opened 2026-08-28
fix: don't let a listener exception wedge EventEmitter

### LINKED ISSUE #2401: Listener exception permanently wedges EventEmitter, deadlocking the debuggee at startup
> **Filed by GitHub Copilot** on behalf of @jruales. Root cause verified in source; fix validated by a controlled A/B in a real VS Code window.

## Summary

A single listener exception permanently wedges `EventEmitter`, which silently kills the CDP message pump for that connection. When it happens during target attach, js-debug never sends `Runtime.runIfWaitingForDebugger`, so the debuggee stays paused at startup forever.

Symptom when debugging an Electron app (the `microsoft/vscode` workbench): the window renders but accepts no input — nothing clickable, keyboard dead, command palette opens but neither responds nor dismisses.

## Root cause

**1. `EventEmitter.fire` doesn't restore `_deliveryQueue` when a listener throws** — [`src/common/events.ts#L55-L65`](https://github.com/microsoft/vscode-js-debug/blob/main/src/common/events.ts#L55-L65)

```ts
if (!dispatch) return;
for (let index = 0; index < this._deliveryQueue.length; index++) {
  const { data, event } = this._deliveryQueue[index];
  data.listener.call(data.thisArg, event);   // throws here
}
this._deliveryQueue = undefined;             // never reached
```

`_deliveryQueue` stays truthy, so every later `fire()` computes `dispatch === false`, enqueues, and returns without dispatching. The emitter is permanently deaf. Since `Connection` consumes its transport via `this._transport.onMessage(...)` ([`connection.ts#L59`](https://github.com/microsoft/vscode-js-debug/blob/main/src/cdp/connection.ts#L59)), one bad message permanently stops all CDP traffic on that connection.

**2. `Connection._onMessage` throws on an unknown session id** — [`src/cdp/connection.ts#L104-L110`](https://github.com/microsoft/vscode-js-debug/blob/main/src/cdp/connection.ts#L104-L110)

Messages for *disposed* sessions are ignored, but messages for *unknown* sessions throw — which is what enters the wedge above.

## Why an unknown session id occurs

`Inspector.workerScriptLoaded` is a Blink event fired from `WorkerGlobalScope::RunWorkerScript()` when a **web worker** finishes evaluating its script. It is never handled by js-debug at runtime (it appears only in the generated `src/cdp/api.d.ts`), so it arrives unsolicited.

Session registration appears to be deferred: `Target.setAutoAttach({ flatten: true })` delivers worker targets, but the `targetCreated` handler is wrapped in `enqueueLifecycleFn` ([`browserTargetManager.ts#L171-L174`](https://github.com/microsoft/vscode-js-debug/blob/main/src/targets/browser/browserTargetManager.ts#L171-L174)), so `createSession(sessionId)` runs on a later microtask. A `workerScriptLoaded` landing in that window finds no registered session.

This matches the observed behaviour: it is intermittent, and it reproduces with workbench web workers such as `editorWorkerService`.

## Evidence

Error at the exact moment of the freeze, with `EventEmitter.fire` directly beneath the throw:

```
[error] Error: Unknown session id: B9878282765DBAAB165E4A5851627507 while processing: Inspector.workerScriptLoaded
    at ut._onMessage (.../src/extension.js:58:8166)
    at U.fire        (.../src/extension.js:39:11654)
```

The debuggee was confirmed *waiting for the debugger*, not paused at a breakpoint:

- `Runtime.evaluate` of synchronous code succeeded
- `setTimeout` / `requestAnimationFrame` never fired; event handlers never ran
- `Debugger.resume` returned `Can only perform operation while paused`
- `Runtime.runIfWaitingForDebugger` immediately restored the window

Minimal repro against the unmodified `EventEmitter`:

```ts
const emitter = new EventEmitter<string>();
const delivered: string[] = [];
emitter.event(v => { if (v === 'poison') throw new Error('boom'); delivered.push(v); });

emitter.fire('before');
try { emitter.fire('poison'); } catch {}
emitter.fire('after-1');

console.log(delivered); // ["before"]  <-- 'after-1' is never delivered
```

## Fix

#2402 restores `_deliveryQueue` in a `finally` block, and treats unknown sessions the way disposed ones are already treated (warn and ignore).

Validated by a controlled A/B: the extension built both ways, loaded into a real VS Code Insiders window via `--extensionDevelopmentPath`, same machine/repo/profile/launch config, only the diff varying.

| Build | Outcome |
|---|---|
| Unpatched | Workbench freezes. Reproduced. |
| Patched | Works every time, across repeated runs. |

## Environment

| | |
|---|---|
| js-debug | 1.117.0 (bundled with VS Code Insiders) |
| VS Code | 1.136.0-insider |
| Electron (debuggee) | 42.8.1 |
| OS | macOS (arm64) |
| Scenario | Debugging `microsoft/vscode` via its `Renderer and Main processes` compound |

jruales:
> **Authored by GitHub Copilot** on behalf of @jruales.

Fixes #2401

## Problem

`EventEmitter.fire` leaves `_deliveryQueue` populated when a listener throws:

```ts
if (!dispatch) return;
for (let index = 0; index < this._deliveryQueue.length; index++) {
  const { data, event } = this._deliveryQueue[index];
  data.listener.call(data.thisArg, event);   // throws
}
this._deliveryQueue = undefined;             // never reached
```

Because the queue is never cleared, every later `fire` computes `dispatch === false`, enqueues, and returns without dispatching. **One listener exception permanently stops all delivery on that emitter.**

`Connection` reads its transport through `this._transport.onMessage(...)`, so this silently kills the CDP message pump for the whole connection. The debuggee is then never sent `Runtime.runIfWaitingForDebugger` and stays paused at startup — the window paints but accepts no input.

The exception comes from `Connection._onMessage`, which ignores messages for *disposed* sessions but throws for *unknown* ones. `Inspector.workerScriptLoaded` (a Blink event fired when a web worker finishes evaluating) is never handled by js-debug and can arrive before the worker's session is registered, since `targetCreated` is queued through `enqueueLifecycleFn`:

```
Error: Unknown session id: B9878282765DBAAB165E4A5851627507 while processing: Inspector.workerScriptLoaded
    at ut._onMessage (.../src/extension.js:58:8166)
    at U.fire        (.../src/extension.js:39:11654)
```

## Changes

- **`src/common/events.ts`** — restore `_deliveryQueue` in a `finally` block so one bad listener can't wedge the emitter. This is the load-bearing fix and is independent of the CDP specifics.
- **`src/cdp/connection.ts`** — treat unknown sessions the way disposed sessions are already treated (warn and ignore) instead of throwing.
- **`src/common/events.test.ts`** — regression test.

## Validation

- New test fails without the `events.ts` change (`expected [ 'before' ] to deeply equal [ 'before', 'after' ]`) and passes with it.
- `npm run test:unit` — **268 passing**. `tsc --noEmit` and `gulp lint` clean.
- End-to-end A/B: this extension built both ways, loaded into a real VS Code Insiders window via `--extensionDevelopmentPath`; same machine, repository (`microsoft/vscode`, Electron 42.8.1), isolated profile and launch compound, only this diff varying.

| Build | Outcome |
|---|---|
| Unpatched | Workbench freezes. Reproduced. |
| Patched | Works every time, across repeated runs. |

Happy to split the `connection.ts` change into a separate PR if you'd prefer to evaluate the two independently.

### DIFF
--- a/src/cdp/connection.ts
+++ b/src/cdp/connection.ts
@@ -105,9 +105,12 @@ export default class Connection {
     if (!session) {
       const disposedDate = this._disposedSessions.get(object.sessionId);
       if (!disposedDate) {
-        throw new Error(
-          `Unknown session id: ${object.sessionId} while processing: ${object.method}`,
+        this.logger.warn(
+          LogTag.Internal,
+          'Got message for an unknown session',
+          { sessionId: object.sessionId, method: object.method },
         );
+        return; // We just ignore messages for unknown sessions
       } else {
         const secondsAgo = (Date.now() - disposedDate.getTime()) / 1000.0;
         this.logger.warn(
--- a/src/common/events.test.ts
+++ b/src/common/events.test.ts
@@ -1,0 +1,27 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+import { expect } from 'chai';
+import { EventEmitter } from './events';
+
+describe('EventEmitter', () => {
+  it('keeps delivering events after a listener throws', () => {
+    const emitter = new EventEmitter<string>();
+    const delivered: string[] = [];
+
+    emitter.event(event => {
+      if (event === 'throws') {
+        throw new Error('listener error');
+      }
+
+      delivered.push(event);
+    });
+
+    emitter.fire('before');
+    expect(() => emitter.fire('throws')).to.throw('listener error');
+    emitter.fire('after');
+
+    expect(delivered).to.deep.equal(['before', 'after']);
+  });
+});
--- a/src/common/events.ts
+++ b/src/common/events.ts
@@ -57,11 +57,14 @@ export class EventEmitter<T> implements IDisposable {
     if (!this._deliveryQueue) this._deliveryQueue = [];
     for (const data of this._listeners) this._deliveryQueue.push({ data, event });
     if (!dispatch) return;
-    for (let index = 0; index < this._deliveryQueue.length; index++) {
-      const { data, event } = this._deliveryQueue[index];
-      data.listener.call(data.thisArg, event);
+    try {
+      for (let index = 0; index < this._deliveryQueue.length; index++) {
+        const { data, event } = this._deliveryQueue[index];
+        data.listener.call(data.thisArg, event);
+      }
+    } finally {
+      this._deliveryQueue = undefined;
     }
-    this._deliveryQueue = undefined;
   }
 
   dispose() {


jruales:
## Update: end-to-end validation completed — the earlier caveat is resolved

I previously downgraded this PR's claims because a standalone-adapter harness didn't reproduce the freeze. I've now validated it properly in a real VS Code environment, and **the fix resolves the bug**.

I built this extension both ways and loaded each into a real VS Code Insiders window via `--extensionDevelopmentPath`, overriding the bundled js-debug. Same machine, same repository (`microsoft/vscode`, Electron 42.8.1), same isolated `--user-data-dir`, same launch compound, no workaround active. The only variable was this diff.

| Build | Outcome |
|---|---|
| **Unpatched** | Workbench freezes — renders but accepts no input. Reproduced. |
| **Patched** | Works **every time**, across repeated runs. |

The unpatched build logs the throw with `EventEmitter.fire` directly beneath it — the wedge path, observed live:

```
Error: Unknown session id: B9878282765DBAAB165E4A5851627507 while processing: Inspector.workerScriptLoaded
    at ut._onMessage (/private/tmp/jsdbg-dist-unpatched/src/extension.js:58:8166)
    at U.fire (/private/tmp/jsdbg-dist-unpatched/src/extension.js:39:11654)
```

The patched build logged zero such errors.

My earlier harness simply lacked VS Code's auto-attach infrastructure, so the untracked session that emits `Inspector.workerScriptLoaded` never appeared — a limitation of the harness, not evidence against the diagnosis. Details in #2401.

PR #2408, opened 2026-09-05
Add Nub loader TypeScript debug snippet

colinhacks:
Adds an opt-in Node launch snippet for TypeScript programs using `@nubjs/loader`.

The snippet keeps `node` as the runtime executable and passes the loader through `--import`, so it uses the existing Node debug path.


colinhacks:
Could a maintainer approve the pending CI run?

colinhacks:
The naming hold is lifted. Continuing with the published `@nubjs/loader` package.

colinhacks:
The existing Node launch configuration already supports this; a dedicated snippet is not needed.

PR #2411, opened 2026-09-11
fix: avoid template local clash for primitive completions

### LINKED ISSUE #2399: Completions fail when the completed name matches a variable in the enumeration template
### Problem
Completing properties on a primitive (`target.`, `'abc'.`, `42.`) while paused evaluates `enumerateProperties` as a string on the call frame using template. Code that will be evaluated is constructed on line 351.

https://github.com/microsoft/vscode-js-debug/blob/d17455cb40a09fc35ce3396896642b4f9b7bc302/src/adapter/completions.ts#L346-L367

The template binds the user expression to its own parameters:

```js
(() => {
  let target = (target);
  let prefix = "";
  let isGlobal = false;
  // ... enumerate properties of `target`
})()
```

If the name being completed is also a parameter or local in that template (`target`, `prefix`, `isGlobal`, or a minified name like `n`), the inner target refers to the template binding instead of the user variable. Evaluation throws and completions are empty.

This does not affect objects: they are enumerated with `Runtime.callFunctionOn` on an `objectId`, so the helper never runs as source in the call frame.

In the currently released version, the following completion won't work because of described error (minified extension sources contain variable named `n`):
```js
let n = 1
debugger
// n.
```

### Proposed fix
The first evaluate already returned the primitive. Interpolate that value into the template (`let target = 1`) and run it with `Runtime.evaluate` (no call frame). The template no longer reads locals, so names cannot clash.

**VS Code Version:** 1.134.0

tanvir-ux:
## Summary
When completing properties on a paused primitive whose name matches a local in `enumeratePropertiesTemplate` (for example minified `n`, or `target`), the helper was re-evaluated on the call frame as `let n = (n)` / `let target = (target)`. That binds the template local instead of the user value, so evaluation throws and completions are empty.

This change interpolates the already-evaluated primitive into the template and runs it with `Runtime.evaluate` in the execution context (no call frame), matching the approach discussed in #2399.

## Test plan
- [x] Added `completes primitive when name clashes with template locals (#2399)` integration coverage for `n.` and `target.` while paused
- [ ] CI completion tests

Fixes #2399


tanvir-ux:

connor4312 (approved):
thanks!

dmitrivMS (approved):

rzhao271 (approved):

PR #2413, opened 2026-09-13
fix: honor explicit pause requests while smart stepping

### LINKED ISSUE #2412: Explicit DAP pause is swallowed by smart stepping on an idle Node/Express server
An explicit DAP `pause` can be acknowledged, then consumed by automatic stepping without producing a `stopped` event while the debuggee continues handling requests.

## Environment
- js-debug DAP server v1.112.0 (unmodified release bundle)
- Node.js v24.14.1; Windows x64
- Reproduced with a standalone raw TCP/DAP client, without mcp-debugger in the protocol path.

## Reproduction
Use this small Express server (Express 5.2.1):

```js
import express from 'express';
const app = express();
app.get('/ping', (_req, res) => res.json({ ok: true }));
const server = app.listen(0, '127.0.0.1', () => {
  console.log(`listening ${server.address().port}`);
});
```

1. Launch it through the standalone DAP server with `type: 'pwa-node'`, `stopOnEntry: false`, `smartStep: true`, `skipFiles: ['<node_internals>/**', '**/node_modules/**']`, `sourceMaps: true`, `pauseForSourceMap: false`, and `autoAttachChildProcesses: false`.
2. Complete initialize/configurationDone and the reverse startDebugging child adoption. Wait until the server is listening, make one successful /ping request, then wait one second.
3. Send `pause` to the adopted child's actual thread ID returned by `threads`.
4. Send GET /ping every 250 ms for ten seconds while recording DAP events.

## Observed
- The pause response succeeds.
- With `smartStep: true`: 1,517 `continued` events, 40 successful HTTP responses, and **zero `stopped` events** during the 10.2-second observation window.
- With `smartStep: false` and the same skip list: a `stopped` event with reason `pause`, no continued events, and an inspectable `emitHook` frame in `<node_internals>/internal/async_hooks`. The subsequent HTTP request remains pending while paused, as expected.
- These counts describe one run, not required thresholds for reproduction.

## Expected
Once JavaScript is executing and the requested pause is reached, expose a usable stopped state. An internal or skipped frame is preferable to silently continuing to execute indefinitely. Choosing to search for a user frame first could be reasonable, but that search needs a fallback that actually stops.

## Suspected mechanism and possible fix
[Thread.pause()](https://github.com/microsoft/vscode-js-debug/blob/41bd9bb831a46c2668cab14d8455e8e3bff58477/src/adapter/threads.ts) records `_expectedPauseReason = { reason: 'pause' }`. [SmartStepper.getSmartStepDirection()](https://github.com/microsoft/vscode-js-debug/blob/41bd9bb831a46c2668cab14d8455e8e3bff58477/src/adapter/smartStepping.ts) exempts breakpoint/exception/entry stops, but does not exempt that explicit pause intent; it chooses another step for a skipped frame and eventually switches to step-out after its threshold.

A focused candidate fix would bypass synthetic stepping for an explicitly requested pause, preserving normal smart stepping and skipFiles behavior for actual step requests. Regression coverage should include manual pause in skipped frames and unchanged stepping through dependencies and generated helpers. This candidate is now implemented and has been tested against upstream main at `41bd9bb831a46c2668cab14d8455e8e3bff58477`.

## Current-main confirmation
A minimal integration fixture (`setInterval(() => {}, 10)` plus a readiness message), with both its script and Node internals skipped, also reproduces this on current main. Before the fix, the explicit pause succeeds but the test times out at ten seconds; the DAP send trace contains 1,437 `continued` events and no `stopped` event. After the fix, it receives a `stopped` event with reason `pause`, reads the stack, evaluates an expression, and resumes successfully. The standalone Express/DAP probe also pauses successfully with `smartStep: true` after rebuilding the adapter with the fix.

Related: #1085 introduced synthetic blackboxing; #429 exempts other deliberate stops. This report is separate from [debugmcp/mcp-debugger#687](https://github.com/debugmcp/mcp-debugger/issues/687), which deliberately retained smartStep's launch default after measuring the costs of disabling it. The mcp-debugger-side launch/auto-continue fixes in [#708](https://github.com/debugmcp/mcp-debugger/pull/708) are present in the downstream reproduction.

debugmcpdev:
An explicit DAP pause in skipped or unmapped code can be acknowledged and then consumed by automatic stepping, leaving the debuggee running without a usable `stopped` event.

Honor the explicit pause intent already recorded by `Thread.pause()` and reset the automatic-step counter. A manual pause can now expose an internal or generated frame; ordinary stepping, pauses without explicit pause intent, and the existing breakpoint/exception/entry exemptions retain their behavior. No launch defaults or public configuration change.

Fixes #2412

## Validation
Tested on Windows x64 with Node v24.14.1, based on `41bd9bb831a46c2668cab14d8455e8e3bff58477`.

- New unit coverage exercises skipped and unmapped frames, preserved step directions and deliberate stops, and counter reset. Before the fix, four assertions fail; all 18 cases pass afterward.
- The new Node integration test times out on unpatched main after an acknowledged pause and 1,437 `continued` events without a `stopped` event. With the fix, it pauses, reads the stack, evaluates an expression, and resumes.
- `npm run compile`, `npm run test:types`, `npm run test:unit` (284 passing), and `npm run test:lint` pass.
- `npm run test:golden`: 435 passing, 8 pending, 2 failing. The Deno `doesn't duplicate --allow-all` failure also reproduces on unpatched main. `pretty print sources / bps` passes when rerun on both main and this branch; a focused rerun with the new pause regression has 2 passing tests.
- The rebuilt standalone DAP server passes the Express reproduction with `smartStep: true`. Downstream dogfooding through mcp-debugger's development proxy also succeeds: pause an idle mcp-debugger server, inspect/evaluate its state, hit a source-mapped TypeScript breakpoint in its HTTP handler, step over, and resume an HTTP 200 response.


debugmcpdev:

connor4312 (approved):
thanks!

dmitrivMS (approved):

roblourens (approved):

